PR #115, opened 2021-10-01
Fix redirect syntax

ryanahamilton:


paolaperaza (approved):
last one's the charm 🍀

PR #126, opened 2021-10-13
Deferrable Operators documentation

jwitz:
This PR includes our dedicated documentation page for Deferrable Operators, which is based on Andrew's draft on Notion.


docs/deferrable-operators.md
```diff
@@ -0,0 +1,85 @@
...
+
+### How It Works
+
+Airflow 2.2 introduces a new concept called a Trigger. Triggers are small asynchronous Python objects designed to be lightweight and highly-available. In order for an Operator to be deferrable, it must have its own Trigger code that determines when and how the Operator is deferred.
+
+Because of this, thousands of Triggers can be run in a single process, resulting in significant efficiency improvements over running sensors and Operators in a single process each.
+
+Deferrable Operators are simply Operators that are designed around a companion Trigger. The process for running a Deferrable Operator is as follows:
+
+1. The Operator calls Airflow to determine if it should defer based on a new Trigger.
```
jaketf, line 27: what does "calls Airflow" mean? 

I feel this section is struggling by trying to abstract too much from the reader but actually going through the lifecycle of a Defferable task (exampl https://cloud.google.com/dataproc/docs/concepts/jobs/life-of-a-job)

I'll take a stab at walking through each state transition, and technically what's going on @andrewgodwin keep me honest here:
0. Task is `queued` by the scheduler
1. Task gets picked up by a worker slot (based on whatever executor you are using) to execute an initial piece of code to initialize the task (this could be nothing, or taking an action like submitting a job to BigQuery). During this time it will be in the `running` state
2. The Task then defines a Trigger to defer some work to the Triggerer system (e.g. "triggerer call me back when the BigQuery Job is Done"). At this point we can free up the worker slot to take on other tasks. The task instance will enter the `deferred` state.
3. While in the `deferred` The Triggerer will handle runnning this trigger periodically to evaluate success.
4. Once the Trigger succeeds / fails the Task is again `queued` by the scheduler then `running` on a worker to do any completion work before being marked as a `success` or `failed`.

docs/deferrable-operators.md
```diff
@@ -0,0 +1,85 @@
...
+2. Airflow suspends the Operator's task and starts up the Trigger.
+3. If the Trigger fires, the Task is resumed and passed a small amount of state from before it suspended. From here, it can finish execution. If the Trigger doesn't fire, the task stays suspended in a deferred state.
+
+A new Airflow component called the Triggerer takes care of running Triggers and signaling tasks to resume when their conditions have been met. Like the Scheduler, it is designed to be highly-available: If a machine running Triggers shuts down unexpectedly, the Triggers can be recovered and moved to any warm-standby machine also running a Triggerer.
+
+To learn more about about deferring and implementation details, read the Apache Airflow [Deferring documentation](https://airflow.apache.org/docs/apache-airflow/stable/concepts/deferring.html).
+
+## Prerequisites
+
+To use Runtime's Deferrable Operators, you must first install the `astronomer-operator-wrappers` package. To do so, add the package to the `packages.txt` file of your Astronomer project. You also need to upgrade to [Astronomer Runtime 4.0.0](release-notes#astronomer-runtime-4-0-0) as described in [Upgrade Runtime](upgrade-runtime).
```
jaketf, line 37: aside: @andrewgodwin why is this called `astronomer-operator-wrappers`? Are they not just `astronomer-operators` or does this package provide wrappers to modify behavior of exisiting operators (like a decorator)?

docs/runtime-release-notes.md
```diff
@@ -29,13 +29,11 @@ For more information on using timetables, read the [Apache Airflow Documentation
 
 #### Deferrable Operators
 
-Deferrable operators are a new type of Airflow operator that promises improved performance and lower resource costs. While standard operators and sensors take up a "worker slot" even when they are waiting for an external trigger and not actually executing a task, deferrable operators are designed to suspend themselves and free up that "worker slot" while they wait. This is made possible by a new, lightweight Airflow component called the Triggerer.
+[Deferrable operators](https://airflow.apache.org/docs/apache-airflow/stable/concepts/deferring.html) are a new type of Airflow operator that promises improved performance and lower resource costs. While standard operators and sensors take up a "worker slot" even when they are waiting for an external trigger and not actually executing a task, Deferrable Operators are designed to suspend themselves and free up that "worker slot" while they wait. This is made possible by a new, lightweight Airflow component called the Triggerer.
```
jaketf, line 32: this paragraph is slightly different than the one in the deferrable-operators.md and has some stylisitic inconsistencies (e.g. "worker slot" in quotes)

jaketf (changes requested):

docs/deferrable-operators.md
```diff
@@ -0,0 +1,85 @@
...
+id: 'deferrable-operators'
+---
+
+## Overview
+
+Apache Airflow 2.2 introduces the concept of deferring, which is a powerful new framework for executing tasks. This framework is implemented through Deferrable Operators, which are a new type of Airflow Operator that promises improved performance and lower resource costs compared to standard Operators.
+
+While standard Operators and sensors take up a Worker or Scheduler slot when they are waiting for an external trigger, Deferrable Operators suspend themselves and free up that slot when in a defferred state. This results in a significant performance improvement for operators or sensors that spend a long time in a deferred state, such as the `S3Sensor`, the `HTTPSensor`, and the `DatabricksSubmitRunOperator`.
+
+The deferring function of Deferrable Operators, known as a Trigger, runs on a new, highly-available Triggerer component within Airflow. This means that Deferrable Operators waiting in a deferred phase are also highly-available, which makes it easy to recover from multi-machine outages.
```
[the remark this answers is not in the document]
jwitz: I might actually cut out triggers entirely from this overview. It's not necessary to know right away, since a user could theoretically use Deferrable Operators without ever interacting with Triggers right?

docs/deferrable-operators.md
jwitz: Great suggestions here. I did some minor rewording, but generally took this language as-is. When you say that the Trigger can succeed/fail in step 4, what happens if the Trigger "fails"? Does that mean the task retries or gives an error?

docs/deferrable-operators.md
```diff
@@ -0,0 +1,85 @@
...
+sidebar_label: 'Deferrable Operators'
+title: 'Deferrable Operators'
+id: 'deferrable-operators'
+---
+
+## Overview
+
+Apache Airflow 2.2 introduces the concept of deferring, which is a powerful new framework for executing tasks. This framework is implemented through Deferrable Operators, which are a new type of Airflow Operator that promises improved performance and lower resource costs compared to standard Operators.
+
+While standard Operators and sensors take up a Worker or Scheduler slot when they are waiting for an external trigger, Deferrable Operators suspend themselves and free up that slot when in a defferred state. This results in a significant performance improvement for operators or sensors that spend a long time in a deferred state, such as the `S3Sensor`, the `HTTPSensor`, and the `DatabricksSubmitRunOperator`.
```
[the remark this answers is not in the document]
jwitz: I think keeping language around "Conditions" and only using trigger in the context of the Python object itself is a good call.

jwitz:
Great comments here @jaketf , made some updates to the doc and left some comments/ questions in the replies.

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+id: 'deferrable-operators'
+---
+
+## Overview
+
+Apache Airflow 2.2 introduces the concept of [deferring](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=177050929), which is a powerful new framework for executing tasks. This framework is implemented through Deferrable Operators, which are a new type of Airflow Operator that promises improved performance and lower resource costs compared to standard Operators.
+
+While standard Operators and sensors take up a Worker or Scheduler slot when they are waiting for an external condition to be met, Deferrable Operators suspend themselves and free up that slot when in a deferred state. This results in a significant performance improvement for tasks using operators or sensors that spend a long time in a deferred state, such as the `S3Sensor`, the `HTTPSensor`, and the `DatabricksSubmitRunOperator`.
+
+Deferred tasks using these Operators run on a new, highly-available Triggerer component within Airflow. This means that Deferrable Operators waiting in a deferred phase are also highly-available, which makes it easy to recover from multi-machine outages.
```
paolaperaza, line 13:
```suggestion
Deferred tasks using these Operators run on a new, highly-available component within Airflow called the Triggerer. This means that Deferrable Operators waiting in a deferred phase are also highly-available, which makes it easy to recover from multi-machine outages.
```

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+
+On Astronomer Cloud, Triggerers are automatically managed by Astronomer Cloud, meaning you can start using Deferrable Operators in your DAGs without additional implementation steps. Additionally, you have access to several deferrable versions of open source operators that are available only on Astronomer.
```
paolaperaza, line 15:
```suggestion
The Triggerer component is supported and entirely managed on Astronomer Cloud, which means you can start using Deferrable Operators in your DAGs as long as you're running Astronomer Runtime 4.0+. As an Astronomer customer, you additionally have exclusive access to several deferrable versions of open source operators written by our team.
```

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+
+This guide explains how Deferrable Operators work, how to implement Deferrable Operators in your DAGs, and how Astronomer's own Deferrable Operators differ from their non-deferrable counterparts.
```
paolaperaza, line 17:
```suggestion
This guide explains how Deferrable Operators work, how to implement Deferrable Operators in your DAGs, and how Astronomer's Deferrable Operators differ from their non-deferrable counterparts.
```

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+
+### How It Works
+
+Airflow 2.2 introduces a new concept called a Trigger. A Trigger is a small, asynchronous Python function that quickly and continuously evaluates a given condition. In order for an Operator to be deferrable, it must have its own Trigger code that determines when and how the Operator is deferred.
+
+Because of this, thousands of Triggers can be run in a single process, resulting in significant efficiency improvements over running sensors and Operators in a single process each.
+
+Deferrable Operators are simply Operators that are designed around a companion Trigger. The process for running a task using a Deferrable Operator is as follows:
```
paolaperaza, line 25: This first sentence here is slightly redundant with second sentence in first paragraph - maybe pull this one up? I like the "this is what a trigger is" and "deferrable operators are just ones that use this thing" sequence

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+
+1. The task is picked up by a Worker slot to execute an initial piece of code to initialize the task. During this time, it will be in a "running" state.
+2. The task defines a Trigger and defers some condition checking work to the Triggerer. Because all of the deferring work happens in the Triggerer, the task instance can now enter a "deferred" state. This frees the Worker slot to take on other tasks.
```
paolaperaza, line 28:
```suggestion
2. The task defines a Trigger and defers the function of checking on some condition to the Triggerer. Because all of the deferring work happens in the Triggerer, the task instance can now enter a "deferred" state. This frees the Worker slot to take on other tasks.
```

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+3. The Triggerer runs the task's Trigger periodically to evaluate whether the condition is met.
```
paolaperaza, line 29: Might actually be good to note how often it checks

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+4. Once the Trigger condition succeeds, the Task is again queued by the Scheduler. This time, when the task is picked up by a Worker, it begins to complete its main function.
+
+A new Airflow component called the Triggerer takes care of running Triggers and signaling tasks to resume when their conditions have been met. Like the Scheduler, it is designed to be highly-available: If a machine running Triggers shuts down unexpectedly, the Triggers can be recovered and moved to any warm-standby machine also running a Triggerer.
```
paolaperaza, line 32: This is good info, but you kinda already mentioned the Triggerer a few times - maybe pull up this level of detail above, or split it up a bit to acknowledge that you already defined it and it doesn't sound like this is the first time you're introducing the concept

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+
+To learn more about deferring and implementation details, read the Apache Airflow [Deferring documentation](https://airflow.apache.org/docs/apache-airflow/stable/concepts/deferring.html).
+
+## Prerequisites
+
+To use Runtime's Deferrable Operators, you must first install the `astronomer-operator-wrappers` package. To do so, add the package to the `packages.txt` file of your Astronomer project. You also need to upgrade to [Astronomer Runtime 4.0.0](release-notes#astronomer-runtime-4-0-0) as described in [Upgrade Runtime](upgrade-runtime).
```
paolaperaza, line 38: You know I like bullet point lists! Why not:

> To use Runtime's Deferrable Operators, you must:
> 
> - Upgrade to [Astronomer Runtime 4.0.0](release-notes#astronomer-runtime-4-0-0) as described in [Upgrade Runtime](upgrade-runtime)
> - Add the `astronomer-operator-wrappers` package to the `packages.txt` file of your Astronomer project

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+
+Note that these Operators are available only in commercial Astronomer environments, such as Astronomer Cloud. The package is designed to fall back to the non-deferrable options in other environments, but the deferrable portions work only on Astronomer.
```
paolaperaza, line 40: Which operators? The Astro ones? Got a lil lost initially reading this sentence - make sure to not imply that deferrable operators only work on Astronomer, it's just THESE operators

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+Some additional notes about using Deferrable Operators:
+
+- Currently, not all Operators have a deferrable version. There are a few open source Deferrable Operators, plus additional Operators designed and maintained by Astronomer.
+- You can write your own Deferrable Operators and Triggers for custom API or service usage. Astronomer is happy to help customers with this process.
+
+## Astronomer's Deferrable Operators
+
+Astronomer Runtime includes a collection of Deferrable Operators available exclusively to Astronomer Cloud users. These are drop-in replacements for non-Deferrable Operators, meaning you only have to change the import statements in your DAGs to begin using them.
+
+This section contains information on each available Deferrable Operator, including example import statements. If there are any differences between a Deferrable Operator and its non-deferrable counterpart, those changes are listed in a Notes section.
```
paolaperaza, line 64: Yea, to my comment above maybe delete the sentences above and just add the whole "only available on Astronomer" here where it's relevant

docs/deferrable-operators.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Databricks Operators
+
+We provide deferrable versions of both the `DatabricksSubmitRunOperator` and the `DatabricksRunNowOperator`. These Operators wait in a deferred state while waiting for their respective Databricks jobs to complete.
+
+### Import statement
```
paolaperaza, line 70: Should this be H4?

docs/deferrable-operators.md
jwitz: I think the Trigger defines this?

docs/deferrable-operators.md
jwitz: Oof not sure how this got down here, should be before the steps!

jwitz:
@paolaperaza Thanks for these comments! I didn't change much language on this pass - it was mostly just reorganizing content to fit in the right place.

docs/deferrable-operators.md
jaketf: @jwitz I assume if a trigger fails the task will be marked as failed and if it's configured to be re-tried the entire task would be retried from the beginning.

@andrewgodwin is there any concept of retrying just the trigger if something fails in the trigger itself?

docs/deferrable-operators.md
```diff
@@ -0,0 +1,113 @@
...
+sidebar_label: 'Deferrable Operators'
+title: 'Deferrable Operators'
+id: 'deferrable-operators'
+---
+
+## Overview
+
+[Apache Airflow 2.2](https://airflow.apache.org/blog/airflow-2.2.0/) introduces [Deferrable Operators](https://airflow.apache.org/docs/apache-airflow/stable/concepts/deferring.html), a powerful new framework for executing tasks. Available in Astronomer Runtime 4.0+, Deferrable Operators are a new type of Airflow Operator that promises improved performance and lower resource costs.
+
+In Airflow, it's common for users to configure tasks that wait for some condition to be met before executing or triggering another task. While standard Airflow Operators and Sensors perform that function well, tasks have historically needed to take up a Worker or Scheduler slot the entire time that they are waiting for an external condition to be met.
```
jaketf, line 11: > tasks have historically needed to take up a Worker or Scheduler slot the entire time

this is not entirely technically true for sensors 🤓 . There is a concept of "reschedule" mode since 1.10.2 ish. which frees up the worker slot inbetween pokes (basically deferrable sensor but pokes each run on workers with an up for reschedule mode). One of the main benefits is you don't need 2 separate tasks (submit operator and poll for success sensor) and of course the triggerer is wayyy more resource efficient because it is built to quickly run many concurrent quick little pokes (allowing concurrency of 1000s instead of 10s on the same pod).

docs/deferrable-operators.md
```diff
@@ -0,0 +1,113 @@
...
+3. The Triggerer runs the task's Trigger periodically to check whether or not the condition has been met.
+4. Once the Triggerer condition succeeds, the Task is again queued by the Scheduler. This time, the task begins executing its main function when as soon as it's picked up by a Worker.
+
+To learn more about deferring, read the Apache Airflow [documention on Deferring](https://airflow.apache.org/docs/apache-airflow/stable/concepts/deferring.html).
+
+## Using Deferrable Operators
+
+Deferrable Operators are drop-in replacements for their traditional Operator counterparts. Airflow 2.2.0 shipped with `DateTimeSensorAsync` and `TimeDeltaSensorAsync`, but other Sensors and Operators need to be written. As an Astronomer customer, you have exclusive access to an additional set of Deferrable Operators designed and maintained by our team.
+
+If you're interested in writing and contributing a Deferrable Operator and Trigger for your own use case, we're happy to help with the process.
```
jaketf: > we're happy to help with the process

who exactly is happy to help ? are we going to have a services engagement model for this sort of thing (field team)? is this just airflow team out of goodness of their hearts?

docs/deferrable-operators.md
jaketf: @paolaperaza bump on this comment I still think we should acknowledge sensor reschedule mode and how deferrable operators are different.

docs/deferrable-operators.md
```diff
@@ -6,82 +6,108 @@ id: 'deferrable-operators'
...
+### Prerequisites
 
-To use Runtime's Deferrable Operators, you must first:
+To use Deferrable Operators on Astronomer, you must first:
 
 - Add the `astronomer-operator-wrappers` package to the `packages.txt` file of your Astronomer project.
 - Upgrade to [Astronomer Runtime 4.0.0](release-notes#astronomer-runtime-4-0-0) as described in [Upgrade Runtime](upgrade-runtime).
 
-## Using Deferrable Operators
+### Import a Deferrable Operator
```
jwitz, line 52: Don't think we should make the core component of this document -- how to use Deferrable Operators -- into an h3 header.

docs/deferrable-operators.md
```diff
@@ -0,0 +1,97 @@
...
+- Currently, not all Operators have a deferrable version. There are a few open source Deferrable Operators, plus additional Operators designed and maintained by Astronomer.
+- You can write your own Deferrable Operators and Triggers for custom API or service usage. If you need help with writing a custom Deferrable Operator, reach out to your Astronomer representative.
+
+## Astronomer's Deferrable Operators
+
+Astronomer Runtime includes a collection of Deferrable Operators that are available exclusively to Astronomer Cloud users. These are drop-in replacements for non-Deferrable Operators, meaning that you only have to change the import statements in your DAGs to begin using them.
+
+This section contains information and example import statements for all Deferrable Operators available exclusively on Astronomer Runtime.
+
+> **Note:** When Operators in the `astronomer-operator-wrappers` in environments that don't run on Astronomer Runtime, the package falls back to using the open source, non-deferrable versions of each Operator.
```
andrewgodwin, line 66: "Operators in the `astronomer-operator-wrappers` in environments" needs a grammar fix - either a "package" added or "the ... in" removed and both would need "are used" or something.

andrewgodwin (commented):
Should we also add a small note somewhere that mentions how deferrable operators only make sense when waiting for more than (say) 5 seconds?

docs/deferrable-operators.md
```diff
@@ -0,0 +1,97 @@
...
+sidebar_label: 'Deferrable Operators'
+title: 'Deferrable Operators'
+id: 'deferrable-operators'
+---
+
+## Overview
+
+Introduced in [Apache Airflow 2.2](https://airflow.apache.org/blog/airflow-2.2.0/), Deferrable Operators are a powerful type of Airflow Operator that you can use to improve the performance of your Deployments and lower your resource costs.
+
+Tasks using standard Operators and sensors take up a Worker or Scheduler slot when checking if an external condition has been met. In comparison, Deferrable Operators suspend themselves and don't take up a Worker or Scheduler slot when they are in a deferred state. Using the Deferrable versions of Operators or Sensors that typically spend a long time waiting for a condition to be met, such as the `S3Sensor`, the `HTTPSensor`, or the `DatabricksSubmitRunOperator`, can result in significant per-task cost savings and performance improvements.
```
jaketf, line 11: I'm fine if you feel strongly we shouldn't acknowledge reschedule mode operators explicitly in this doc, but they aren't just an item of the past users can still use them in airflow today. Standard sensors can be used in reschedule mode and "free themselves up between pokes" so I still find this paragraph misleading.

It was the first thing I noticed when reading these docs, and it can damage our credibility, and even lead to confusion to users who have used reschedule mode sensors in the past.

If we were going to contrast the two.
Compared to triggers, reschedule mode sensors

- don't require managing this new triggerer component (everything happens on workers)
- don't require separate async implementations (just an enabled with a kwarg)
- don't require airflow >2.2.
- cannot pass any state between rescheduled pokes which can be limiting to sensor authors (can be limiting in complex sensor use cases e.g. "a sensor that detects a certain amount of time between files landing in a bucket")
- take up a worker slot during a poke (which could just be done by the triggerer which could handle much higher concurrency of these pokes than a worker)

docs/deferrable-operators.md
```diff
@@ -56,6 +56,7 @@ Some additional notes about using Deferrable Operators:
 - In the previous example, the statement imports the Deferrable Operator class with its non-deferrable class name, meaning that you don't have to replace existing instances of non-deferrable Operators in your DAGs. If you don't include this part of the import statement, you additionally need to replace all instances of non-deferrable Operators with Deferrable Operators.
 - Currently, not all Operators have a deferrable version. There are a few open source Deferrable Operators, plus additional Operators designed and maintained by Astronomer.
 - You can write your own Deferrable Operators and Triggers for custom API or service usage. If you need help with writing a custom Deferrable Operator, reach out to your Astronomer representative.
+- There are some scenarios where it's more appropriate to use a traditional sensor in [`reschedule`](https://github.com/apache/airflow/blob/1.10.2/airflow/sensors/base_sensor_operator.py#L46-L56) mode. If your task doesn't often defer because it needs to wait only a few seconds for a condition to be met, then we recommend using a traditional sensor instead of a Deferrable Operator.
```
jwitz, line 59:
```suggestion
- There are some scenarios where it's more appropriate to use a traditional sensor in [`reschedule`](https://github.com/apache/airflow/blob/1.10.2/airflow/sensors/base_sensor_operator.py#L46-L56) mode for checking if a condition has been met. If your task doesn't often defer because it needs to wait only a few seconds for a condition to be met, then we recommend using a traditional sensor instead of a Deferrable Operator.
```

docs/deferrable-operators.md
```diff
@@ -0,0 +1,98 @@
...
+```python
+# Remove this import:
+# from airflow.operators.sensors import ExternalTaskSensor
+# Replace with:
+from airflow.sensors.time_sensor import TimeSensorAsync as TimeSensor
+```
+
+Some additional notes about using Deferrable Operators:
+
+- In the previous example, the statement imports the Deferrable Operator class with its non-deferrable class name, meaning that you don't have to replace existing instances of non-deferrable Operators in your DAGs. If you don't include this part of the import statement, you additionally need to replace all instances of non-deferrable Operators with Deferrable Operators.
```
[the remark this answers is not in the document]
paolaperaza: - "Previous example" makes it sound like you're now talking about a "current" example
- Made the implication here clearer at the end

docs/deferrable-operators.md
```diff
@@ -0,0 +1,98 @@
...
+> **Note:** When you use Deferrable Operators in the `astronomer-operator-wrappers` package outside of a Runtime environment, the package falls back to using the open source, non-deferrable versions of each Operator.
+
+### Databricks Operators
+
+Astronomer Runtime includes the following Databricks Operators:
+
+- `DatabricksSubmitRunOperator`
+- `DatabricksRunNowOperator`
+
+These Operators wait in a deferred state while waiting for their respective Databricks job to complete.
```
paolaperaza, line 76:
```suggestion
Tasks configured with these Operators remain in a deferred state while waiting for their respective Databricks job to complete.
```

docs/deferrable-operators.md
```diff
@@ -0,0 +1,98 @@
...
+
+To run Astronomer's Databricks Operators, use the following import statement:
+
+```python
+from astronomer.operators import DatabricksSubmitRunOperator, DatabricksRunNowOperator
+```
+
+### ExternalTaskSensor
+
+This is a drop-in replacement for Airflow's [`ExternalTaskSensor`](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/sensors/external_task/index.html#module-airflow.sensors.external_task). It defers itself while waiting for a given Task or DAG to complete.
```
paolaperaza, line 88:
```suggestion
This is a drop-in replacement for Airflow's [`ExternalTaskSensor`](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/sensors/external_task/index.html#module-airflow.sensors.external_task). It defers itself while waiting for a given task or DAG to complete.
```

docs/deferrable-operators.md
```diff
@@ -0,0 +1,98 @@
...
+
+> **Note:** There is a difference between the deferrable ExternalTaskSensor and the non-deferrable ExternalTaskSensor: If there are multiple matching Tasks or DAGs and any of them have failed, the deferrable sensor also fails, whereas the non-deferrable sensor freezes indefinitely.
```
paolaperaza, line 90:
```suggestion
> **Note:** There is a difference between the deferrable ExternalTaskSensor and the non-deferrable ExternalTaskSensor. If any tasks or DAGs that the Sensor is listening to fail, the deferrable sensor also fails, whereas the non-deferrable sensor freezes indefinitely.
```

docs/deferrable-operators.md
```diff
@@ -0,0 +1,98 @@
+---
+sidebar_label: 'Deferrable Operators'
+title: 'Deferrable Operators'
+id: 'deferrable-operators'
+---
+
+## Overview
+
+Introduced in [Apache Airflow 2.2](https://airflow.apache.org/blog/airflow-2.2.0/), Deferrable Operators are a powerful type of Airflow Operator that you can use to improve the performance of your Deployments and lower your resource costs.
```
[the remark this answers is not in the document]
jwitz: We should always strive for active voice in documentation. "Brings" is ascribing an action/ agency to the Operators themselves

docs/deferrable-operators.md
```diff
@@ -0,0 +1,97 @@
...
+
+In Airflow, you can configure tasks to wait for some external condition to be met before executing or triggering another task. Tasks using standard Operators and Sensors take up a Worker or Scheduler slot when checking if an external condition has been met. In comparison, tasks using Deferrable Operators suspend themselves and don't take up a Worker or Scheduler slot when checking if a condition has been met. Using the Deferrable versions of Operators or Sensors that typically spend a long time waiting for an condition to be met, such as the `S3Sensor`, the `HTTPSensor`, or the `DatabricksSubmitRunOperator`, can result in significant per-task cost savings and performance improvements.
```
paolaperaza, line 11:
```suggestion
In Airflow, it's common to use [Sensors](https://airflow.apache.org/docs/apache-airflow/stable/concepts/sensors.html) and some [Operators](https://airflow.apache.org/docs/apache-airflow/stable/concepts/operators.html) to configure tasks that wait for some external condition to be met before executing or triggering another task. While tasks using standard Operators and Sensors take up a Worker or Scheduler slot when checking if an external condition has been met, Deferrable Operators suspend themselves during that process. This releases the Worker to take on other tasks. Using the Deferrable versions of Operators or Sensors that typically spend a long time waiting for an condition to be met, such as the `S3Sensor`, the `HTTPSensor`, or the `DatabricksSubmitRunOperator`, can result in significant per-task cost savings and performance improvements.
```

- Added link to OSS docs on Sensors and Operators docs to bridge the logic between the first two sentences
- I _really_ just don't really love the transition and use of "In comparison here." Feels too academic and neutral to be using to introduce a super powerful new feature. It's like you're saying, "this happens this way, and that happens that way" but not tying those together into a "so what".
- I'd like to explicitly say "this means the worker can take on other tasks"  - that's the whole thing and important piece of logic.

docs/deferrable-operators.md
```diff
@@ -0,0 +1,97 @@
...
+
+Deferrable Operators rely on a new, highly available Airflow component called the Triggerer. The Triggerer is entirely managed on Astronomer Cloud, meaning that you can start using Deferrable Operators in your DAGs as long as you're running Astronomer Runtime 4.0+. As an Astronomer customer, you additionally have exclusive access to several deferrable versions of open source Operators.
```
paolaperaza, line 13:
```suggestion
Deferrable Operators rely on a new Airflow component called the Triggerer. The Triggerer is highly available and entirely managed on Astronomer Cloud, meaning that you can start using Deferrable Operators in your DAGs as long as you're running Astronomer Runtime 4.0+. As an Astronomer customer, you additionally have exclusive access to several deferrable versions of open source Operators.
```

Follow up on comment - it's _designed_ to be HA if a user configures it that way, but we actually have to support that setup. Note the: "If you want to...run multiple copies....". We (Astronomer) have to make the decision to do that.

> Triggers are designed from the ground-up to be highly-available; if you want to run a highly-available setup, simply run multiple copies of triggerer on multiple hosts

docs/deferrable-operators.md
```diff
@@ -0,0 +1,97 @@
...
+1. The task is picked up by a Worker, which executes an initial piece of code that initializes the task. During this time, the task is in a "running" state and takes up a Worker slot.
+2. The task defines a Trigger and defers the function of checking on some condition to the Triggerer. Because all of the deferring work happens in the Triggerer, the task instance can now enter a "deferred" state. This frees the Worker slot to take on other tasks.
+3. The Triggerer runs the task's Trigger periodically to check whether the condition has been met.
+4. Once the Trigger condition succeeds, the task is again queued by the Scheduler. This time, when the task is picked up by a Worker, it begins to complete its main function.
+
+For implementation details on Deferrable Operators, read the [Apache Airflow documentation](https://airflow.apache.org/docs/apache-airflow/stable/concepts/deferring.html).
+
+## Prerequisites
+
+To use Deferrable Operators on Astronomer Cloud, you must first upgrade your Deployments to [Astronomer Runtime 4.0.0](release-notes#astronomer-runtime-4-0-0) as described in [Upgrade Runtime](upgrade-runtime).
```
paolaperaza, line 36:
```suggestion
To use Deferrable Operators on Astronomer Cloud, you must first upgrade to [Astronomer Runtime 4.0+](release-notes#astronomer-runtime-4-0-0) as described in [Upgrade Runtime](upgrade-runtime).
```

docs/deferrable-operators.md
```diff
@@ -0,0 +1,97 @@
...
+
+Additionally, if you want to use Deferrable Operators available exclusively on Astronomer Runtime, you must add the `astronomer-operator-wrappers` package to the `packages.txt` file of your Astronomer project.
```
paolaperaza, line 38:
```suggestion
To use Deferrable Operators available exclusively on Astronomer Runtime, you must additionally add the `astronomer-operator-wrappers` package to the `packages.txt` file of your Astronomer project.
```

docs/deferrable-operators.md
```diff
@@ -0,0 +1,97 @@
...
+# from airflow.operators.sensors import ExternalTaskSensor
+# Replace with:
+from airflow.sensors.time_sensor import TimeSensorAsync as TimeSensor
+```
+
+Some additional notes about using Deferrable Operators:
+
+- For open source Operators, we recommend importing the Deferrable Operator class as its non-deferrable class name. If you don't include this part of the import statement, you need to replace all instances of non-deferrable Operators in your DAGs. In the above example, that would require replacing all instances of `TimeSensor` with `TimeSensorAsync`.
+- Currently, not all Operators have a deferrable version. There are a few open source Deferrable Operators, plus additional Operators designed and maintained by Astronomer.
+- You can write your own Deferrable Operators and Triggers and contribute these to the open source project. If you need help with writing a custom Deferrable Operator, reach out to your Astronomer representative.
```
paolaperaza, line 57:
```suggestion
- If you're interested in a Deferrable Operator that isn't generally available, you can write your own and contribute these to the open source project. If you need help with writing a custom Deferrable Operator, reach out to your Astronomer representative.
```

I feel like the guiding language here is still helpful?

docs/deferrable-operators.md
```diff
@@ -0,0 +1,97 @@
...
+- There are some use cases where it's more appropriate to use a traditional Sensor instead of a Deferrable Operator. If your task needs to wait only a few seconds for a condition to be met, we recommend using a Sensor in [`reschedule`](https://github.com/apache/airflow/blob/1.10.2/airflow/sensors/base_sensor_operator.py#L46-L56] mode to avoid unnecessary resource overhead.
```
paolaperaza, line 58:
```suggestion
- There are some use cases where it can be more appropriate to use a traditional Sensor instead of a Deferrable Operator. If your task needs to wait only a few seconds for a condition to be met, we recommend using a Sensor in [`reschedule`](https://github.com/apache/airflow/blob/1.10.2/airflow/sensors/base_sensor_operator.py#L46-L56] mode to avoid unnecessary resource overhead.
```

I'd still use "can be" to keep it looser - what is or isn't appropriate is not black and white

docs/deferrable-operators.md
```diff
@@ -0,0 +1,97 @@
...
+> **Note:** When you use Deferrable Operators in the `astronomer-operator-wrappers` package outside of an Astronomer Runtime environment, the package falls back to using the open source, non-deferrable versions of each Operator.
+
+### Databricks Operators
+
+Astronomer Runtime includes the following Databricks Operators:
+
+- `DatabricksSubmitRunOperator`
+- `DatabricksRunNowOperator`
+
+Tasks using these Operators wait in a deferred state while waiting for their respective Databricks job to complete.
```
paolaperaza, line 75:
```suggestion
Tasks using these Operators remain in a deferred state while waiting for their respective Databricks job to complete.
```

double wait

PR #155, opened 2021-11-06
New Modify Cluster Doc

paolaperaza:
This PR does a few things:

1. Adds two sections to our "AWS Resource Reference" doc:
    - Overview
    - Supported Cluster Configurations

This change explicitly lists all the AWS regions and node instance types that we support. This way, we can tell users to refer to this source of truth instead of having to re-list it.

2. Create a new "Modify Cluster doc". This doc is bare, but people will inevitably ask for new instance types and we can send them this URL. It also specifies expect impact on a change.

@chomipi88 Do you know if users can modify AWS regions for an existing Cluster? Anything else we want to add to this list?


chomipi88:
@paolaperaza I don't believe that's been tested. Modifying a region has significant implications on networking stack so I would default to "not supported" at this time.

As for listing instance types, this list may grow to be un-manageable pretty soon, not too mention the complexity of region availability. Not all instance types are available in all regions.

paolaperaza:
@chomipi88 I just added a comment to clarify that modifying AWS region for an existing Cluster is not supported. Totally acknowledge that this source-of-truth list may get unruly... I think we just adjust/automate this over time? Maybe even remove once customers can create/modify clusters on their own?

I don't want to crowd the install doc and I think sales/customer teams need a consolidated list of supported configs that users can pick from.

chomipi88 (approved):

PR #158, opened 2021-11-11
Add tip about CLI executable configuration

jwitz:
Resolves https://github.com/astronomer/cloud-docs/issues/148

Tagging Steven for review as he described the documented setup in Slack


docs/install-cli.md
```diff
@@ -81,7 +81,19 @@ As noted above, the Astronomer Cloud CLI is installed as a local executable. For
...
+To run the CLI more easily, we recommend configuring your local machine to run Astronomer CLI commands without needing to specify the absolute path to the executable file . You can do so in one of two ways:
+
+- Add the directory you created for your Astronomer project to your `$PATH` variable. For more information, see [Linux instructions](https://linuxize.com/post/how-to-add-directory-to-path-in-linux/#adding-a-directory-to-your-path).
+- Create a symbolic link in your `bin` directory using a command similar to the following:
+
+   ```sh
+   sudo ln -s /source/path/to/astro /usr/local/bin/astro
+   ```
+
+   If you're an existing [open source Astronomer CLI](https://github.com/astronomer/astro-cli) user, we recommend changing the name of this symbolic link to something other than `astro` (for example, `astrocloud`).
```
paolaperaza, line 95:
```suggestion
   If you're an existing user of the original [Astronomer CLI](https://github.com/astronomer/astro-cli), we recommend changing the name of this symbolic link and creating an alias to something other than `astro` (for example, `astrocloud`).
```

paolaperaza:
@jwitz Thanks for adding this - a little tricky here since I think we'll wanna open source both CLIs. We honestly might just wanna edit this doc to say `Astronomer Cloud CLI`?

paolaperaza:
Also, per this Slack convo with @ben-astro we should change this line:

> As noted above, the Astronomer Cloud CLI is installed as a local executable. For that reason, commands must be run from the directory in which you installed the CLI and explicitly reference the binary (`astro`). For the equivalent of `astro dev init`, for example, you'd have to run:

-->

>  As noted above, the Astronomer Cloud CLI is installed as a local executable. For that reason, commands must explicitly reference the absolute path of the binary (`astro`). For the equivalent of `astro dev init`, for example, you'd have to run:

paolaperaza (approved):

docs/install-cli.md
ben-astro: Looks good to me! Thanks @paolaperaza !

ben-astro (approved):

PR #161, opened 2021-11-15
Add diagram to Deploy Code doc

jwitz:
Resolves https://github.com/astronomer/cloud-docs/issues/113

Check out the preview for this one to see the full diagram. I'm curious if people think this image might be too big? I could make it smaller, but that might sacrifice readability.


paolaperaza (approved):
I can't see the deploy preview and image, but the language around this LGTM @jwitz

PR #162, opened 2021-11-16
Add Docker Compose information to Develop Project doc

jwitz:
Resolves https://github.com/astronomer/cloud-docs/issues/140

Slight wording updates to our platform version of this doc, but the general content should be the same. My biggest changeL For verifying that the override was successful, I removed the code fence with showing the output of the `docker exec` command.


docs/develop-project.md
```diff
@@ -220,6 +220,32 @@ For example, if you want to run `ls` when your image builds, your `Dockerfile` w
 RUN ls
 `}</code></pre>
 
+## Override the CLI's Docker Compose File (Local Development Only)
+
+The Astronomer CLI is built on top of [Docker Compose](https://docs.docker.com/compose/), which is a tool for defining and running multi-container Docker applications. You can override the CLI's Docker Compose configurations by adding a `docker-compose.override.yml` file to your Astronomer project. Any values in this file override the CLI's default settings whenever you run `astro dev start`.
+
+To see what values you can override, reference the CLI's [Docker Compose file](https://github.com/astronomer/astro-cli/blob/main/airflow/include/composeyml.go). The linked file is for the original Astronomer CLI, but the values here are identical to those used in the Astronomer Cloud CLI.
```
paolaperaza, line 227:
```suggestion
To see what values you can override, reference the CLI's [Docker Compose file](https://github.com/astronomer/astro-cli/blob/main/airflow/include/composeyml.go). The linked file is for the original Astronomer CLI, but the values here are identical to those used in the Astronomer Cloud CLI. Common use cases for Docker Compose overrides include:

- Modifying the ports at which the Airflow Webserver or Postgres database start on if another service is already running on those same ports (8080 and 5432, respectively). You can override this default and point your containers to a different port.
- Adding extra containers to mimic services that your Airflow environment needs to interact with locally, such as an SFTP server.
- Change the volumes mounted to any of your local containers.
```

docs/develop-project.md
```diff
@@ -220,6 +220,32 @@ For example, if you want to run `ls` when your image builds, your `Dockerfile` w
...
+
+Make sure to specify `version: "2"` and follow the format of the source code file linked above.
+
+To see your override file live in your local Airflow environment, run the following command for any container running Airflow:
+
+```sh
+docker exec -it <container-name> ls -al
+```
+
+> **Note:** The Astronomer CLI does not support overrides to environment variables. To learn more about setting environment variables, read [Environment Variables](environment-variables).
```
paolaperaza, line 247:
```suggestion
:::info

The Astronomer CLI does not support overrides to environment variables that are required globally. For the list of environment variables that Astronomer enforces, see [Global Environment Variables](global-variables). To learn more about environment variables, read [Environment Variables](environment-variables).

:::
```

paolaperaza (approved):
Two comments here for additional context, but this LGTM @jwitz ! Go ahead and merge whenever you're ready 😊

PR #168, opened 2021-11-22
Clean up trust policy with two External Ids in "Create Cluster" Doc

paolaperaza:
Per Slack conversation here: https://astronomer.slack.com/archives/C01UA03TPV1/p1637601248314100

Simplifying our example Trust Policy in the "Create Cluster" doc.

### DIFF
--- a/docs/create-cluster.md
+++ b/docs/create-cluster.md
@@ -57,7 +57,7 @@ If you only have 1 existing Cluster, you will see the following under **Policy D
       "Action": "sts:AssumeRole",
       "Condition": {
         "StringEquals": {
-          "sts:ExternalId": "<your-1st-External-ID>"
+          "sts:ExternalId": "<External-ID-1>"
         }
       }
     }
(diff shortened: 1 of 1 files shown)


docs/create-cluster.md
```diff
@@ -57,48 +57,39 @@ If you only have 1 existing Cluster, you will see the following under **Policy D
...
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "arn:aws:iam::406882777402:root"
+      },
+      "Action": "sts:AssumeRole",
+      "Condition": {
+        "StringEquals": {
+          "sts:ExternalId": [
+            <External-ID-1>,
+            <External-ID-2>,
```
chomipi88, line 86: @paolaperaza @cmarteepants should the comma after the second external ID be there?

cmarteepants: No, it should not. Great catch!

paolaperaza, line 86:
```suggestion
            <External-ID-2>
```

paolaperaza: Oh man! Good eye @chomipi88 - removed.

chomipi88 (approved):

cmarteepants (changes requested):
Small nit, but I would add quotes around the external-id references, since we're doing that above.

PR #170, opened 2021-11-23
Runtime 4.0.4 release notes

paolaperaza:
Couple things here:

- Added Runtime 4.0.4 to release notes
- Updated release notes file with latest Runtime + CLI (CLI was outdated)

@jwitz Should we have those "latest" CLI/Runtime versions in release notes be called from the Env Var you set? Realizing we're still adding those manually


jwitz (approved):

jwitz:
LGTM!

PR #174, opened 2021-11-25
temp fix for broken link

chomipi88:
current link gives a page not found


jwitz:
Closing this out as we've found the root problem of these relative links

PR #252, opened 2021-12-22
Add release notes for Runtime and Certified

jwitz:
Based on https://github.com/astronomer/ap-airflow/blob/master/2.2.3/CHANGELOG.md and https://github.com/astronomer/astro-runtime/blob/main/CHANGELOG.md


paolaperaza (approved):
LGTM! Going ahead and merging, @jwitz. Thank you for this!

Note for when you get back: I know we're not actively maintaining docs for earlier Enterprise versions, but I think let's make an exception for Certified versioning. I'd hate for a customer on 0.26 to go check which versions of Certified they can download and think that they can't move to Airflow 2.2.3 (which we absolutely want them to do).

I'll go ahead and merge this and let's backport those changes to prior Enterprise versions when you get back 😊

PR #278, opened 2022-01-07
Fix external ES config example

BasPH:


paolaperaza (approved):
LGTM - thanks @BasPH. Not sure why we had this as `false` originally! Merge whenever you're ready.

PR #297, opened 2022-01-17
Astronomer namespace is required for Let's Encrypt cert

BasPH:
Docs in Let's Encrypt guide don't set a namespace, and the cert ends up in the `default` namespace which fails the installation. This PR adds a namespace flag.

### DIFF
--- a/enterprise/renew-tls-cert.md
+++ b/enterprise/renew-tls-cert.md
@@ -109,12 +109,12 @@ Once you set up a TLS certificate for Astronomer, you'll need to establish a pro
     Then, create the certificate by running the following command and waiting a few minutes:
 
     ```sh
-    kubectl apply -f certificate.yaml
+    kubectl apply -f certificate.yaml -n <your-namespace>
     ```
 
 5. Ensure that the certificate was created by running:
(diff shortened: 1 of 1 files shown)


enterprise/renew-tls-cert.md
```diff
@@ -109,12 +109,12 @@ Once you set up a TLS certificate for Astronomer, you'll need to establish a pro
     Then, create the certificate by running the following command and waiting a few minutes:
 
     ```sh
-    kubectl apply -f certificate.yaml
+    kubectl apply -f certificate.yaml -n <your-namespace>
```
jwitz, line 112:
```suggestion
    kubectl apply -f certificate.yaml -n astronomer
```

If we assume the namespace `astronomer` like in our install guides, then I think we should do the same here.

jwitz (approved):
Small comment about using a placeholder value, but otherwise LGTM @BasPH . Thanks for jumping on this!

BasPH:
Sounds good to me. There are two more `-n <your-namespace>` in the doc, I'll change those too to keep things consistent.

PR #306, opened 2022-01-20
Fix versions which received extended support

jedcunningham:
`1.10.12` and `1.10.14` where extended through 2022-01, not `1.10.15`.


jwitz (approved):
Thanks for catching this @jedcunningham !

PR #310, opened 2022-01-23
Add Explicit K8s Compatibility Statement, Change EOM Date for Enterprise v0.25

astrorfox:
In order to be able to reference for existing and new customers, adding a statement on our strategy of K8s support under Enterprise/Platform. Statement has been reviewed with engineering.

PR #314, opened 2022-01-25
Extend 0.25 through December 2022

astrorfox:
After reviewing with Vishwas, we want to extend maintenance of 0.25 through December of this year to ease the burden on customers.


paolaperaza:
Closing this PR out as it conflicts with and is the same as #310 !

Issue #319, opened 2022-01-25
Airgap Guide Should Ommit Pinned Images


rob-1126:
**Affected Document:**

enterprise/install-airgapped.md

**Problem Type (Priority):**

- [ X ] Error (Critical): This document has existing information that's incorrect and prevents readers from completing the documented setup (Add the `error/critical` label to this issue)
- [ ] Misleading (Important): This document could benefit from additional context, guidance, or clarity (Add the `error/FYI` label to this issue)
- [ ] Typo (Low): Any issue related to spelling/ punctuation/ formatting (Add the `error/typo` label to this issue)

If users pin their versions in their config file they will end up running out of date images and miss out on most security updates (probably without realizing it) and likely start to accumulate broken behavior as users fail to update their pinnings.

The chart files specify a default tag level that should trickle through.

Remove all explicit tag: entries from the provided example file.

PR #320, opened 2022-01-25
remove explicit tagging of image versions that can break upgrades

rob-1126:
https://github.com/astronomer/docs/issues/319

### DIFF
--- a/software_versioned_docs/version-0.27/install-airgapped.md
+++ b/software_versioned_docs/version-0.27/install-airgapped.md
@@ -91,27 +91,20 @@ alertmanager:
   images:
     alertmanager:
       repository: 012345678910.dkr.ecr.us-east-1.amazonaws.com/myrepo/myteam-ap-alertmanager
-      tag: 0.23.0
 astronomer:
   images:
     commander:
       repository: 012345678910.dkr.ecr.us-east-1.amazonaws.com/myrepo/myteam-ap-commander
-      tag: 0.26.4
     registry:
       repository: 012345678910.dkr.ecr.us-east-1.amazonaws.com/myrepo/myteam-ap-registry
-      tag: 3.14.2-2
     houston:
       repository: 012345678910.dkr.ecr.us-east-1.amazonaws.com/myrepo/myteam-ap-houston-api
-      tag: 0.27.1
     astroUI:
       repository: 012345678910.dkr.ecr.us-east-1.amazonaws.com/myrepo/myteam-ap-astro-ui
-      tag: 0.27.1
     dbBootstrapper:
       repository: 012345678910.dkr.ecr.us-east-1.amazonaws.com/myrepo/myteam-ap-db-bootstrapper
-      tag: 0.26.1
     cliInstall:
       repository: 012345678910.dkr.ecr.us-east-1.amazonaws.com/myrepo/myteam-ap-cli-install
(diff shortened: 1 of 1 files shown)


rob-1126:
looks like this needs to cary a 0.27.2 as a minimum version as I'm showing prior to that nats and stan expected the tag as part of the image name.

jwitz:
@rob-1126 At the top of this code fence we state `# Example configuration only. To determine the necessary images, render Helm template for your Astronomer installation version.` Should I also include a warning note before to make that more clear? I think it's helpful to have a complete example even if it might be out of date.

rob-1126:
@jwitz I'd not even demonstrate explicit tag pinning as it's basically inevitable it will break something.

BasPH:
The idea was to show the yaml structure so that users can set images/tags themselves. I'm fine with removing the tag numbers but would add a placeholder to demonstrate where to set tags. Would that work @rob-1126?

jwitz:
Hey @rob-1126 I might go forward with Bas' recommendation here and get this PR out, want to send a quick thought either way? My plan is to format each section like this:
```
prometheus:
  images:
    prometheus:
      repository: 012345678910.dkr.ecr.us-east-1.amazonaws.com/myrepo/myteam-ap-prometheus
      tag: <version-tag>
```

rob-1126:
Revisiting this, over-riding specific image tags is not something image users should be doing - they should always be in lockstep with the chart-defined tags, even if they are moving the images. The root issue is that the option #1 as defined in the doc simply does not work. It's not sufficient. There are currently 9 places most airgapped customers will need to over-ride images at on 0.27 and 7 on 0.28. On 0.28 the following must be over-ridden: 

1. astronomer.houston.config.deployments.helm.airflow.defaultAirflowRepository
2. global.authSidecar.repository (used by most airgapped customers, must include tag; I have a PR open awaiting review in ap-vendor and astronomer to remove the need to override this I'd like in 0.29)
2. astronomer.houston.config.deployments.helm.images.airflow
3. astronomer.houston.config.deployments.helm.images.statsd
4. astronomer.houston.config.deployments.helm.images.redis
5. astronomer.houston.config.deployments.helm.images.pgbouncer
6. astronomer.houston.config.deployments.helm.images.pgbouncerExporter
7. astronomer.houston.config.deployments.helm.images.airflow

and in versions prior to 0.28
8. stan.exporter.image
9. nats.exporter.image

BasPH:
@rob-1126 Added a commit to replace two more pinned versions with placeholders.

BasPH:
@rob-1126 so to summarize: for option 1, we should configure a private registry and _additionally_ configure those 7 or 9 custom images as shown below? Is that correct?

```yaml
global:
  authSidecar:
    repository: [repo url]
astronomer:
  houston:
    config:
      deployments:
        helm:
          defaultAirflowRepository: [repo url]/ap-airflow
          images:
            airflow: [repo url]/ap-airflow
            statsd: [repo url]/ap-statsd-exporter
            redis: [repo url]/ap-redis
            pgbouncer: [repo url]/ap-pgbouncer
            pgbouncerExporter: [repo url]/ap-pgbouncer
```

jwitz:
Closing this out in favor of https://github.com/astronomer/docs/pull/402, which has updated setup based on a discussion @rob-1126 and I had last week.

PR #371, opened 2022-02-07
Update release-notes.md

chomipi88:


cloud/release-notes.md
```diff
@@ -29,9 +29,13 @@ The Astronomer Cloud CLI (`astrocloud`) is now generally available as the offici
...
 For more information on the Astronomer Cloud CLI, see [CLI Release Notes](cli-release-notes.md). For install instructions, read [Install the CLI](install-cli.md).
 
+### Greater Flexibility with User Login
+
+You can now log into your Astronomer Cloud organization using the same email address across GitHub, Google, and Username / Password authentication types, as well as a third-party identity provider (if configured). Regardless of the authentication type used to login, you will only have one user record and associated roles. With this release, we will be enabling GitHub, Google, and Username / Password authentication types for all organizations.
+
 ### Additional Improvements
 
-- Changed the default RDS instance type for new Clusters from `db.r5.xlarge` to `db.r5.large`
+- Changed the default RDS instance type for new Clusters from `db.r5.xlarge` to `db.r5.large`, which represents a monthly cost reduction of ~50%
```
jwitz, line 38: Can we be sure about this? Will all users experience this cost reduction, or is there something else they have to do? Maybe we can use slightly lighter language here, like: `Reduced default maintenance costs by changing the default RDS instance type for new Clusters from `db.r5.xlarge` to `db.r5.large``

jwitz (approved):
Two small comments/ suggestions, otherwise lgtm!

cloud/install-aws.md
```diff
@@ -62,7 +62,7 @@ If you want to continue with the second option, you'll additionally need:
 
 To begin the Astronomer Cloud install process, first create an account at https://cloud.astronomer.io/.
 
-When you first authenticate to Astronomer Cloud, you can sign in with a Google account, a GitHub account, or an email and password. Once you create an account, your authentication method cannot be changed.
```
jwitz, line 65: Jumped in to the branch and removed this sentence. Pretty sure this is the only other place where we mention this behavior

cloud/release-notes.md
chomipi88: Yes the on demand AWS hourly cost is on average 50% cheaper (in some regions a bit more, in others a bit less).
Only newly provisioned clusters would see this cheaper rate. For existing customers, they would have to request a downsize. We would only downsize if there is enough capacity / available overhead to do so.

cloud/install-aws.md
chomipi88: Great find!

paolaperaza (approved):
Thanks both!

cloud/release-notes.md
```diff
@@ -29,9 +29,13 @@ The Astronomer Cloud CLI (`astrocloud`) is now generally available as the offici
 
 For more information on the Astronomer Cloud CLI, see [CLI Release Notes](cli-release-notes.md). For install instructions, read [Install the CLI](install-cli.md).
 
+### Greater Flexibility with User Login
+
+You can now log in to your Astronomer Cloud account with your email across all authentication methods, including GitHub, Google, username/password, and third-party IDP authentication (if configured). Regardless of the authentication type you use to log in, you will have a single user record associated with a single email address. This also means that all organizations now have GitHub, Google, and username/password authentication methods enabled by default for all users.
```
paolaperaza, line 34:
```suggestion
Astronomer now supports multiple authentication methods for a single user account. This means that as long as you're using a consistent email address, you can authenticate with GitHub, Google, username/password, and/or [an external identity provider (idP)](configure-idp.md). Previously, a single user account could only ever authenticate with one method, which could not be changed after the account was created.

This also means that all Organizations now have GitHub, Google, and username/password authentication methods enabled by default for all users.
```

cloud/release-notes.md
```diff
@@ -29,9 +29,13 @@ The Astronomer Cloud CLI (`astrocloud`) is now generally available as the offici
 
 For more information on the Astronomer Cloud CLI, see [CLI Release Notes](cli-release-notes.md). For install instructions, read [Install the CLI](install-cli.md).
 
+### Greater Flexibility with User Login
```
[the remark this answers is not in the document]
chomipi88: Fine with me

cloud/release-notes.md
chomipi88: LGTM

PR #402, opened 2022-02-22
Update image pulling instructions for airgapped installations

jwitz:
Resolves https://github.com/astronomer/docs/issues/319

Updated PR that includes:
- Steps to manually configure images that can't be pulled automatically from global config
- Removal of manual configuration as a default option because of static tag usage


rob-1126 (approved):
I think this represents a solid improvement. Note that there are a few other optional images that don't yet obey the global repo setting but dev is bringing them in line over the next version or so.

PR #432, opened 2022-03-01
Fixes #430 heading font weight

### LINKED ISSUE #430: Fix fonts on headers to be consistent
It doesn't look like our headers in docs are using the same font that body text is using. Should be `Inter`. Let's confirm and change if needed.

image

allenbrady-astro:


jwitz (approved):
Looks great @allenbrady-astro thank you!! Merging now

PR #445, opened 2022-03-02
Fix Algolia search

jwitz:
When we rebranded, there were some leftover variables from our custom components which didn't get updated. This should make the appropriate updates to our search so that it once again filters based on a user's current docset.

### DIFF
--- a/src/theme/SearchBar/README.md
+++ b/src/theme/SearchBar/README.md
@@ -1,0 +1,11 @@
+## NOTE on custom SearchBar component
+
+This component was created using [Docusaurus swizzle eject](https://docusaurus.io/docs/swizzling#ejecting).
+
+This overwrites the SearchBar component so that we can use our custom `useAlgoliaContextualFacetFilters.js` search hook, which filters results based on the product being viewed (Astro vs Software).
+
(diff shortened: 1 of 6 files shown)


jwitz:
@brittanyleigh It doesn't look like my change was enough, at least according to the preview site. Can you look to see if there's anything else that we need to do to restore this feature?

brittanyleigh:
> @brittanyleigh It doesn't look like my change was enough, at least according to the preview site. Can you look to see if there's anything else that we need to do to restore this feature?

Will do!

brittanyleigh:
@jwitz @paolaperaza Ok I believe this did the trick! Check the preview to confirm.

I had to create a copy of our entire SearchBar component so that I could overwrite _**one**_ line to get our custom hook for filtering the results working again, I included a README with details on how/why/etc.

brittanyleigh (approved):

jwitz:
omg 🤦  I knew swizzling this was gonna maybe break in the future, but I didn't realize it would be this bad! Hopefully Docusaurus goes out of beta soon

I love the README, and it's probably something I should be doing with the other custom elements in the site! Thank you so much for this @brittanyleigh you are a proper rockstar

PR #460, opened 2022-03-04
Update REST API link

stuart23:
The URL pointed to in Step 2 is different to the URL pointed to in the Overview. This old URL (https://airflow.apache.org/docs/stable/rest-api-ref.html) actually redirects to the deprecated API doc (https://airflow.apache.org/docs/apache-airflow/stable/deprecated-rest-api-ref.html). This change updates that URL to point to the current API doc.

### DIFF
--- a/astro/airflow-api.md
+++ b/astro/airflow-api.md
@@ -37,7 +37,7 @@ To retrieve your Deployment URL, open your Deployment in the Cloud UI and click 
 
 ## Step 2: Make an Airflow API Request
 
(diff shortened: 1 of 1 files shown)


jwitz (commented):
Nice catch @stuart23, LGTM! Thanks for jumping in to fix it

paolaperaza (approved):
Argh, I actually knew about this one and thought we had fixed it. Thanks @stuart23.

PR #463, opened 2022-03-04
Update cli-quickstart.md (Software)

RNHTTR:
The edited sentence currently reads as if `dags`, `plugins`, and `include` are files and not directories.


paolaperaza (approved):
Oof, good catch @RNHTTR. Thank you! I just committed the same for Astro docs, we had the mistake there too.

Much appreciated! Approving and merging.

RNHTTR:
No prob! It happens :D

PR #484, opened 2022-03-09
remove postgres flag

sunkickr:
The postgres flag is depreciated for now and the triggerer flag is new. Only updated this doc because the astrocloud reference guide appears to up to date


jwitz (approved):
LGTM

PR #493, opened 2022-03-14
Fix file paths in `CHANGELOG.rst`

kaxil:
I had old import paths : `astronomer_operators` -> `from astronomer.providers` for some of the operators


paolaperaza (approved):
Yikes, we should've caught this earlier too 😬  Thanks @kaxil.

PR #496, opened 2022-03-14
Software 0.28.2 and CLI 0.28.1 release notes

jwitz:


software/release-notes.md
```diff
@@ -15,6 +15,23 @@ Astronomer v0.28 is the latest Stable version of Astronomer Software, while v0.2
...
 We're committed to testing all Astronomer Software versions for scale, reliability and security on Amazon EKS, Google GKE and Azure AKS. If you have any questions or an issue to report, don't hesitate to [reach out to us](https://support.astronomer.io).
 
+## v0.28.2
+
+Release date: March 14, 2022
+
+### Additional Improvements
+
+- System Admins can now update the name and description for any Workspace on their installation.
+- You can now configure `global.external_labels` and `remote_write` options in Prometheus Helm chart.
```
danielhoherd, line 25: This one is a bit confusing and we might be able to make more clear. The `global.external_labels` and `remote_write` options are for the prometheus configuration file, not for helm. I think that if we can find a way to point out that these are options in the prometheus config file it will be more clear what's going on here. Maybe something like:

```suggestion
- You can now specify the Prometheus config file's `global.external_labels` and `remote_write` options through the Helm chart.
```

cmarteepants (approved):
Agree with @danielhoherd re Prometheus Config

software_configs/0.28/default.yaml
```diff
@@ -239,7 +253,7 @@ jwt:
   # API.
   #passphrase: ~
 
-  # JWT / Cookie duration in milliseconds.
+  # Duration in milliseconds after user has to be logged out, irrespective of current JWT's expiry status.
```
vishwas-astro, line 256: This comment does not make sense.

Are we saying this is the duration after which a user has to be logged out irrespective of her expiry status?

neel-astro (approved):
changes related to CLI release LGTM 👍. Should we also mention `houston.dial_timeout` to this config list: https://github.com/astronomer/docs/blob/main/software/cli-reference.md#usage-3

software_configs/0.28/default.yaml
jwitz: I took this directly from our current `default.yaml` file. My interpretation is that this setting defines the maximum amount of time that a user can stay logged in regardless of the status of their token, but I'm really not sure. In any case, this would need to be fixed both in docs and on `astronomer/astronomer`.

vishwas-astro: @aliotta @bote795 can one of you review my suggestion here please?

jwitz:
Going to merge now and we can hotfix the `default.yaml` text if needed.

PR #519, opened 2022-03-18
Only mention Helm major required version

BasPH:
Currently, the Software install docs require Helm v3.2.1, but this is an old version. Any Helm 3 version will do, so I suggest only mentioning the major version.


jwitz:
@BasPH The original prerequisite is definitely out of date, but I see that we've been maintaining Helm 3.6 as the minimum required version in our [Version Compatibility Reference](https://docs.astronomer.io/software/version-compatibility-reference). Do you know which of these is more accurate, and depending on the answer can you also update the version compatibility reference and/ or the install guides?

BasPH:
@jwitz don't know the technical reason for minimum v3.6, but I updated the install guide to match v3.6.

jwitz (approved):
Sweet, LGTM!

PR #520, opened 2022-03-18
Second svc acct key is required

aspain:
Two separate passwords need to be provided if using a DEV & Prod instance


jwitz (changes requested):
Great catch @aspain , can't make the suggestion in this PR but could we also update step 2 and 3 before the code fence to say the following? 

```
2. Create two [Deployment-level service accounts](ci-cd.md#step-1-create-a-service-account): One for your Dev Deployment and one for your Production Deployment. 
3. Follow instructions in [GitHub documentation](https://docs.github.com/en/actions/reference/encrypted-secrets) to add your service accounts as secrets to your repository. In the example below, these secrets are named `SERVICE_ACCOUNT_KEY` and `SERVICE_ACCOUNT_KEY_DEV`.
```

Because this change is a full error in the code block, could you also please update the other maintained doc versions (0.23+) in `software_versioned_docs` with this change? Thank you!

aspain:
@jwitz still getting the hang of this, but think that I added your suggested change properly - let me know if this looks good

jwitz (approved):
Looks great! Thank you @aspain !

Issue #518, opened 2022-03-18
Robots.txt Required


astrorfox:
We should publish a robots.txt file for docs.astronomer.io

PR #521, opened 2022-03-18
Create robots.txt

jwitz:
Resolves https://github.com/astronomer/docs/issues/518#event-6266161466

I disallowed some files/directories that I assume we don't want crawlers to access. Are there any other crawlers/ files that we should disallow?

### DIFF
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -1,4 +1,8 @@
 User-agent: *
-Allow: /
+Disallow: /src/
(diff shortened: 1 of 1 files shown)


astrorfox:
Deferring to @allenbrady-astro on the review here

allenbrady-astro (approved):
Looks good! Thanks @astrorfox!

PR #522, opened 2022-03-18
Redirect latest version to stable links

jwitz:
We had a customer report that it was frustrating to not have a versioned link for the latest version of Software (see https://astronomer.zendesk.com/agent/tickets/8583). However, we can't natively support this unless we replace our "stable" linking.

To remedy this, I created redirects for all existing Software docs so users have an option to version them. I am still hesitant to merge this solution because it would require updating these redirects with every minor release of Enterprise. However, I still think this would result in the least maintenance and changed behavior out of all the other options I thought about.


static/_redirects
```diff
@@ -92,6 +92,68 @@
...
+/software/0.28/kubepodoperator-local /software/kubepodoperator-local/ 301
+/software/0.28/kubepodoperator /software/kubepodoperator/ 301
+/software/0.28/kubernetes-executor /software/kubernetes-executor/ 301
+/software/0.28/logs-to-s3 /software/logs-to-s3/ 301
+/software/0.28/manage-airflow-versions /software/manage-airflow-versions/ 301
+/software/0.28/manage-platform-users /software/manage-platform-users/ 301
+/software/0.28/manage-workspaces /software/manage-workspaces/ 301
+/software/0.28/overview /software/overview/ 301
+/software/0.28/platform-alerts /software/platform-alerts/ 301
+/software/0.28/pre-create-namespaces /software/namespace-pools/ 301
```
jwitz, line 141:
```suggestion
/software/0.28/namespace-pools /software/namespace-pools/ 301
```

allenbrady-astro (approved):
Looks good to me! 🙌

astrorfox:
@jwitz agreed on the maintenance burden here...what was the specific customer concern?

jwitz:
@astrorfox I assume that they are maintaining their own internal docs with links to our docs, and they're on 0.28. They don't want to share the stable/current doc links because those will change when a new version is released, and there's no existing `software/0.28/*` links as an alternative.

I will say that this is the first I'm hearing of this use case - most other docsets do not support this option and only maintain a "current" docset (Databricks, Confluent, Docusaurus itself, etc)

astrorfox:
Got it. I would tend to take the feedback, but not act on it at this point. Your references to other projects is helpful there.

jwitz:
Sounds good, I will close this PR for now and will wait until we can approach this problem in a more automated way. 

@allenbrady-astro Do you think it would be possible to pull `docusaurus/plugin-content-docs.versions.current.label` from our Docusaurus config and use that in these redirects? so instead of a redirect starting with `software/0.28/overview`, it would start with `software/{version.envvar}/overview`? This way we'd be able to maintain the version redirects without having to manually update them

allenbrady-astro:
Hey @jwitz I think for the variable to be available for redirects we'll need it available as a [Netlify environment variable](https://docs.netlify.com/configure-builds/environment-variables/). I _think_ we'd also need to set that variable in a netlify.toml file and build the redirects there ([based on this doc](https://docs.netlify.com/configure-builds/file-based-configuration/#inject-environment-variable-values)) since it wouldn't be directly available in the `_redirects` file.

I can take a shot at getting that setup if you'd like! Worse case we might have to set the latest version number both in `docusaurus/plugin-content-docs.versions.current.label` and `netlify.toml` unless we find a way to pull that value and inject it into `netlify.toml` 🤔

jwitz:
If we have to update the version in `netlify.toml`, then we'll still increase our maintenance burden and have to update a value with each release. Let's close this for now @allenbrady-astro and reapproach in the future. Might even open a feature issue with Docusaurus to see if we can specify multiple URL types for the "current" version.

PR #526, opened 2022-03-18
Fix missing parameter and update example dag

pohek321:


jwitz (approved):
Looks great, thanks @pohek321 ! Merging 🚢

PR #529, opened 2022-03-21
Update ci-cd.md

aspain:
This is an update suggestion to the example implementation to follow the diagram we show/suggest in the example CI/CD workflow section: https://docs.astronomer.io/software/ci-cd#example-cicd-workflow

The current example would deploy immediately to prod deployment prior to the PR being merged, which is likely not ideal/intended by users. Instead, this update requires a **merged** PR from dev to main, (simply closing the PR without merging will not trigger a deploy, for example)

### DIFF
--- a/software/ci-cd.md
+++ b/software/ci-cd.md
@@ -188,12 +188,17 @@ The following setup is an example implementation of CI/CD using GitHub Actions. 
 4. Go to the Actions tab of your GitHub repo and create a new action with a `main.yml` file. To achieve the recommended workflow described in [Overview](ci-cd.md#overview), use the following action:
 
     ```yaml
-    name: CI
+    name: Astronomer CI - Deploy Code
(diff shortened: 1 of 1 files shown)


jwitz (approved):
Looks great @aspain , great call on updating this to not push immediately to prod. For my own understanding, does `on.push` also apply to pull requests? I had assumed that the `push` only occurred once the PR was merged. 

Since this is still more of a recommendation than a requirement, I think it's okay to only update the most recent version of this doc. Giving the approve, youre good to merge! 🚢

aspain:
@jwitz if you were to push directly to the main branch from your local, for example, it would immediately kick off the deploy - the PR is not required for the `on push`. You can add more limits with branch protection settings etc, but if nothing else is in place it would deploy prior to PR being created/merged

Issue #475, opened 2022-03-08
New documentation for Astro Deployment Analytics Page


howardyoo:
**Request type:**

- [x] **New document:**: Add the `docs-content` label to this issue.
- [ ] **Addition to an existing document**: Add the `enhancement` label to this issue.

**What is the purpose of the document/enhancement? Select all that apply:**

- [x] Document a new feature for an upcoming release. Add the appropriate release version label to this issue.
- [ ] Information for an existing feature that isn't yet documented.
- [ ] Additional considerations/ reference material for an existing document.

**Problem/feature that requires documentation:**
New Deployment Analytics Page is currently looking for its initial release, probably under https://docs.astronomer.io/astro/deployment-metrics Jus to briefly introduce, the deployment analytics page will provide deeper insights of how the scheduler, DAGs, tasks, and worker nodes are performing in terms of its parsing, CPU, memory usage, and dag run times, etc.

**Are you the right person to review a draft of the document for technical accuracy? If not, who is?**
Yes I am.

howardyoo:
The documentation is ready - it will be published on March 31 (Fri) 2022 - which coincide with the planned release date of deployment analytics page for all the users in PROD environment.

PR #539, opened 2022-03-24
Deployment Analytics doc updates

jwitz:
Resolves https://github.com/astronomer/docs/issues/475

This PR contains all of the doc updates for the new Deployment analytics page. Release notes will be in a separate PR (ignore the branch name). I also used this opportunity to minimize our content on the overview metrics, since these are essentially replicas of what one would fine in the Analytics page.


astro/deployment-metrics.md
```diff
@@ -9,6 +9,82 @@ description: Monitor Deployment performance, health, and total task volume in th
...
+
+    :::info
+    The number of Celery workers per Deployment autoscales based on a combination of worker concurrency and the number of `running` and `queued` tasks, which means that the total available CPU and memory for a single Deployment may change at any given time.
+    :::
+
+- **Network Usage Per Pod (MB)**: This metric charts each worker/ Scheduler Pod's peak network usage over time. Sudden, irregular spikes in this metric should be investigated as a possible error in your project code.
+- **Pod Count per Status**: This metric charts the number of worker/ Scheduler Pods in a given Kubernetes container state. Because Astro operates on a one-container-per-pod model, the state of the container state is also the Pod state. For more information about container states, read the [Kubernetes Documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states).
```
jwitz, line 63: Could definitely use some more insight on what users can understand from this metric

astro/deployment-metrics.md
```diff
@@ -9,6 +9,82 @@ description: Monitor Deployment performance, health, and total task volume in th
 
 The Cloud UI exposes a suite of observability metrics that show real-time data related to the performance and health of your Deployments. These metrics are a useful reference as you troubleshoot issues and can inform how you allocate resources. They can also help you estimate the cost of your Deployments. This document explains each available metric and where to find them.
 
+## Deployment Analytics
+
+Located in the Workspace view of the Astro UI, the **Analytics** page contains a suite of metrics for your Deployments. This page includes metrics that give you insight into the performance of both your data pipelines and infrastructure. Because metrics are pulled from Splunk and Thanos in real time, you can use this page to detect irregularities in your pipelines or infrastructure as they happen.
```
howardyoo, line 14: Maybe we may need to remove the part where the doc mentions about Splunk and Thanos, but rather let's simply say `metrics are collected in real time`. That's because the users do not really need to know from where we are collecting these metrics (and it might change in the future)

astro/deployment-metrics.md
```diff
@@ -9,6 +9,82 @@ description: Monitor Deployment performance, health, and total task volume in th
...
+
+### Workers / Schedulers
+
+These metrics contain information about your worker and Scheduler Pods' resource usage over time. Different worker and Scheduler Pods will appear on these charts as differently colored lines.
+
+![Worker analytics in the Astro UI](/img/docs/analyics-workers.png)
+
+#### Available Metrics
+
+- **CPU Usage Per Pod (%)**: This metric charts a worker's peak CPU usage over a given time interval. The maximum allowed CPUs per Pod as defined in **Worker Resources** appears as a dotted red line. Different worker/ Scheduler Pods will appear on this chart as differently colored lines. This metric should be at or below 50% at any given time.
```
howardyoo, line 55: I'm not sure whether the last sentence `should be at or below 50% at any given time` is correct. Maybe instead we should mention that if the CPU % goes over 90%, the bar color starts to turn from green to red.

howardyoo (commented):
Hey! just did some reviews on the doc - overall, nice job!

astro/deployment-metrics.md
jwitz, line 14:
```suggestion
Located in the Workspace view of the Astro UI, the **Analytics** page contains a suite of metrics for your Deployments. This page includes metrics that give you insight into the performance of both your data pipelines and infrastructure. Because metrics are collected in real time, you can use this page to detect irregularities in your pipelines or infrastructure as they happen.
```

jwitz: You're right, maybe we shouldn't mention it directly in the overview. But I do think it might be worth providing some more technical reference for metrics in the future once we have that info- it seems like the kind of info that a more technical prospective customer would dig through our docs for.

astro/deployment-metrics.md
```diff
@@ -9,6 +9,82 @@ description: Monitor Deployment performance, health, and total task volume in th
 
 The Cloud UI exposes a suite of observability metrics that show real-time data related to the performance and health of your Deployments. These metrics are a useful reference as you troubleshoot issues and can inform how you allocate resources. They can also help you estimate the cost of your Deployments. This document explains each available metric and where to find them.
 
+## Deployment Analytics
+
+Located in the Workspace view of the Astro UI, the **Analytics** page contains a suite of metrics for your Deployments. This page includes metrics that give you insight into the performance of both your data pipelines and infrastructure. Because metrics are pulled from Splunk and Thanos in real time, you can use this page to detect irregularities in your pipelines or infrastructure as they happen.
+
+To view metrics for a given Deployment, click the **Analytics** button in the left-hand menu. From here, you can select a Deployment and time range to view metrics from:
```
jwitz, line 16:
```suggestion
To view metrics for a given Deployment, click the **Analytics** button in the left-hand menu. From here, you can select a Deployment and a time range for your metrics:
```

astro/deployment-metrics.md
jwitz: Took that figure from the Software version of these metrics, can definitely update

jwitz:
Thanks for the quick review @howardyoo ! I should have addressed all comments with my latest commits.

howardyoo:
No, thank you for a very good documentation and quick response! I hope some of the comments were helpful!

PR #558, opened 2022-04-01
update storage task execution tip

chomipi88:
Removed the warning given its of no value considering we don't support ephemeral storage

### DIFF
--- a/astro/resource-reference-aws.md
+++ b/astro/resource-reference-aws.md
@@ -97,7 +97,7 @@ Currently, a single Cluster on Astro cannot be configured with more than one nod
 
 :::tip
 
-With the exception of `m5d` nodes, all suppported node types have a maximum of 20GB of storage per node for system use only. If you need locally attached storage for task execution, we recommend modifying your cluster to run `m5d` nodes, which Astronomer provisions with NVMe SSD volumes out-of-the-box.
+With the exception of `m5d` nodes, all suppported node types have a maximum of 20GB of storage per node for system use only. If you need locally attached storage for task execution, we recommend modifying your cluster to run `m5d` nodes, which Astronomer provisions with NVMe SSD volumes out of the box.
 
 Astronomer plans to support optional ephemeral storage for all node instance types in the first half of 2022.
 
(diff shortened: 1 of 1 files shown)


astro/resource-reference-aws.md
```diff
@@ -97,11 +97,11 @@ Currently, a single Cluster on Astro cannot be configured with more than one nod
 
 :::tip
 
-With the exception of `m5d` nodes, all suppported node types have a maximum of 20GB of storage per node. If you need more than 20GB of storage, we recommend modifying your cluster to run `m5d` nodes, which Astronomer provisions with NVMe SSD volumes out-of-the-box.
+With the exception of `m5d` nodes, all suppported node types have a maximum of 20GB of storage per node for system use only. If you need locally attached storage for task execution, we recommend modifying your cluster to run `m5d` nodes, which Astronomer provisions with NVMe SSD volumes out-of-the-box.
```
jwitz, line 100:
```suggestion
With the exception of `m5d` nodes, all suppported node types have a maximum of 20GB of storage per node for system use only. If you need locally attached storage for task execution, we recommend modifying your cluster to run `m5d` nodes, which Astronomer provisions with NVMe SSD volumes out of the box.
```

jwitz (approved):
One small grammar fix that existed before this PR, otherwise LGTM!

astro/resource-reference-aws.md
paolaperaza: @chomipi88 Why not make the deciding factor here whether the user needs more than 20 GB of storage, and not whether they want locally attached storage or not? I feel like the # of GBs is usually the tipping point? I think I'd say:

```suggestion
With the exception of `m5d` nodes, all suppported node types have a maximum of 20GB of storage per node for system use only. If your tasks require more than 20GB of storage and you want to enable locally attached storage, we recommend modifying your cluster to run `m5d` nodes, which Astronomer provisions with NVMe SSD volumes out-of-the-box.
```

paolaperaza (approved):
Minor wording suggestion here @chomipi88 but LGTM otherwise too - feel free to finish & merge whenever you're ready.

astro/resource-reference-aws.md
chomipi88: @paolaperaza In this case the storage requirement doesn't matter because we don't want them using the system only disk that is attached. Once we support additional ephemeral storage, then that configuration will be driven by the customer.

Issue #504, opened 2022-03-15
Use correct Prism styling on code blocks


jwitz:
**Affected Document:** Many

**Problem Type (Priority):**

- [X] Typo (Low): Any issue related to spelling/ punctuation/ formatting (Add the `error/typo` label to this issue)

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):** Prism includes Docker formatting for Dockerfile. All Dockerfile code fences should use this styling.

Additionally, we sometimes use `py` instead of `python` in a few different Python code blocks. These should also be fixed.

PR #563, opened 2022-04-05
Update prism formatting on Dockerfile and Python blocks

jwitz:
Resolves https://github.com/astronomer/docs/issues/504


paolaperaza (approved):
LGTM. @jwitz I was curious how Dockerfile would render differently and I'm actually not seeing it be different when `dockerfile` is specified, but maybe it's not supposed to 🤷‍♀️ 

Deploy screenshot - 

image

PR #567, opened 2022-04-06
Fix Typos in Hashicorp Vault Secrets Backend Setup

manmeetkaur:
This PR fixes the following:

1. Typo in Dockerfile example for Hashicorp Vault where `variables` should be in double quotes.
2. Fix the spelling of `connections` in Dockerfile
3. Edit the warning to talk about removal of single quotes (`''`) in `KWARGS` not double quotes (`"`)


manmeetkaur:
@paolaperaza @jwitz a prospect faced issue when they copied the content as it is. hence creating this PR :) please advise if anything else needs to be added.

jwitz:
Thank you for opening a PR for this @manmeetkaur , I jumped in and made the same changes to other versions of this setup. I also clarified that you need to remove both `"` _and_ `'`whenever they're wrapping an environment variable in Dockerfile. This should make things a lot more clear for users in the future!

jwitz (approved):

PR #570, opened 2022-04-06
Bump Runtime version to 4.2.4

jwitz:
Removed `4.2.2` and added `4.2.4` to Runtime release notes, as the last three patch versions are really just one version that needed fixes to changes that didn't exist in prior versions. This seems like a good policy for any other "hotfix" versions in the future, as only the latest patch will ever be available in the Cloud UI

However, if there's more than a week between hotfixes, we should consider how to handle situations where someone might be on an "in between" version of Runtime. Should we contact those customers directly, or add a note in release notes like we did with the [Astro CLI](https://docs.astronomer.io/astro/cli-release-notes#v132)?

### DIFF
--- a/astro/release-notes.md
+++ b/astro/release-notes.md
@@ -12,7 +12,7 @@ Astronomer is committed to continuous delivery of both features and bug fixes to
 
 If you have any questions or a bug to report, don't hesitate to reach out to [Astronomer support](https://support.astronomer.io).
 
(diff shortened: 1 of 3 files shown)


paolaperaza (approved):
This LGTM ✅ 

I'm cool to ignore Runtime 4.2.2 + 4.2.3 given they're both unusable

PR #581, opened 2022-04-11
cli release 1.3.4

sunkickr:
Small release to get Colima working for prospects that includes some other small improvements. Colima support is not official yet just cleaned up some issues to get it working for some prospects. We don't want to announce this to customs or create a doc yet. We'll be working on full support after the convergence

### DIFF
--- a/astro/cli-release-notes.md
+++ b/astro/cli-release-notes.md
@@ -17,8 +17,8 @@ Release date: April 11, 2022
 
 ### Additional Improvements
 
-- The astrocloud dev start command should now be slightly faster
-- Astrocloud deploy output not links to the Deployment’s page in the Astronomer UI and its Webserver URL. This gives users easy access to these URLs after a deploy
+- Improved the performance of `astrocloud dev start`
+- When you successfully push code to a Deployment via `astrocloud deploy`, the CLI now provides URLs for accessing the Deployment's Cloud UI and Airflow UI pages.
 
 ## v1.3.3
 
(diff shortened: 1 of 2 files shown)


astro/cli-release-notes.md
```diff
@@ -11,6 +11,15 @@ This document provides a summary of all changes made to the [Astro CLI](install-
 
 If you have any questions or a bug to report, don't hesitate to reach out to us via Slack or Intercom. We're here to help.
 
+## v1.3.4
+
+Release date: April 11, 2022
+
+### Additional Improvements
+
+- The astrocloud dev start command should now be slightly faster
```
jwitz, line 20:
```suggestion
- Improved the performance of `astrocloud dev start`
```

astro/cli-release-notes.md
```diff
@@ -11,6 +11,15 @@ This document provides a summary of all changes made to the [Astro CLI](install-
...
+- Astrocloud deploy output not links to the Deployment’s page in the Astronomer UI and its Webserver URL. This gives users easy access to these URLs after a deploy
```
jwitz, line 21:
```suggestion
- When you successfully push code to a Deployment via `astrocloud deploy`, the CLI now provides URLs for accessing the Deployment's Cloud UI and Airflow UI pages.
```

astro/cli-release-notes.md
jwitz: How noticeable do you think this performance would be to an end user? I'm inclined not to include this in release notes if it's only a minor performance improvement, but if it's anything more than a few seconds then we should keep it. 

In any case, I think we should put the note about the Deployment URLs first, since that has a greater user impact.

jwitz (approved):
Some minor suggestions but otherwise LGTM!

astro/cli-release-notes.md
sunkickr: this really depends on the machine. Probably will b e noticeable on M1

Issue #337, opened 2022-01-28
Add example for how to use CLI/Local Webserver Port Override


aspain:
**Affected Document:**
https://docs.astronomer.io/cloud/develop-project/#override-the-clis-docker-compose-file-local-development-only

**Problem Type (Priority):**

- [ ] Error (Critical): This document has existing information that's incorrect and prevents readers from completing the documented setup (Add the `error/critical` label to this issue)
- [X] Misleading (Important): This document could benefit from additional context, guidance, or clarity (Add the `error/FYI` label to this issue)
- [ ] Typo (Low): Any issue related to spelling/ punctuation/ formatting (Add the `error/typo` label to this issue)

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**

The doc mentions using `docker-compose.override.yml` to modify ports but only shows an example for adding a volume mount.

I found instead that these ports can be overridden in the `.astro/config.yaml` file such as:

(Edit per @paolaperaza 's [link](https://forum.astronomer.io/t/i-already-have-the-ports-that-the-cli-is-trying-to-use-8080-5432-occupied-can-i-change-the-ports-when-starting-a-project/48) with postgres included):
```
project:
  name: my-project-name
webserver:
  port: 8081
postgres:
  port: 5435
```

**Additional Notes:**
Tested this using astrocloud CLI and confirmed webserver worked on port 8081 with this method

paolaperaza:
Notes from conversation with @sunkickr. Let's -

- Add a new section (maybe to local troubleshooting doc) on `config.yaml` file & how to change your port
- Remove port example from docker override doc
- Clarify in docker override section that it doesn't "override", it just adds

These are the full configs you can set in `config.yaml` - 

```
CloudAPIProtocol:  newCfg("cloud.api.protocol", "https"),
		CloudAPIPort:      newCfg("cloud.api.port", "443"),
		CloudWSProtocol:   newCfg("cloud.api.ws_protocol", "wss"),
		CloudAPIToken:     newCfg("cloud.api.token", ""),
		Context:           newCfg("context", ""),
		Contexts:          newCfg("contexts", ""),
		LocalAstro:        newCfg("local.astrohub", "http://localhost:8871/v1"),
		LocalPublicAstro:  newCfg("local.public_astrohub", "http://localhost:8871/graphql"),
		LocalRegistry:     newCfg("local.registry", "localhost:5555"),
		PostgresUser:      newCfg("postgres.user", "postgres"),
[...]

PR #582, opened 2022-04-11
Add info about how to override default ports on the Astro CLI

jwitz:
Resolves https://github.com/astronomer/docs/issues/337

The original issue discussed having this information in "Develop project" but I think we should only include it there once we have more to say about `config.yaml` in general. In the meantime, I think this content is better suited in "Create a Project", since a user would run into this issue when first starting to run a project locally. This could also theoretically live under a new "Troubleshooting" section in "Test and Troubleshoot locally" since that doc is a lot more test-y than troubleshoot-y

### DIFF
--- a/astro/create-project.md
+++ b/astro/create-project.md
@@ -117,7 +117,7 @@ The default credentials are admin:admin
 
 :::info
 
-By default, the Astro CLI uses port `8080` for Airflow's Webserver and port `5432` for Airflow's metadata database. If these ports are already in use on your local machine, you can change the ports for these components using the following steps:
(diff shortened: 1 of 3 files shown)


aspain (approved):
Thanks @jwitz this looks great

paolaperaza (approved):
Sweet, this looks great @jwitz ! After we merge this PR, let's actually:

- Move the details of this workflow to "Test and Troubleshoot" doc once we have a "Common Issues" H2 space there
- Reference ^that section in this Create Project doc exactly where you have it

Can you create a GH issue for ^that? In the meantime, this is a nice/clarifying win ✅

Issue #438, opened 2022-03-01
Add more detail on AWS IAM Role Requirement


paolaperaza:
Based on - https://astronomer.slack.com/archives/C015V2JFKT5/p1646154980152769

@chomipi88 Any chance you can help get this one started in a PR that me and/or @jwitz can review?

PR #583, opened 2022-04-11
Fix AWS instal prerequisites and clarify permissions for AWS CLI usage

jwitz:
Resolves https://github.com/astronomer/docs/issues/438

Added the appropriate base prerequisites for running our CloudFormation script to create a new role (I think!).

Remaining question here is, now that the permissions are different, would we prefer that people create the role using the CloudFormation link or the AWS CLI? Seems like the AWS CLI option is more compelling now because it requires less permissions.

### DIFF
--- a/astro/install-aws.md
+++ b/astro/install-aws.md
@@ -23,7 +23,15 @@ For a complete list of the AWS resources that our team will provision in your AW
 Before completing this setup, make sure that you have:
 
 - A dedicated AWS account with minimum EC2 service quotas.
-- A user with `CreateRole` and `cloudformation:*` privileges.
+- A user with the following permissions:
+    - `cloudformation:*`
+    - `GetRole`
+    - `GetRolePolicy`
+    - `CreateRole`
+    - `DeleteRolePolicy`
+    - `PutRolePolicy`
+    - `ListRoles`
+    - `UpdateAssumeRolePolicy`
 - A subscription to the [Astro Status Page](https://cloud-status.astronomer.io/). This will ensure that you're alerted in the case of an incident or scheduled maintenance.
 
 Astro requires a clean AWS account with a minimum set of EC2 service quotas. For security reasons, the install process is not currently supported on an AWS account that has other tooling running in it. For instructions on creating a new AWS account, follow [AWS documentation](https://aws.amazon.com/premiumsupport/knowledge-center/create-and-activate-aws-account/).
@@ -98,7 +106,7 @@ Then, click the link below to create an [cross-account admin IAM role](https://d
 
 - [Create cross-account IAM Role](https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/quickcreate?templateURL=https://astro-cross-account-role-template.s3.us-east-2.amazonaws.com/customer-account.yaml&stackName=AstroCrossAccountIAMRole&param_AstroAccountId=406882777402)
 
-Alternatively, if you only have `CreateRole` permissions, you can create a cross-account IAM Role by running the following AWS CLI command:
+Using this CloudFormation link is the recommended way to create an IAM role for Astro. Alternatively, you can create a cross-account IAM Role by running the following AWS CLI commands:
 
 ```bash
 $ aws iam create-role --role-name astronomer-remote-management --assume-role-policy-document "{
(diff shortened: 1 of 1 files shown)


astro/install-aws.md
```diff
@@ -23,10 +23,10 @@ For a complete list of the AWS resources that our team will provision in your AW
 Before completing this setup, make sure that you have:
 
 - A dedicated AWS account with minimum EC2 service quotas.
-- A user that has `CreateRole` permissions on that account.
-- Subscribed to the [Astro Status Page](https://cloud-status.astronomer.io/). This will ensure that you're alerted in the case of an incident or scheduled maintenance.
+- A user with `CreateRole` and `cloudformation:*` privileges.
```
jwitz, line 26: @chomipi88 Is there a subset of cloudformation permissions that we can list here instead of the catch-all?

astro/install-aws.md
```diff
@@ -90,15 +90,15 @@ If not specified, Astronomer will create a Cluster with two `m5.xlarge` nodes in
...
+## Step 3: Create a Cross-Account IAM Role for Astro
 
 Once your Astro Cluster has been created, an Astronomer team member will provide you with an [External ID](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html) that will allow Astronomer to connect to your AWS account. Save the External ID as a secret or in an otherwise secure format for use in the AWS CLI.
 
 Then, click the link below to create an [cross-account admin IAM role](https://docs.aws.amazon.com/IAM/latest/UserGuide/getting-started_create-admin-group.html#getting-started_create-admin-group-console) for Astro in your new AWS account:
 
 - [Create cross-account IAM Role](https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/quickcreate?templateURL=https://astro-cross-account-role-template.s3.us-east-2.amazonaws.com/customer-account.yaml&stackName=AstroCrossAccountIAMRole&param_AstroAccountId=406882777402)
 
-Alternatively, run the following AWS CLI command:
+Alternatively, if you only have `CreateRole` permissions, you can create a cross-account IAM Role by running the following AWS CLI command:
```
jwitz, line 101: @chomipi88 Should we promote the CLI option more now that it uses less permissions?

astro/install-aws.md
chomipi88: @dnarain-astro are the above perms vetted?

astro/install-aws.md
chomipi88: I think we should promote the more "safe" method (Stack) as there is much less risk. That said, we can add a note that it requires a bit more permissions to use.

paolaperaza (approved):
LGTM -- will let you and @chomipi88 finish up

astro/install-aws.md
dnarain-astro: here is what we need for the stack method. I did not invest time in breaking down cloudformation to its constituents - we could scope this down a lot further, down to constraining it to a specific YAML file in a specific bucket.  I thought we were reaching the law of diminishing returns on this, but am open to revisiting it.  


`{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "cloudformation:*"
            ],
            "Resource": "*"
        },
        {
            "Effect": "Allow",
            "Action": [
                "iam:GetRole",
                "iam:GetRolePolicy",
                "iam:CreateRole",
                "iam:DeleteRolePolicy",
                "iam:PutRolePolicy",
                "iam:ListRoles",
                "iam:UpdateAssumeRolePolicy"
            ],
            "Resource": "arn:aws:iam::YOUR-AWS-ACCOUNT-ID:role/astronomer-remote-management"
        }
    ]
}`

astro/install-aws.md
dnarain-astro: Here's what we need for the CLI method - CreateRole and AttachRolePolicy.  The policy document is:

`{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": [
                "iam:CreateRole",
                "iam:AttachRolePolicy"
            ],
            "Resource": "*"
        }
    ]
}`

astro/install-aws.md
dnarain-astro: If we don't want to publish the policy document, then the list is:
- cloudformation:*
- GetRole
- GetRolePolicy
- CreateRole
- DeleteRolePolicy
- PutRolePolicy
- ListRoles
- UpdateAssumeRolePolicy

Issue #200, opened 2021-12-14
Algolia docsearch GitHub action


jwitz:
@jwitz commented on [Fri Oct 29 2021](https://github.com/astronomer/docs/issues/505)

PR #616, opened 2022-04-22
Fix Algolia

stuart23:
This action runs when changes are made to main (including PRs merged to main). We could also do this on a [regular schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) if that would be better.

This is working fine when testing on this branch https://github.com/astronomer/docs/actions/runs/2208951112 (but will not run on further changes due to it now only being scoped to pushes to main in the [last commit](https://github.com/astronomer/docs/pull/616/commits/75b908864f4a907ab1e96f16557b1532c3084e20))

Fixes this: https://github.com/astronomer/docs/issues/200

### DIFF
--- a/.github/workflows/reindex.yml
+++ b/.github/workflows/reindex.yml
@@ -1,0 +1,41 @@
+- name: Reindex Algolia
+  on:
+    push:
+      branches:
+        - main
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


paolaperaza (approved):
Approving this! Let's see 🚀  @stuart23 I'll let you do the merging honors

PR #626, opened 2022-04-24
Remove Trailing Slash on Astro

astrorfox:
Adding a trailing slash for consistency of destination to docs.astronomer.io/astro/

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -42,7 +42,7 @@ module.exports = {
       },
       items: [
         {
-          to: 'astro',
+          to: 'astro/',
           label: 'Astro',
           position: 'left',
         },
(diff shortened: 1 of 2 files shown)


paolaperaza (approved):

paolaperaza:
@astrorfox Are we sure we want to have a trailing slash on _every_ docs page? I like it for `/astro/` since there are many child pages from there, but I don't like it on something like `/deploy-code/` since there's no sub page there. At first glance, it doesn't look like these folks do it -

- https://stripe.com/docs/sigma/transactions
- https://docs.github.com/en/get-started/quickstart/create-a-repo
- https://www.apollographql.com/docs/tutorial/schema

astrorfox:
Fair point, let's see if we can go the other way and strip out trailing slashes sufficiently.

paolaperaza:
Recent changes LGTM!

PR #630, opened 2022-04-25
Fix typos

stuart23:
This PR adds the [Typos](https://github.com/crate-ci/typos) GitHub action to run everytime a change is pushed to GitHub. Typos will only check the contents of files, not filenames (I'm looking at you `system-componenets.md`...), and will ignore yarn.lock and css files.

Right now, this workflow is not a check so it will not stop any PRs that do not pass the test, however alerts should still be looked at to stop typos from making it to production.


paolaperaza:
Thank you @stuart23 for this! LGTM. We're gonna hire you for this team soon...

@jwitz Is our resident expert on docs workflows like this so I'm gonna have him take a look at this when he gets back from vacation next week and have him give you the official ✅

jwitz (changes requested):
Thanks for opening this PR @stuart23 ! It's definitely a good instinct to improve the doc workflows in any way that we can. However, we've had spelling and formatting GitHub actions in the past, and we deliberately removed them for a few reasons:

- The false positives are difficult to maintain, especially in our world where there's so many weird proper nouns and company names to include in every doc
- I want PR reviewers, myself included, to get into the habit of reviewing PRs closely, meaning that a human should always be responsible for checking grammar/ spelling before a PR goes out. I sometimes feel like a GitHub action can give a reviewer (or writer) an excuse to not do their own spell checking, and then typos inevitably still slip past both the Action and the reviewer. 

As you can see from the affected files, we've done a pretty good job of catching typos in review thus far, so I would prefer if we didn't implement the action for now. Happy to have you merge the typos in this PR that it caught though!

stuart23:
:white_frowning_face: OK.

I removed the action.

jwitz (approved):
Sorry @stuart23 😢  I love a good GH Action as much as the next committer, but I also want to foster an elite squad of super-editors at Astronomer. This PR LGTM!

PR #633, opened 2022-04-26
`--deployment-id` flag missing from example commands

aspain:


sunkickr:
missing from these examples:  these examples https://docs.astronomer.io/astro/cli-reference/astrocloud-deployment-variable-create#examples

aspain:
I screwed up the branches, will reopen

PR #634, opened 2022-04-26
`--deployment-id` flag missing in examples

aspain:
Examples with missing flag: 
- https://docs.astronomer.io/astro/cli-reference/astrocloud-deployment-variable-list#examples
- https://docs.astronomer.io/astro/cli-reference/astrocloud-deployment-variable-create#examples


paolaperaza (approved):
Thanks @aspain - LGTM. Missed this when we first published because I saw the Deployment ID itself 😮‍💨 

@sunkickr Can you confirm this is right?

sunkickr (approved):
LGTM

PR #642, opened 2022-04-28
Update develop-project.md

### LINKED ISSUE #641: Docs say to use runtime 4.2.10 but the latest on Quay is 4.2.6
Where did 4.2.10 come from?

https://quay.io/repository/astronomer/astro-runtime?tab=tags

stuart23:
Resolves #641

### DIFF
--- a/astro/develop-project.md
+++ b/astro/develop-project.md
@@ -405,7 +405,7 @@ This example assumes that the name of each of your Python packages is identical 
 
 2. In `Dockerfile.build`, add `AS stage` to the `FROM` line which specifies your Runtime image. For example, if you use Runtime ${siteVariables.runtimeVersion}, your `FROM` line would be:
 
-  <pre><code parentName="pre">{`FROM quay.io/astronomer/astro-runtime:${siteVariables.runtimeVersion}-base`}</code></pre>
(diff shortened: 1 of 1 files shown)


paolaperaza:
@stuart23 Thanks for this! Looks like the formatting isn't quite right, seeing the variable show in in plain-text in preview. TBH it might be a bug in Docusaurus?

Screenshots:

image

image

stuart23:
Hey @paolaperaza yeah I'm having a bit of trouble with this, so it's just a draft. Might have to wait until @jwitz gets back because I can't seem to get the templating to work well.

stuart23:
@jwitz welcome back! Could you please give us a hand with templating in the latest version number?

jwitz (commented):
@stuart23 One of the downsides of my variable system is that it doesn't play nice with other inline markdown. You basically have to be able to wrap everything in a `div` or `pre` element or else the whole thing breaks. Because of that, I haven't yet attempted using variables outside of code blocks like you're doing here. 

I should definitely figure out a better env var system down the line, but for now I think this content is fine as-is. We don't make any specific recommendations for using `4.2.10`. Rather, we just say "For example" to show that this is how you would do it with any given version. I imagine that it's just as likely that someone would be configuring these private repos for an earlier version of Runtime, so having a specific version here doesn't change the meaning or the intention of the content. In an ideal world though, we would definitely always have the example Runtime version show the latest!

PR #643, opened 2022-04-29
Set Google Analytics siteSpeedSampleRate

astrorfox:
Increasing from default (1%) to 20% of sessions that will record site speed to get a better picture of performance in Google Analytics.


paolaperaza (approved):
@astrorfox This sounds good - happy to put this in and see what the impact is, though it doesn't look like site checks are passing.

@allenbrady perhaps you can take a look?

PR #681, opened 2022-05-06
Update install-gcp.md

dnarain-astro:
I know that we say Google Cloud Shell is a pre-requisite, but Stride funding still used their command prompt and run into issues because $GOOGLE_CLOUD_PROJECT wasn't set.

### DIFF
--- a/astro/install-gcp.md
+++ b/astro/install-gcp.md
@@ -70,7 +70,7 @@ The Data Plane is a collection of infrastructure components for Astro that run i
 
 To activate the Data Plane on your GCP project:
 
-1. Run the following commands in your Google Cloud Shell:
+1. Run the following commands in your Google Cloud Shell (if you aren't using the Google Cloud Shell, you will need to set $GOOGLE_CLOUD_PROJECT to your Project ID):
 
     ```sh
     $ gcloud services enable storage-component.googleapis.com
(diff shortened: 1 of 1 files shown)


astro/install-gcp.md
```diff
@@ -70,7 +70,7 @@ The Data Plane is a collection of infrastructure components for Astro that run i
 
 To activate the Data Plane on your GCP project:
 
-1. Run the following commands in your Google Cloud Shell:
+1. Run the following commands in your Google Cloud Shell (if you aren't using the Google Cloud Shell, you will need to set $GOOGLE_CLOUD_PROJECT to your Project ID):
```
jwitz, line 73:
```suggestion
1. Run the following commands in your Google Cloud Shell. If you aren't using Google Cloud Shell, replace `$GOOGLE_CLOUD_PROJECT` with your project ID.
```
We say that users need Shell in the prerequisites, then explicitly tell them to use shell here. Is there any reason someone wouldn't use Shell besides preference? If this is the method we recommend, then I don't see a problem with documenting just the Shell setup.

jwitz:
@dnarain-astro Has anyone besides Stride not used Google Shell? Is there a practical reason you wouldn't? I'm leaning towards closing it out, but if you really think this info needs to be here then let's:

- Remove the current changes
- Add a single line to the prerequisites: 

```
- [Google Cloud Shell](https://cloud.google.com/shell). If you don't use Google Cloud Shell, replace every instance of `$GOOGLE_CLOUD_PROJECT` in this setup with your Project ID.
```

dnarain-astro:
No, please close it. 10+ activations down, we haven't seen this again.

Issue #618, opened 2022-04-22
Google shows Software Docs first > Investigate SEO


paolaperaza:
### Problem

When you search Astronomer documentation on Google, the Astronomer Software version of our docs is persistently shown at the top, and the Astro docs are less prominent.

![image](https://user-images.githubusercontent.com/31111752/164772897-0fb9b4b1-39c7-4fd3-9b90-e68610d99715.png)

### Next Step
Let's work with marketing team to look into this.
- Is this because of the longer history those docs have on Software?
- Is there anything we can do with titles or meta descriptions to boost our SEO?

astrorfox:
Before removing /software from Google's index altogether, three recommendations I suggest we take up:
1. Enable [trailingSlash on the site](https://docusaurus.io/docs/api/docusaurus-config#trailing-slash). We're doing this through Netlify optimization anyhow, but will make the Sitemap match reality, which should improve SEO for new pages.
2. Set the [Priority](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-sitemap#priority) for Astro pages to `0.7` and Software pages to `0.3` to indicate to search engines the relative priority between them.
3. Set the [changefreq](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-sitemap#changefreq) to a more realistic value of `daily`.

astrorfox:
Whoops @jwitz!...we (after a fair concern from @paolaperaza) decided to keep trailingSlash to false and optimize for consistency without the trailing slash. Can we set this back to `false`?

I'm also seeing all pages as 0.3 now instead of Astro at 0.7 for priority in the sitemap and still seeing the versioned Software pages.

jwitz:
Yeah I was trying to troubleshoot this without deploying, but it's hard to see how the site actually generates a sitemap without deploying first. It seems like there's some bugs with the Docusaurus site plugin with regards to running multiple docs instances. I've tried a few different configurations and it doesn't seem to recognize the difference between docset ids when generating those priorities. I'll see if this is a proper bug by opening an issue in Docusaurus. 

[...]

PR #684, opened 2022-05-13
Upgrade/ refactor Docusaurus to support new SEO configs

jwitz:
Resolves https://github.com/astronomer/docs/issues/618

Upgrading Docusaurus and tweaking some SEO configs to foreground Astro docs in our search results.


jwitz:
So I'm unable to configure Docusaurus to filter out versioned docs from our sitemap, but I am almost positive that versioned docs in general are unindexed by searches ("Astronomer Software Upgrade to v0.25" doesn't show the doc in google search), so we don't run a huge risk by just keeping it in the sitemap

paolaperaza (approved):
TY @jwitz !

PR #702, opened 2022-05-19
Update resource-reference-gcp.md

dnarain-astro:
We snuck in two regions at the last minute, just updating the docs to reflect this.


jwitz (approved):
LGTM!

PR #724, opened 2022-05-24
Simplified VPC Peering info 

dnarain-astro:


astro/install-gcp.md
```diff
@@ -32,20 +32,6 @@ To install Astro on GCP, you need:
 
 For more information about the resources required to run Astro on GCP, see [GCP Resource Reference](resource-reference-gcp.md).
 
-### VPC Peering Prerequisites (Optional)
```
jwitz, line 35: @dnarain-astro Do you have a sense for how long this installation process takes? The benefit of having this as a prerequisite is that customers are not caught off guard when we request this information in later steps. If removing this means a customer might have to fetch this information mid-install, then I would prefer to keep it as a prerequisite.

jwitz: If anything, I'd keep this detail and simplify the info box later on.

astro/install-gcp.md
```diff
@@ -110,8 +96,7 @@ Once you've activated your Data Plane, provide Astronomer with:
 If you don't specify your organization's preferred configurations, Astronomer will create a Cluster in `us-central1` with default configurations for Astro on GCP. For more information, see [GCP Resource Reference](resource-reference-gcp.md).
 
 :::info
-
-If you need to VPC peer with Astronomer, additionally provide the following information to Astronomer:
+Astro supports Private Services Connect, which allows private consumption of services across VPC networks that belong to different projects or organizations. If you have created custom services that are not published via Private Services Connect, then you may want to peer with Astronomer.  To set up peering, provide the following information to Astronomer:
```
jwitz, line 99:
```suggestion
Astro supports [Private Services Connect](https://cloud.google.com/vpc/docs/private-service-connect), which allows private consumption of services across VPC networks that belong to different projects or organizations. If you have created custom services that are not published using Private Services Connect, then you might want to peer with Astronomer. To set up peering, provide the following information to Astronomer:
```

astro/install-gcp.md
```diff
@@ -110,8 +96,7 @@ Once you've activated your Data Plane, provide Astronomer with:
 If you don't specify your organization's preferred configurations, Astronomer will create a Cluster in `us-central1` with default configurations for Astro on GCP. For more information, see [GCP Resource Reference](resource-reference-gcp.md).
 
 :::info
```
jwitz, line 98:
```suggestion
:::info VPC Peering with Astronomer
```

astro/install-gcp.md
```diff
@@ -110,8 +96,7 @@ Once you've activated your Data Plane, provide Astronomer with:
...
-
-If you need to VPC peer with Astronomer, additionally provide the following information to Astronomer:
+Astro supports Private Services Connect, which allows private consumption of services across VPC networks that belong to different projects or organizations. If you have created custom services that are not published via Private Services Connect, then you may want to peer with Astronomer.  To set up peering, provide the following information to Astronomer:
 
 - VPC Name/ID and region for peering with Astronomer.
 - The IPs of your DNS servers.
```
jwitz, line 102: I can't add a suggestion to this part of the PR, but at the end of this `info` section can you add a note that provides guidance if your VPC peering considerations are more complex than described here? Something like:

"If you have a more complicated connection to external services that is not covered by this setup, reach out to Astronomer Support."

jwitz, line 102: Is this the same thing as the CIDRs in the prerequisites?

astro/install-gcp.md
jwitz, line 35: Any prerequisites we should include here for Private Services Connect?

jwitz (changes requested):
Some final questions/ suggestions here, but I think this is looking good to go soon! For the record, @dnarain-astro and I discussed that "VPC" peering as described here is a pretty rare setup for GCP customers, so we should reprioritize that in the setup.

My main comments relate to two points:
- Do we provide all information here so people can be successful with Private Services Connect? 
- Do we provide some sort of option for anyone connecting in a more complex way, even if that option is just to get in touch with Support?

astro/install-gcp.md
```diff
@@ -109,9 +95,8 @@ Once you've activated your Data Plane, provide Astronomer with:
 
 If you don't specify your organization's preferred configurations, Astronomer will create a Cluster in `us-central1` with default configurations for Astro on GCP. For more information, see [GCP Resource Reference](resource-reference-gcp.md).
 
-:::info
-
-If you need to VPC peer with Astronomer, additionally provide the following information to Astronomer:
+:::info VPC Peering with Astronomer
```
jwitz, line 98:
```suggestion
:::info VPC Peering with Astronomer

```

PR #726, opened 2022-05-24
Update AIRFLOW__SECRETS__BACKEND_KWARGS for Hashicorp

frankbreetz:
There was some missing quotes around variables causing airflow to crash when you set the AIRFLOW__SECRETS__BACKEND_KWARGS environment variable


paolaperaza (approved):

paolaperaza:
😬 Good catch - merging!

PR #736, opened 2022-05-26
Update hasicorp to work with astrocloud cli

frankbreetz:
Currently url parameter of the AIRFLOW__SECRETS__BACKEND_KWARGS is set up to connect with k8s pod. changed to connect with docker container. if someone is running vault locally, as describe in docs it will connect while using the astrocloud cli after change.


paolaperaza (approved):
Nice this SGTM @frankbreetz ! Feel free to merge. It doesn't look like this applies to other providers -- we're good right?

PR #747, opened 2022-05-30
Remove dollar sign from the doc (Astro on GCP)

magdagultekin:
Dollar signs make it impossible to copy the code and paste it as-is in the terminal.

### DIFF
--- a/astro/install-gcp.md
+++ b/astro/install-gcp.md
@@ -73,17 +73,17 @@ To activate the Data Plane on your GCP project:
 1. Run the following commands in your Google Cloud Shell:
 
     ```sh
-    $ gcloud services enable storage-component.googleapis.com
-    $ gcloud services enable storage-api.googleapis.com
-    $ gcloud services enable compute.googleapis.com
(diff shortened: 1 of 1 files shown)


jwitz (changes requested):
This is a great idea @magdagultekin, thanks for the fix! One question before we merge:

If someone copies/ pastes the entire command block (as it is now) into the Google Cloud Shell, will these commands work? We wanted to show that each of these lines is a separate command, hence the `$`. 

If possible, it would be great to format each of these blocks as a single command so that users can fully copy/paste. Otherwise, if users still have to copy/paste line by line, then I think it'd be better to keep the `$` so that we clearly show where each command begins and ends.

magdagultekin:
@jwitz Yeah, it'll work, I just did it myself yesterday after removing $ :)

jwitz (approved):
Ah sweet! Then in that case, we are good to go @magdagultekin 🚀

magdagultekin:
Fantastic @jwitz, thank you! 🔥  (I don't have write access so I can't merge this PR :()

PR #749, opened 2022-05-31
Update create-project.md

magdagultekin:
After adding Triggerer, astrocloud dev start spins up 4 Docker containers, not 3.


jwitz (approved):
LGTM! Thanks for the fix @magdagultekin

PR #770, opened 2022-06-03
Fix activeBaseRegex for all versions

jwitz:
This PR makes our navbar items appear "active" based on what file you're browsing. The current behavior is inaccurate, and this change should fix that.


allenbrady-astro (approved):
LGTM!

PR #777, opened 2022-06-06
Changed astrocloud to astro

magdagultekin:
Updated `astrocloud` to `astro` and the Astro Runtime tag.


jwitz (approved):
Thanks for catching this @magdagultekin ! I jumped in to the PR and fixed some additional instances of `astrocloud` throughout the docs. I also made it so that we no longer have to update the Runtime version number in the code block on the Upgrade Runtime doc.

magdagultekin:
Fantastic, thank you @jwitz!

PR #804, opened 2022-06-13
fix yaml snippet in Option 2

cmarteepants:
This yaml snippet was from the old way, this way is a bit more user-friendly and keeps people out of the lower level configs

### DIFF
--- a/software/namespace-pools.md
+++ b/software/namespace-pools.md
@@ -218,21 +218,21 @@ For every namespace you want to add to a pool, you must create a [namespace](htt
 
 1. Set the following values in your `config.yaml` file, making sure to specify all of the namespaces you created in the `namespaces.names` object:
 
-  ```yaml
-  global:
-    features:
-      namespacePools:
-        # if this is false, everything in this section can be ignored. default should be false
-        enabled: true
-        namespaces:
-          # automatically creates namespace, role and rolebinding for commander if set to true
-          create: false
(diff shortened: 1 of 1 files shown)


software/namespace-pools.md
```diff
@@ -216,35 +216,23 @@ For every namespace you want to add to a pool, you must create a [namespace](htt
...
+        enabled: true
+        namespaces:
+          # automatically creates namespace, role and rolebinding for commander if set to true
+          create: false
+          # this needs to be populated (something other than null) if global.features.namespacePools.enabled is true
+          # add the namespace names that were created in the previous setp
+          names:
+            - <your-namespace-1>
+            - <your-namespace-2>
+  ```
```
jwitz, line 235:
```suggestion
    ```yaml
    global:
      features:
        namespacePools:
          # if this is false, everything in this section can be ignored. default should be false
          enabled: true
          namespaces:
            # automatically creates namespace, role and rolebinding for commander if set to true
            create: false
            # this needs to be populated (something other than null) if global.features.namespacePools.enabled is true
            # add the namespace names that were created in the previous setp
            names:
              - <your-namespace-1>
              - <your-namespace-2>
    ```
```
Indenting by 4 spaces on each line so that the code appears inline with the step. This should render as normal indentation and will be copyable on the website.

jwitz (approved):
Minor suggestion on indentation, otherwise LGTM

PR #820, opened 2022-06-15
Add 5.0.4 release notes

jedcunningham:


astro/runtime-release-notes.md
```diff
@@ -11,6 +11,16 @@ Astro Runtime is a Docker image built and published by Astronomer that extends t
...
 
+## Astro Runtime 5.0.4
+
+- Release date: June 15, 2022
+- Airflow version: 2.3.2
+
+### Additional Improvements
+
+- Update `astronomer-providers` to v1.5.0. For more information, see the [Astronomer Providers Changelog](https://astronomer-providers.readthedocs.io/en/stable/changelog.html#id1).
+- Handle istio-proxy for Kubernetes Pods
```
jwitz, line 22: Will users experience this change in any way? If not, I think we're good to remove it. If they do, can we say a little more about what the implications are here?

jwitz (approved):
One small comment, but LGTM! Also bumped the version number in some other key spots.

astro/runtime-release-notes.md
jedcunningham: No, this just adds support for clusters with Istio.

PR #821, opened 2022-06-15
Use new Docusaurus features to fine-tune sitemap

jwitz:
Resolves https://github.com/astronomer/docs/issues/618

This PR:
- Removes versioned Software docs from our sitemap 
- Removes sitemap configs from Algolia so that it uses recursive crawling for indexing (as opposed to our sitemap, which is now incomplete)

The website broke when I enabled trailingSlash at the config level, and I think this is because Netlify is already processing the trailing slashes. 

If the Algolia change results in lost search functionality, I might have to revert this change. But I've confirmed that the sitemap is generated as expected.


astrorfox (changes requested):
Two issues I'm seeing inline

astrorfox (approved):

jwitz:
@astrorfox All sitemap entries now have a priority of 0.5 and a `changefreq` of `daily`. The same weight between Software and Astro should be okay because Google does not currently consume priority values - we'll have to find another way of optimizing searches so Astro shows first.

PR #840, opened 2022-06-22
Fix yaml indentation

stuart23:


stuart23:
This section only existed in 0.27+

PR #841, opened 2022-06-22
config.yaml editing instructions should be consistent for each page

stuart23:
Most of the pages under **Platform Setup** give instructions in the same format: update config.yaml and then see the instructions in **Apply a Software Platform Configuration Change**.

These two pages are doubling up by also instructing the user how to retrieve the config.yaml, which is differs from the precedence. I would like to make all the change-config-instructions pages consistent.

### DIFF
--- a/software/custom-image-registry.md
+++ b/software/custom-image-registry.md
@@ -59,14 +59,7 @@ Deploying code changes to a custom image registry requires triggering a GraphQL 
 
   :::
 
-2. Locate your `config.yaml` file. To retrieve it programmatically, run:
-
-    ```bash
-    # platform-release-name is usually "astronomer"
-    helm get values <your-platform-release-name> astronomer/astronomer -n astronomer
-    ```
-
-3. Add the following to your `config.yaml` file:
+2. Add the following to your `config.yaml` file:
 
     ```yaml
     astronomer:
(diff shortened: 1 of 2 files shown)


software/export-task-logs.md
```diff
@@ -51,14 +51,7 @@ With this implementation, the Vector sidecars each utilize 100m cpu and 384Mi me
...
 
-1. Locate your Astronomer Software `config.yaml`. To retrieve it programatically, run the following:
-
-    ```bash
-    # platform-release-name is usually "astronomer"
-    helm get values <your-platform-release-name> astronomer/astronomer -n astronomer
-    ```
-
-2. Add the following entry to your `config.yaml:`
+1. Add the following entry to your `config.yaml:`
```
jwitz, line 54:
```suggestion
1. Add the following entry to your `config.yaml` file:
```

software/custom-image-registry.md
```diff
@@ -59,14 +59,7 @@ Deploying code changes to a custom image registry requires triggering a GraphQL
 
   :::
 
-2. Locate your `config.yaml` file. To retrieve it programmatically, run:
```
jwitz, line 62: What do you think about doubling up on linking to the doc? Like:

```
2. Open your `config.yaml` file. See [Apply a Config Change](https://docs.astronomer.io/software/apply-platform-config).
```

I'm hesitant to link this doc only on the later step. What if a user doesn't have their `config.yaml` file and doesn't know where to get it? It's not clear that they need to access the doc in a later step in order to do that.

jwitz (commented):
One minor suggestion that might change the structure a bit here @stuart23 , let me know what you think.

software/custom-image-registry.md
stuart23: Makes sense to me - should we update all the **Platform Setup** instructions to have this same format of link-code snippet-link?

jwitz: Let's start with the ones that you included in this PR. When I touch other docs that have a similar setup, I'll update them accordingly - I imagine the implementation will be slightly different for each one, plus I notice that don't mention the Helm command for most other Platform Setup docs.

PR #860, opened 2022-06-28
Fix max AU for m5.xlarge instance type

afastronomer:


jwitz (approved):
✅ thank you @afastronomer !

PR #868, opened 2022-06-29
Make font sizes a little smaller

allenbrady-astro:
Fixes #863 by tweaking the header fonts sizes to be a little smaller.

Before
Screen Shot 2022-06-29 at 3 54 41 PM

After
Screen Shot 2022-06-29 at 3 54 30 PM


jwitz (approved):
Looks great! Perfect amount of adjustment, just enough to make a difference but not enough to be disruptive. Thanks @allenbrady-astro !

PR #876, opened 2022-07-01
Archive Software 0.26 and 0.27

jwitz:
Software 0.26 and 0.27 are no longer maintained. Per our policies, we should archive the docsets for those versions.


cmarteepants (approved):
🚢

PR #890, opened 2022-07-06
fix 'should should' typo

sunkickr:
I found a typo in a few of the docs


jwitz (approved):
LGTM! Thanks for catching @sunkickr

PR #902, opened 2022-07-12
add deployment variable in the environment

manmeetkaur:
Hello Docs Team, the details for this PR are as below:

**Problem Statement:** A prospect was referring the docs for CICD to get it working on their end. But they were running into error - ```Using an Astro API key
 CONTEXT DOMAIN                      WORKSPACE                           
 astronomer.io                                                           
 Switched context
Authenticated to Astro 
Error: Insufficient permissions.
Error: Process completed with exit code 1. ```
The reason was no `DEPLOYMENT_ID` was set in the environment variable or mentioned in the `astro deploy` command. 

**Changes:** Have made the change to set it up in the Environment Variables. Not familiar with CircleCI so have not touched it, but that also needs to be fixed. I have tested on my end with GitHub Actions that we set up Deployment_id in environment variable it works even when we do not give `DEPLOYMENT_ID` with the `astro deploy` command.

### DIFF
--- a/astro/ci-cd.md
+++ b/astro/ci-cd.md
@@ -72,6 +72,7 @@ To automate code deploys to a Deployment using [GitHub Actions](https://github.c
 
    - `ASTRONOMER_KEY_ID` = `<your-key-id>`
    - `ASTRONOMER_KEY_SECRET` = `<your-key-secret>`
+   - `DEPLOYMENT_ID` = `<your-deployment-id>`
 
 2. In your project repository, create a new YAML file in `.github/workflows` that includes the following configuration:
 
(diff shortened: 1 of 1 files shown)


jwitz:
@manmeetkaur What version of the CLI was the prospect using? In later versions, the CLI _should_ be able to determine which Deployment to access based only on the Deployment API key environment variables. See this release note for more information: https://docs.astronomer.io/astro/cli/release-notes#easier-way-to-determine-deployment-id-on-deployment-commands

We try to maintain the CI/CD templates so that they always work with the latest version of the Astro CLI, which does mean that they are sometimes incompatible with earlier CLI versions

manmeetkaur:
@jwitz the prospect was using the latest CLI version 1.3

jwitz:
Hmmm. I wonder if this is a bug with the CLI rather than a documentation inaccuracy. @sunkickr Shouldn't these pipelines work without having to specify a Deployment ID?

paolaperaza:
Hey folks -- just talked to @sunkickr and gonna close this out. In conclusion:

- The user here had a reference to a Deployment ID environment variable on GitHub actions that did not exist
- They knew that their Deployment ID is not needed and so they deleted the actual env var for it (good) but forgot to delete the _reference_ to it in their script, which is why they were seeing an error
- In theory this would've worked if the Env Var _did_ exist, but again -- per recent product changes, it's not needed. If they have it, users need to delete both the env var + reference to it

Gonna close this out ✅

PR #921, opened 2022-07-18
removeContradictoryExampleYaml

aliotta:
Description: the example yaml in these sections makes it seem like implicit is a valid value when using teams when it is not.


jwitz:
@aliotta This doc isn't directly related to configuring Teams, so I don't think it's necessary to limit the notes/ values to only what's allowed for Teams.

PR #924, opened 2022-07-18
Software 0.29.2 release notes

jwitz:
Based on opened and closed issues in https://github.com/astronomer/issues/labels/0.29.2

### DIFF
--- a/software/configure-deployment.md
+++ b/software/configure-deployment.md
@@ -98,44 +98,6 @@ Airflow 2.2 introduces the triggerer, which is a component for running tasks wit
 
 By adjusting the **Triggerer** slider in the Software UI, you can provision up to 2 triggerers on any Deployment running Airflow 2.2+. To take advantage of the Triggerer's high availability, we recommend provisioning 2 triggerers for production Deployments.
 
-Note that this feature must first be enabled by a System Admin before it appears in your Deployments. To enable triggerers, follow the steps in the following section.
-
-#### Enable triggerers
-
-Triggerers are available only in Deployments running Airflow 2.2+. Additionally, the feature must be explicitly enabled on your platform by a user with System Admin permissions. To enable the feature, update your `config.yaml` file with the following values:
-
-```yaml
-astronomer:
-  houston:
-    config:
-      deployments:
-        triggererEnabled: true
-```
-
-If you have overridden  `astronomer.houston.config.deployments.components`, you additionally need to add the following configuration:
-
-```yaml
-astronomer:
-  houston:
-    config:
-      deployments:
-        components:
-          # add this block to the other components you've overridden
-          - name: triggerer
-            au:
-              default: 5
-              limit: 30
-              request: ~
-            extra:
-              - name: replicas
-                default: 0
-                minimum: 0
-                limit: 2
-                minAirflowVersion: "2.2.0"    
-```
-
-After you save these changes, push your `config.yaml` file to your installation as described in [Apply a config change](apply-platform-config.md).
-
 ## Kubernetes executor: Set extra capacity
 
 On Astronomer, resources required for the [KubernetesPodOperator](kubepodoperator.md) or the [Kubernetes Executor](kubernetes-executor.md) are set as **Extra Capacity**.
--- a/software/release-notes.md
+++ b/software/release-notes.md
@@ -20,6 +20,7 @@ Release date: July 12, 2022
 
 ### Additional improvements
 
+- You can now configure extra sidecar containers and volume mounts for the Houston API and Commander using `astronomer.houston.extraContainers` and `astronomer.commander.extraContainers` in your `config.yaml` file.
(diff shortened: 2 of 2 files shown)


software/release-notes.md
```diff
@@ -14,6 +14,28 @@ This document includes all release notes for Astronomer Software version 0.29.
 
 We're committed to testing all Astronomer Software versions for scale, reliability and security on Amazon EKS, Google GKE and Azure AKS. If you have any questions or an issue to report, don't hesitate to [reach out to us](https://support.astronomer.io).
 
+## v0.29.2
+
+Release date: July 18, 2022
+
+### Additional improvements
+
+- You can now configure extra sidecar containers and volume mounts for the Houston API and Commander using `astronomer.houston.extraContainers` and `astronomer.commander.extraContainers` in your `config.yaml` file.
```
jwitz, line 23: I didn't know how much we wanted to expose/ document this feature, is this ready for broader use and should we make a bigger deal out of it?

cmarteepants (approved):

software/configure-deployment.md
```diff
@@ -98,44 +98,6 @@ Airflow 2.2 introduces the triggerer, which is a component for running tasks wit
...
-            extra:
-              - name: replicas
-                default: 0
-                minimum: 0
-                limit: 2
-                minAirflowVersion: "2.2.0"    
-```
-
-After you save these changes, push your `config.yaml` file to your installation as described in [Apply a config change](apply-platform-config.md).
-
```
vishwas-astro, line 138: Why are we removing the section above?

software/release-notes.md
vishwas-astro: We can limit exposure to JPMC for now. But good point, we should soon add docs around all these abilities that we are adding to make our software offering lighter weight and more pluggable with existing customer ecosystems.
cc: @cmarteepants

software/release-notes.md
```diff
@@ -14,6 +14,28 @@ This document includes all release notes for Astronomer Software version 0.29.
...
+- You can now configure Vector for use with [sidecar logging](export-task-logs.md#export-logs-using-container-sidecars).
```
vishwas-astro, line 24: configure to do what?

software/release-notes.md
```diff
@@ -14,6 +14,28 @@ This document includes all release notes for Astronomer Software version 0.29.
...
+- Resolved several high and critical CVEs.
+- You can now assign System Viewers and System Editors to [Teams](import-idp-groups.md).
```
vishwas-astro, line 26: @cmarteepants was it just for teams or for users as well?

software/release-notes.md
```diff
@@ -14,6 +14,28 @@ This document includes all release notes for Astronomer Software version 0.29.
...
+
+### Bug fixes
+
+- If you have `customLogging.enabled=true` and `loggingSidecar.customConfig=false` in your Helm configuration, logs now appear in the Software UI as expected.
+- System Admins can no longer update their own role.
+- The Software UI no longer counts inactive users in its user count figures.
+- Fixed an issue where you could still access a Deployment using a URL after logging out of the Software UI.
+- Fixed an issue where you could view Deployment information from a Workspace that was deleted with `astro workspace delete`.
+- Fixed an issue where you could not open Celery from the Software UI.
+- Fixed an issue where logging out of the Software UI would occasionally fail.
```
vishwas-astro, line 36: Are the fixes for this and the other logout issue above the same? Should this be combined from even a bug fix perspective into a single line item?

software/configure-deployment.md
jwitz: Triggerers are now enabled by default, so we no longer need this section in documentation. I imagine disabling/ enabling triggerers will become a more niche configuration now that it's an accepted part of Airflow.

software/release-notes.md
cmarteepants: For line 23 (the sidecar injection feature for Houston and Commander), I agree with being vague - I am hoping the only  ones using it will be JPMC. This is generally an infra-level concern, and something I wish clients would manage using their own standard tools (logstash/fluentd/datadog agent)

For line 24, we can say this: `You can configure Vector to send Airflow Task logs to additional log management systems`. I'm torn on whether we should document this loudly, or not. Same as above, I don't love that people use our log shipping tool to forward Airflow Task logs because it moves us away from being lightweight and increases client reliance on our infra (our fluentd, and now our vector components). Difference is that clients are used to because of some early features, like use Fluentd to [forward logs to S3](https://docs.astronomer.io/software/logs-to-s3) and is something they've come to expect.

software/release-notes.md
jwitz: Just saw that the issue was already addressed in an earlier version, according to this discussion: https://github.com/astronomer/issues/issues/4785

Removing now

vishwas-astro (approved):

software/release-notes.md
cmarteepants: @jwitz I don't recall that you could before, at least not via the UI. It's possible that you could have via the API.

PR #942, opened 2022-07-21
Setting astronomer namespace context for  clusterissuer.yaml file

deepaksharma23:
we didn't had namespace for this command  kubectl apply -f clusterissuer.yaml


jwitz (approved):
Looks good @deepaksharma23 ! I also made this change other supported versions of Software docs.

PR #956, opened 2022-07-25
Add Pypi Binaries to Path in Private Repo Doc

fritz-astronomer:
A customer had an issue - they had modified their Dockerfile like this, and couldn't access a CLI installed via `pip`. This fix copied the binaries into the final stage to make them accessible


jwitz (approved):
Does this same change need to be made in `software/custom-image-registry.md`? If so we should also make the change there, but otherwise LGTM!

fritz-astronomer:
updated

jwitz:
Lookin good 🚢

Issue #958, opened 2022-07-26
Create a custom admonition for CLI callouts


jwitz:
It's undocumented, but Docusaurus is ready to handle custom admonition types via swizzling as described in https://github.com/facebook/docusaurus/pull/7152

We should create one for all of the times we call out parallel CLI functionality in our docs.

PR #959, opened 2022-07-26
Create a custom CLI admonition

jwitz:
Resolves https://github.com/astronomer/docs/issues/958

See the "Create a Deployment" doc for an example of how this would look. To do still: 

- Connect color to admonition
- Find an SVG icon to use that's specific for the CLI


src/css/custom.css
```diff
@@ -98,15 +98,20 @@
...
+  --ifm-color-warning-dark: var(--orange-500);
 
-  /* --ifm-color-danger: var(--orange-400);
+  --ifm-color-danger: var(--orange-400);
   --ifm-color-danger-contrast-background: var(--orange-100);
   --ifm-color-danger-contrast-foreground: var(--orange-900);
-  --ifm-color-danger-dark: var(--orange-500); */
+  --ifm-color-danger-dark: var(--orange-500);
+
+  --ifm-color-cli: var(--purple-400);
```
jwitz, line 111: @allenbrady-astro I tried copying the color formatting exactly as it appears for other note types, but I still can't see the purple color in the note itself. Is there another file I need to touch to make this `cli` class official?

src/theme/Admonition/index.js
```diff
@@ -0,0 +1,198 @@
...
+      />
+    </svg>
+  );
+}
+function CliIcon() {
+  return (
+    <svg viewBox="0 0 14 16">
+      <path
+        fillRule="evenodd"
+        d="M6.3 5.69a.942.942 0 0 1-.28-.7c0-.28.09-.52.28-.7.19-.18.42-.28.7-.28.28 0 .52.09.7.28.18.19.28.42.28.7 0 .28-.09.52-.28.7a1 1 0 0 1-.7.3c-.28 0-.52-.11-.7-.3zM8 7.99c-.02-.25-.11-.48-.31-.69-.2-.19-.42-.3-.69-.31H6c-.27.02-.48.13-.69.31-.2.2-.3.44-.31.69h1v3c.02.27.11.5.31.69.2.2.42.31.69.31h1c.27 0 .48-.11.69-.31.2-.19.3-.42.31-.69H8V7.98v.01zM7 2.3c-3.14 0-5.7 2.54-5.7 5.68 0 3.14 2.56 5.7 5.7 5.7s5.7-2.55 5.7-5.7c0-3.15-2.56-5.69-5.7-5.69v.01zM7 .98c3.86 0 7 3.14 7 7s-3.14 7-7 7-7-3.12-7-7 3.14-7 7-7z"
```
jwitz, line 21: @allenbrady-astro I copied the icon from the NoteIcon since we don't use it in docs, but it would be great if we had a design-approved CLI icon to use here. Maybe something like https://thenounproject.com/browse/icons/term/terminal/?iconspage=1?

src/css/custom.css
allenbrady-astro: Hey Jake! Looks like the `.alert--cli` class was missing just a few other custom properties. Got those added though with the latest commit and should be working now! 🙌 

image

allenbrady-astro (approved):

kwcanuck (approved):
@jwitz it's great that you were able to get this done so quickly. My only concern is that the admonition is visually obtrusive. If there are a number of these admonitions on a single page, it could be _very_ distracting.  If there is any way to make this admonition less obtrusive (transparent, remove color, a simple sidebar) you might want to implement it.

jwitz:
@kwcanuck I'm not sure it's possible to make this any less obtrusive without defeating the purpose of the note itself. I'd only expect one or two of these notes to appear on a doc page at any given time, and the color is unobtrusive and non-urgent. Now that we have a colored note, we could move this to the bottom of topics instead of the top to make it less obtrusive? Users can look for the colored note instead of hoping to find that information later on in a process, meaning that we no longer have to foreground the option like we've been doing without a custom note.

kwcanuck:
@jwitz ideally, admonitions should only be used to convey critical or significant information. Informing users that they can use an alternate methodology (the CLI) to complete a task is not critical information in my opinion. It's informative. Nice to know, but not imperative. This information should be conveyed using a snippet and not an admonition. Snippets are small, reusable chunks of text. [Here](https://redocly.com/docs/developer-portal/guides/reusing-content/#:~:text=Create%20%22chunks%22%20of%20text%2C,target%20pages%20as%20you%20want.) is one example of how to implement snippets in markdown.

As for the location of this information, it should always appear where it's relevant.  If an alternate methodology is available to a user, they should be informed of this _**before**_ beginning a task.  Randomly placing this information at the bottom of a topic isn't recommended because it would lack context. 

I have spent a lot of time at different companies decluttering documentation. One of the most troublesome issues was the overuse of admonitions.

jwitz:
> ideally, admonitions should only be used to convey critical or significant information. Informing users that they can use an alternate methodology (the CLI) to complete a task is not critical information in my opinion.

I think that admonitions, especially with color, can also convey that a piece of information is non-critical and can be skipped over. If UI-first users see these notes, they can comfortably gloss over the information. If a CLI-first user sees these notes, they know that the information is relevant to them. Agreed that excessive notes become an issue in docs, but IMO that issue mainly stems from users having to constantly decide whether a new note is relevant to them or not. Color coding notes this way removes that cognitive burden.

The doc you linked is only for using Markdown in Redocly, I think Markdown is still pretty rough for reusing content in its base format. In any case, though, I'd like to use unique text in these notes so that we can link directly to parts of the CLI reference guide for the commands we're referencing. We can still locate all notes and replace content using the `:::cli` tag.

kwcanuck:
@jwitz you need to be careful relying on color to convey information. From the _[Google developer documentation style guide](https://developers.google.com/style/accessibility?hl=en)_:

_Don't use color, size, location, or other visual cues as the primary way of communicating information._

Also, from the Microsoft Style Guide:

_Choose colors and patterns carefully. High contrast may improve readability for people who have low vision. For people with some types of color blindness, certain color combinations are difficult to distinguish._

Also:

_Don’t convey information with color alone_

astro/create-deployment.md
```diff
@@ -17,7 +17,9 @@ Every Deployment is hosted on a single Astro cluster with its own dedicated reso
 
 ## Create a Deployment
 
+:::cli
 If you prefer, you can run the `astro deployment create` command in the Astro CLI to create a Deployment. See [CLI Command Reference](cli/astro-deployment-create.md).
```
paolaperaza, line 21:
```suggestion
If you prefer, you can also run the `astro deployment create` command in the Astro CLI to create a Deployment. See [CLI Command Reference](cli/astro-deployment-create.md).
```

"also" to indicate the relativity of these two options?

paolaperaza (approved):
@jwitz Let's go! This is cool ✅  Can you create two GH issues?

1. Add a terminal icon to this
2. Add this admonition in other places where CLI functionality is available in addition to Cloud UI steps

Issue #933, opened 2022-07-20
Implement index pages on some doc categories


jwitz:
We can apply Docusaurus' index page styling on select submenus on our docs sidebars. This could be useful in situations where we don't want people to have to navigate ambiguously titled docs just based on their sidebar titles alone. 

Here's how the implementation would work:  https://docusaurus.io/docs/sidebar/items#generated-index-page

I think the best places to apply this menu are:

- Get started
- Reference (entire tab)
- Cloud configuration reference
- Software administration submenus

PR #962, opened 2022-07-26
Add index pages to some sidebar categories

jwitz:
Resolves https://github.com/astronomer/docs/issues/933

This PR is conservative in where we add index pages. I want to try it out on some menus first, and if we like it there then we can try it on other menus. To me, it made the most sense to add index pages for all menus that contain different cloud docs. This way, we can link to the index page instead of having to list 3 links for each cloud whenever we mention those details in our docs.


sidebarsAstro.js
```diff
@@ -84,6 +84,11 @@ module.exports = {
         {
           type: 'category',
           label: 'Install Astro',
+          link: {
+            type:'generated-index',
+            title: 'Install Astro',
+            description: 'Install the Astro platform on your cloud.'
```
kwcanuck, line 90:
```suggestion
            description: 'Install the Astro platform in your cloud.'
```

sidebarsAstro.js
```diff
@@ -129,6 +134,12 @@ module.exports = {
         {
           type: 'category',
           label: 'Cloud configuration reference',
+          link: {
+            type:'generated-index',
+            title: 'Cloud configuration reference',
+            description: 'Learn about all default and supported Astro cloud configurations.'
```
kwcanuck, line 140:
```suggestion
            description: 'Learn about the default and supported Astro cloud configurations.'
```

kwcanuck (approved):

jwitz:
@paolaperaza @kwcanuck This is ready for review again, but on a different front: I was able to load custom properties from the docs so that they use custom icons in their index cards! I imagine that we'll be doing this for all cloud-defined index menus in the future. 

Screen Shot 2022-08-24 at 10 23 33 AM

@katiebrady-astro Adding you for review to see if there's anything I can do to optimize this on the React/ web front. I commented the line in `DocCard/index.js` that I wrote to cause this behavior. Would it be better to use `.svg` for the cloud logos, or is png alright?

kwcanuck:
@jwitz cool stuff! As we are now using third-party logos, I would recommend the addition of an attribution statement to the footer or our docs. Here is an example:

image

The attribution statement formally recognizes ownership of trademarks and logos owned by external organizations. This could potentially save Astronomer some legal hassle.

katiebrady-astro:
Hey @jwitz looks great! 🙌

SVGs can look sharper in multiple settings, but I think the PNGs you've got are already so small that those should work well. I also went ahead and losslessly optimized them with ImageOptim so they're _even smaller_ in size. 😄

PR #964, opened 2022-07-26
Fix link to software version compatibility reference

BasPH:
This PR points the link "Software Version compatibility reference" in https://docs.astronomer.io/software/release-lifecycle-policy to the correct page.


jwitz (approved):
Looks good, thanks for the fix Bas!

PR #978, opened 2022-07-29
Dark mode CLI color

### LINKED ISSUE #976: CLI Admonition doesn't work in night mode
The new CLI admonition keeps the same colors when you switch to night mode despite my attempt to configure an alternative HTML theme. @allenbrady-astro would you be able to help with troubleshooting this?

allenbrady-astro:
Fixes #976

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -427,8 +427,8 @@ html[data-theme='dark'] .alert--cli {
     var(--ifm-color-cli-dark);
     var(15%),
     );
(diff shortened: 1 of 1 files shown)


paolaperaza (approved):
This looks good @allenbrady-astro ! Thank you. One thing I noticed, unrelated to this PR:

The dark mode toggle in light mode isn't actually visible? It comes up in white on dark mode, but we should prob increase contrast there -

image

image

PR #981, opened 2022-08-01
Fix Dark Mode Toggle Color

### LINKED ISSUE #980: Can't see toggle for dark mode in light mode
The dark mode toggle in light mode isn't actually visible. It does come up in white on dark mode, but we should prob increase contrast there 

181973873-34e8b15f-f02d-472b-adcd-8fe6da22fcb2

181974449-6370637d-7364-4ce1-9916-7cafe5e253be

allenbrady-astro:
Fixes #980

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -291,6 +291,14 @@ small {
   font-size: 24px;
 }
 
+.navbar__items--right button.clean-btn {
+  color: white;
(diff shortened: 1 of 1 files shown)


paolaperaza (approved):
Perfect, TY @allenbrady-astro !

PR #998, opened 2022-08-04
Fix GraphQL Params

fritz-astronomer:
Customer reports `runtimeVersion` works (they are using the runtime image) not `airflowVersion`

Can I get confirmation from the Houston team?


kwcanuck (approved):
I recommend awaiting confirmation from the Houston team before merging the changes.

fritz-astronomer:
@andriisoldatenko ? Not sure who else is the Houston team

andriisoldatenko:
code LGTM. Do you want me to validate it?

andriisoldatenko:
> @andriisoldatenko More that that's the right change, and won't break if people aren't using the runtime, I suppose? Are both params valid? Only one? When was it changed?

Correct both params are still valid.  
release 0.29.0.

PR #1000, opened 2022-08-05
we are supporting Azure too

nandlalyadav57:
making changes as per conversation in slack recently here https://astronomer.slack.com/archives/C015V2JFKT5/p1659658484241529


kwcanuck (approved):
@nandlalyadav57 while reviewing the change, I made a few other edits. Thanks for catching this!

PR #1013, opened 2022-08-08
changing "pmm" --> "prod"

aspain:


paolaperaza (approved):
Good catch here, @aspain !

PR #1022, opened 2022-08-10
Update BYO Registry Docs

fritz-astronomer:
image might need to be named after the runtime image? Confirming


fritz-astronomer:
doesn't appear to be an issue

PR #1045, opened 2022-08-17
Clarify API key access

jwitz:
Based on https://astronomer.slack.com/archives/C01UA03TPV1/p1660747961908799

### DIFF
--- a/astro/api-keys.md
+++ b/astro/api-keys.md
@@ -30,8 +30,10 @@ To create an API key for a Deployment:
 3. Give the key a name and description, then click **Create API Key**:
 
     ![Create API key button](/img/docs/create-api-key.png)
-    
+
 From here, you can copy the API key ID and secret for use in API calls and CI/CD pipelines. Make sure to save the key secret securely, as this is the only time you will have access to see it in plain text.
+
+Any user or service with access to the API key and secret can access your Deployment. The only way to delete this access is to [delete the API key](api-keys.md#delete-an-api-key) or [delete the Deployment](configure-deployment-resources.md#delete-a-deployment).
 
 :::tip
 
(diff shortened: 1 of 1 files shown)


paolaperaza (approved):
Heya @jwitz -- I just went in here and made some changes since the part of the file I added hadn't been modified.

I was thinking:

- I think Tony's main question was around "if the user that creates it leaves the Workspace, what happens with the API key?". I didn't think we totally answered the Q so added a line item to the top
- The other statement you added I kinda thought belonged at the top too?
- Added a reminder that API keys can't be recovered once deleting them too

Will let you polish off here!

astro/api-keys.md
```diff
@@ -52,17 +54,19 @@ To use API keys with the Astro CLI, you must make your Deployment API key ID and
...
 export ASTRONOMER_KEY_ID=<your-key-id>
 export ASTRONOMER_KEY_SECRET=<your-key-secret>
 ```
 
-After setting the variables, running `astro deployment update` works for the Deployment and you don't need to manually authenticate to Astronomer. Astronomer recommends storing `ASTRONOMER_KEY_SECRET` as a secret before using it to programmatically update production-level Deployments.
+After setting the variables, running `astro deployment update` works for the Deployment and you don't need to manually authenticate to Astronomer. Astronomer recommends storing `ASTRONOMER_KEY_SECRET` as a secret before using it to programmatically update a production Deployment.
 
 ## Delete an API key
 
+Once deleted, an API key and secret cannot be recovered. If you accidentally delete an API key, you must create a new one.
```
paolaperaza, line 68:
```suggestion
If you delete an API key, make sure that no existing CI/CD pipelines are using that key ID and secret. Once deleted, an API key and secret cannot be recovered. If you unintentionally delete an API key, create a new one and update any CI/CD workflows that used the deleted API key.
```

Opp for better phrasing here ,but point is "if you delete this thing make sure your ci/cd pipelines using it"

Issue #974, opened 2022-07-29
Add "terminal" icon to CLI admonition


jwitz:
Our `:::cli` callout would benefit from a terminal icon instead of a generic info icon

PR #1059, opened 2022-08-23
Use new icon for CLI admonitions

jwitz:
Purchased in https://thenounproject.com/search/icons/?iconspage=1&q=terminal, I thought this one looked the best in the admonition but open to try other options!

Resolved https://github.com/astronomer/docs/issues/974


kwcanuck (approved):
Looks good.  For anyone looking for an example, this is how it appears:

image

For context, see [Create a Deployment](https://deploy-preview-1059--gallant-neumann-56599d.netlify.app/astro/create-deployment). My concern remains that these statements are visually obtrusive In my opinion, these statements don't need to be so prominent.

PR #1066, opened 2022-08-23
add tests to Software CI/CD docs

dintorf:


dintorf:
> Great addition @dintorf ! I just made some minor copy edits but otherwise this LGTM. Do you think I should add this to the nebula docs as well?

I didn't know there were still Nebula docs 😆 IMO I don't think it's necessary to add the changes there.

software/ci-cd.md
```diff
@@ -170,6 +170,16 @@ docker build -t registry.${BASE_DOMAIN}/${RELEASE_NAME}/airflow:ci-${DRONE_BUILD
 
 If you would like to see a more complete working example please visit our [full example using Drone-CI](https://github.com/astronomer/airflow-example-dags/blob/main/.drone.yml).
 
+### Run unit tests
+
+Astronomer recommends adding a unit testing step to production pipelines after your build step to ensure that you don't push an image with breaking changes to your registry. To run basic unit tests, create a pipeline step to execute `docker run` and run `pytest tests` in a container based on your newly built image. Use the following format for the command:
```
paolaperaza, line 175:
```suggestion
For CI/CD pipelines that push code to a production Deployment, Astronomer recommends adding a unit test after the image build step to ensure that you don't push a Docker image with breaking changes. To run a basic unit test, add a step in your CI/CD pipeline that executes `docker run` and then runs `pytest tests` in a container that is based on your newly built image before it's pushed to your registry. Use the following format for the command:
```

Few minor wording suggestions here -- was getting confused on what was a CI/CD pipeline vs a DAG vs a Deployment

software/ci-cd.md
```diff
@@ -170,6 +170,16 @@ docker build -t registry.${BASE_DOMAIN}/${RELEASE_NAME}/airflow:ci-${DRONE_BUILD
...
+
+For example, you can add the following command as a step in a [Drone](ci-cd.md#DroneCI) pipeline:
```
paolaperaza, line 177: Typo here? I see two leading statements:

- `Use the following format for the command:`
- `For example,`

Did we miss the format one before example?

software/ci-cd.md
```diff
@@ -170,6 +170,16 @@ docker build -t registry.${BASE_DOMAIN}/${RELEASE_NAME}/airflow:ci-${DRONE_BUILD
...
+
+```bash
+docker run --rm registry.${BASE_DOMAIN}/${RELEASE_NAME}/airflow:ci-${DRONE_BUILD_NUMBER} /bin/bash -c "pytest tests"
```
paolaperaza, line 180: Maybe add half a sentence that clarifies that these env vars were already set in your CI/CD tool

paolaperaza (approved):
Coupla minor comments here but huge thanks for this @dintorf !

Wondering how this might apply to Astro given lack of Docker image building... feel free to give us feedback there too.

dintorf:
> Coupla minor comments here but huge thanks for this @dintorf !
> 
> Wondering how this might apply to Astro given lack of Docker image building... feel free to give us feedback there too.

@paolaperaza, thanks for the feedback and updates! I don't think this would apply to Astro because we require the Astro CLI to deploy DAGs, and I believe that `astro deploy` automatically runs `pytest` during runtime. If that's not the case, I could add updates to the Astro docs to call `astro dev pytest` before `astro deploy`. Please let me know if that's necessary 😄

software/ci-cd.md
dintorf: good catch, removing the "following format" part since it's displayed in the example.

jwitz:
@dintorf I think it's okay not to mention it, since the default behavior of `astro deploy` on Astro is to automatically run that `pytest` command before anyways. This looks good to merge!

dintorf:
@jwitz do I need to merge it or is there a release process that needs to be followed? This is my first contribution to the docs 😄

jwitz:
Ah sorry should have clarified, you can just merge away once a doc reviewer gives approval! I can do it this time though!

PR #1135, opened 2022-09-02
Add "release-header" class to each h2 on release notes pages

katiebrady-astro:
This PR creates a [Swizzle](https://docusaurus.io/docs/swizzling) of the MDXContent component. If the current page is one with release notes then add the `release-header` class to each h2 element.

### DIFF
--- a/src/theme/MDXContent/index.js
+++ b/src/theme/MDXContent/index.js
@@ -1,0 +1,20 @@
+import React, { useEffect } from 'react';
+import MDXContent from '@theme-original/MDXContent';
+
+export default function MDXContentWrapper(props) {
+  useEffect(() => {
+    // Check if this is a release-notes page
+    let isReleaseNotesPage = document.querySelector('html[class*="release-notes"]') !== null;
+
+    // if it is, then add the "release-header" class to each h2 element
+    if (isReleaseNotesPage) {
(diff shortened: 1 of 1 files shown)


jwitz (approved):
@katiebrady-astro thank you for doing this! I assume that when Juliana said that the headers need to be "more distinct", that meant in reference to the rest of the docs site? We don't have to have distinct classes for each individual release right?

katiebrady-astro:
> 

Hey @jwitz np! Yeah I think she just needs a new, distinct class be added to each release note header to help distinguish them for the RSS feed.

For example `<h2 class="anchor anchorWithStickyNavbar_LWe7 release-header" id="astro-runtime-508">`

Where 'release-header' is the new class, and it's just being added to all `<h2>` tags on release notes pages (Astro Runtime, CLI, etc) but not the other pages.

paolaperaza:
Thanks for this @katiebrady-astro ! Jake I'll let you merge this if you're ready? Let's add to style guide in Notion too

jwitz:
@paolaperaza This change should not require any additional maintenance or updates on our part. Docusaurus will now automatically tag any `##` headers in our release notes with the new class name so that they can be picked up by an RSS feed. 

In the future, we might consider adding an RSS subscribe button to our release notes pages so that people know this feature exists.

PR #1154, opened 2022-09-07
Rework Software "Administration" tab information architecture

jwitz:
After some extended time with the "Administration" tab in production, it's clear that there's some improvements we can make to how it's organized. The docs in each of the setup/ management/ deployment management tabs are too diverse to be grouped like they are, so you really have to do some digging if you're navigating docs purely from the sidebar. 

I'm proposing that have a more topic-based approach that makes it easier to find docs if you already know what you're looking for.

@cmarteepants Is this a step in the right direction? And more specifically:
- Do you think "Manage Resources" is scoped correctly?
- Should `export-task-logs` and `logs-to-s3` be in Observability or an Administrative topic? They are definitely administrative tasks, but I could also see keeping them in Observability because that's where the average person might expect to find that information.

### DIFF
--- a/sidebarsSoftware.js
+++ b/sidebarsSoftware.js
@@ -42,18 +42,10 @@ module.exports = {
       'kubernetes-executor',
         ],
       },
+      'deploy-cli',
       'cli-podman',
       'upgrade-astro-cli',
     ],
-    },
-    {
-      type: 'category',
-      label: 'Deploy',
-      items: [
-      'deploy-cli',
-      'configure-deployment',
-      'environment-variables',
-      ],
     },
     {
       type: 'category',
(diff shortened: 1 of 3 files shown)


kwcanuck (approved):

cmarteepants (approved):
Wow, this is great!

> Do you think "Manage Resources" is scoped correctly?
Yes

> Should export-task-logs and logs-to-s3 be in Observability or an Administrative topic? 
Good point. Let's do Observability then.

sidebarsSoftware.js
```diff
@@ -69,43 +62,57 @@ module.exports = {
...
-        'apply-platform-config',
+          'configure-deployment',
+          'environment-variables',
+          ],
+        },
+        {
+        type: 'category',
+        label: 'CI/CD and automation',
+        items: [
+        'ci-cd.md',
```
jwitz, line 87:
```suggestion
        'ci-cd',
```

Issue #1157, opened 2022-09-07
Add packages installed by `astronomer-providers` to Runtime architecture docs


jwitz:
**Affected Document:** https://docs.astronomer.io/astro/runtime-image-architecture

**Problem Type (Priority):**

- [ ] Error (Critical): This document has existing information that's incorrect and prevents readers from completing the documented setup (Add the `error/critical` label to this issue)
- [X] Misleading (Important): This document could benefit from additional context, guidance, or clarity (Add the `error/FYI` label to this issue)
- [ ] Typo (Low): Any issue related to spelling/ punctuation/ formatting (Add the `error/typo` label to this issue)

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):** `astronomer-providers` installs other OSS packages that we don't document as part of the default Runtime package collection. We should document these and any considerations that arise from them being installed as part of  `astronomer-providers`

**Additional Notes:**

PR #1166, opened 2022-09-08
Add provider packages installed with `astronomer-providers`

jwitz:
Resolves https://github.com/astronomer/docs/issues/1157

@jedcunningham Do we need to add any note about the packages if they're installed through `astronomer-providers`? Is there a functional difference, or is this identical to installing them at the base level of the image?


kwcanuck (approved):

jedcunningham:
There is no functional difference. We might entertain explicitly installing these though, that way if astronomer providers stops installing them down the road we still will.

PR #1172, opened 2022-09-09
Update install-aws.md

dnarain-astro:
The URL for the status page is incorrect


kwcanuck (approved):
Good catch! Verified it's correct in the other reference docs.

PR #1179, opened 2022-09-12
adding advisiory not to upgrade to 0.30 directly

nandlalyadav57:
adding note or advisory not to upgrade to 0.30 directly


jwitz:
Hi @nandlalyadav57, thank you for the addition! I think we have this detail covered in https://github.com/astronomer/docs/pull/1152, so I am going to close this PR for now. If we need more direct advice for 0.30 in the future, we an add this back in then.

PR #1205, opened 2022-09-14
Update install-gcp-standard.md

magdagultekin:
Can't create a secret like shown in the docs - without the `" "`, getting an error:

```
zsh: no matches found: --from-literal=connection=smtp://UUU:PPP@email-smtp.us-east-1.amazonaws.com/?requireTLS=true
```


jwitz (approved):
Thank you @magdagultekin ! I assume this applies to all clouds, so hopped in here and made the same change to AWS and Azure

PR #1217, opened 2022-09-16
DataDog Metrics Refinement

stuart23:
There was a spelling mistake added in the last PR on this document.

I also think logically, the explanation what gets launched during the setup should come before the steps to confirm - but I'm ok if the setup>confirm>explain flow is still preferred.


howardyoo (approved):
I actually like the new proposed order compared to the previous one. When I read it, it gives me a better flow of thoughts of what's happening on each step, rather than explaining everything at the end, so would like to approve it.

jwitz (approved):
LGTM! Thanks for the refinement @stuart23

Issue #1227, opened 2022-09-20
Retire "Known limitations" doc


jwitz:
Let's disperse the info on Known Limitations into the docs where it's actually relevant

Issue #1224, opened 2022-09-20
Create Astro Runtime menu for Astro


jwitz:
We have enough Runtime docs that we should have a top level "Astro Runtime" menu. Currently, they are all in reference. The menu would have 

- Image architecture
- Maintenance policy
- Upgrade Runtime (This is also in "Develop" but I think we can have a duplicate of it here)

PR #1228, opened 2022-09-20
Update Astro product IA

jwitz:
Resolves https://github.com/astronomer/docs/issues/1227
Resolves https://github.com/astronomer/docs/issues/1224

Two comments for this PR:

- I'm playing with the idea of having Runtime docs in multiple menus (Release notes is in both the Runtime and the Release Notes menus now)
- I couldn't find an appropriate home for the limitation about the Celery executor.. we might need our "Migrate DAGs" doc for that one.


kwcanuck (approved):

paolaperaza (changes requested):
Hey Jake -- let's chat about this quick before you merge. From the top, I'm seeing a redundancy in `Astro Runtime` being the name of release notes and the name of the top-level menu:

image

We need to either:
- Get rid of `Release Notes` as a top-level navigation
- Not include Astro Runtime release notes in the Astro Runtime menu

I'm actually not super convinced we _need_ an Astro Runtime menu? Anything wrong with keeping these docs in `Reference` (which I agree, we should clean up)

sidebarsAstro.js
```diff
@@ -77,6 +77,16 @@ module.exports = {
         'airflow-alerts',
       ],
     },
+    {
+      type: 'category',
+      label: 'Astro Runtime',
+      items: [
+        'upgrade-runtime',
+        'runtime-image-architecture',
+        'runtime-release-notes',
```
jwitz, line 86:
```suggestion
```

sidebarsAstro.js
```diff
@@ -77,6 +77,15 @@ module.exports = {
...
     },
+    {
+      type: 'category',
+      label: 'Astro Runtime',
+      items: [
+        'upgrade-runtime',
+        'runtime-image-architecture',
+        'runtime-version-lifecycle-policy',
+      ],
+    },
```
jwitz, line 88:
```suggestion
```

sidebarsAstro.js
```diff
@@ -129,9 +138,6 @@ module.exports = {
       type: 'category',
       label: 'Reference',
       items: [
-        'known-limitations',
-        'runtime-image-architecture',
-        'runtime-version-lifecycle-policy',
         'astro-support',
```
jwitz, line 141:
```suggestion
        'astro-support',
        {
          type: 'category',
          label: 'Astro Runtime',
          items: [
          'runtime-image-architecture',
          'runtime-version-lifecycle-policy',
          ],
      },
```

jwitz:
@paolaperaza I was mainly trying to replicate what we had on Software, where we have both an AC and a Runtime menu. Maybe a compromise would be to have a Runtime submenu in the Reference menu? I've implemented that idea here.

paolaperaza (approved):

paolaperaza:
@jwitz Should we add a redirect from known-limitations to Astro home?

PR #1236, opened 2022-09-21
Remove reference to Runtime features that are Astro only

ashb:
These bullet points are astro only, and aren't activated when Runtime is used in Software.

I'm not sure if this page is "shared" between Astro and Software or meant to be for Software only (which is what I took from the URL of it)


jwitz:
Thanks for the PR @ashb , I jumped in and made some other changes to make the doc more Software specific. Once @kwcanuck or @paolaperaza approves I'll get this merged and push the changes to earlier versions of the doc.

cmarteepants (approved):

Issue #1257, opened 2022-09-26
Astro CLI reference doc for `astro dev upgrade-check`


jwitz:
**Request type:**

- [X] **New document:**: Add the `docs-content` label to this issue.
- [ ] **Addition to an existing document**: Add the `enhancement` label to this issue.

**What is the purpose of the document/enhancement? Select all that apply:**

Create a doc for `astro dev upgrade-check` in the Astro docset

jwitz:
Closing this PR/ issue out. Seems like the command will be broken for the foreseeable future, and we no longer support any AC/ Runtime image that would require this command. Let's assume Airflow 2.0+ on our Astro docs and see how we get on. Would rather not document a broken command and make sure it gets deprecated in a future CLI release.

PR #1272, opened 2022-09-27
Add documentation on `astro dev upgrade-check` to Astro docs

jwitz:
Resolves https://github.com/astronomer/docs/issues/1257

### DIFF
--- a/astro/cli/astro-dev-upgrade-check.md
+++ b/astro/cli/astro-dev-upgrade-check.md
@@ -9,8 +9,12 @@ Run the [Airflow 2 upgrade check script](https://airflow.apache.org/docs/apache-
 
 ## Usage
 
-Use this command only if the DAGs in your Astro project were imported from another Airflow project on version 1.14-1.15. The command is compatible with all supported versions of [Astro Runtime](runtime-image-architecture.md).
+Use this command only if the DAGs in your Astro project were imported from another open source or commercial Airflow project on version 1.14-1.15. The command is compatible with all supported versions of [Astro Runtime](runtime-image-architecture.md).
+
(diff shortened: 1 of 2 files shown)


astro/cli/astro-dev-upgrade-check.md
```diff
@@ -0,0 +1,20 @@
...
+title: "astro dev upgrade-check"
+id: astro-dev-upgrade-check
+description: Reference documentation for astro dev upgrade-check.
+---
+
+Run the [Airflow 2 upgrade check script](https://airflow.apache.org/docs/apache-airflow/stable/upgrading-from-1-10/upgrade-check.html), which parses the DAGs in your Astro project and tells you whether the code is compatible with Airflow 2.
+
+## Usage
+
+Use this command only if the DAGs in your Astro project were imported from another open source or commercial Airflow project on version 1.14-1.15. The command is compatible with all supported versions of [Astro Runtime](runtime-image-architecture.md).
```
jwitz, line 12:
```suggestion
Use this command only if the DAGs in your Astro project are imported from another open source or commercial Airflow project on version 1.14-1.15. The command is compatible with all supported versions of [Astro Runtime](runtime-image-architecture.md).
```

jwitz:
Blocked until we determine what to do about the fact that this command does not work with 2.0+ Airflow images

jwitz:
Closing this PR/ issue out. Seems like the command will be broken for the foreseeable future, and we no longer support any AC/ Runtime image that would require this command. Let's assume Airflow 2.0+ on our Astro docs and see how we get on. Would rather not document a broken command and make sure it gets deprecated in a future CLI release.

PR #1321, opened 2022-10-07
UX Navigation Redesign

katiebrady-astro:
This PR implements the new navbar and mobile navigation design.

Screen Shot 2022-10-07 at 2 37 21 PM
Screen Shot 2022-10-07 at 2 37 37 PM


jwitz:
Wow Katie I'm in love with this redesign 😍 I only have a few tiny notes:

- My guess is that we will not be ready to launch the home page for this current iteration of docs, so I might have to remove that tab for now. Is that as easy as removing its entry in `docusaurus-config`, or are there other dependencies?
- I think the Software version selector should only have "Select Software version" above the version selector, rather than also having the version listed in two places:
Screen Shot 2022-10-10 at 11 17 24 AM


We obviously still have to finalize our landing pages, would it make more sense to merge this into the base PR now and handle the landing pages in a separate PR? I'm cool with whatever's easier.

katiebrady-astro:
> Wow Katie I'm in love with this redesign 😍 I only have a few tiny notes:
> 
> * My guess is that we will not be ready to launch the home page for this current iteration of docs, so I might have to remove that tab for now. Is that as easy as removing its entry in `docusaurus-config`, or are there other dependencies?
> * I think the Software version selector should only have "Select Software version" above the version selector, rather than also having the version listed in two places:
>   Screen Shot 2022-10-10 at 11 17 24 AM
> 
> We obviously still have to finalize our landing pages, would it make more sense to merge this into the base PR now and handle the landing pages in a separate PR? I'm cool with whatever's easier.

Thanks Jake, glad you like it! 😄 I've removed the version number from the select label.

And I should have the new Home page staged in the next day or two in time for launch unless there's more edits outside the Figma file it's dependent on. A separate PR sounds good to me if the rest of this looks good! 🙂

jwitz:
Sweet, let's do a separate PR for the landing pages. I'll merge this now then so I can start making my own changes in the base PR

PR #1330, opened 2022-10-12
Update get-started-with-airflow-part-2.md

magdagultekin:
Added missing word. 

PS. In the step 8 there's a sentence "Notice how both the SimpleHttpOperator as well as the GithubTagSensor are part of the provider packages you installed." - to be fair, we only install Github provider as HTTP is pre-installed :)


jwitz (approved):
Thank you @magdagultekin !

PR #1346, opened 2022-10-14
Bug fix for home activeBaseRegex

katiebrady-astro:
This PR fixes a bug so the "Home" tab is properly underlined when on the Docs Overview home page.

image


jwitz (approved):
brilliant @katiebrady-astro I was trying to fix this by matching to an empty string, didn't realize it needed to match `/`!

PR #1358, opened 2022-10-17
add runtime 5.0.10 release notes

paolaperaza:
Based on: https://astronomer.slack.com/archives/C020PVBJPKK/p1666023956020409

### DIFF
--- a/astro/runtime-release-notes.md
+++ b/astro/runtime-release-notes.md
@@ -83,9 +83,8 @@ For a complete list of commits, see the [Apache Airflow 2.4.0 release notes](htt
 
 ### Bug fixes
 
-- Revert “Cache the custom secrets backend so the same instance gets re-used” (#25556)
-- Remove TaskFail duplicates check (#26714)
-- Fix faulty executor config serialization logic (#26191)
+- Revert “Cache the custom secrets backend so the same instance gets re-used” ([#25556](https://github.com/apache/airflow/pull/25556))
+- Fixed faulty Kubernetes executor config serialization logic
 
 ## Astro Runtime 5.0.9
 
--- a/astro/runtime-release-notes.md
+++ b/astro/runtime-release-notes.md
@@ -11,39 +11,6 @@ Astro Runtime is a Docker image built and published by Astronomer that extends t
 
 To upgrade Astro Runtime, see [Upgrade Astro Runtime](upgrade-runtime.md). For general product release notes, see [Astro Release Notes](release-notes.md). If you have any questions or a bug to report, contact [Astronomer support](https://cloud.astronomer.io/support).
 
-## Astro Runtime 4.2.7
-
-- Release date: October 11, 2022
-- Airflow version: 2.2.5
-
-### Backported Airflow bug fixes
-
-Astro Runtime 4.2.7 includes the following bug fixes from later Apache Airflow releases:
-
-- Make sure finalizers are not skipped during exception handling ([#22475](https://github.com/apache/airflow/pull/22475))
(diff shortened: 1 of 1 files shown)


astro/runtime-release-notes.md
```diff
@@ -101,9 +68,25 @@ For a complete list of commits, see the [Apache Airflow 2.4.0 release notes](htt
...
+- Airflow version: 2.3.4
+
+### Additional improvements
+
+- Upgraded `astronomer-providers` to 1.10.0, which includes two new deferrable versions of operators, `SFTPSensorAsync` and `ExternalDeploymentTaskSensorAsync`. See the [Astronomer Providers changelog](https://github.com/astronomer/astronomer-providers/blob/1.10.0/CHANGELOG.rst).
+- Upgraded `openlineage-airflow` to version `0.15.1`. See the [OpenLineage changelog](https://github.com/OpenLineage/OpenLineage/blob/main/CHANGELOG.md).
+
+### Bug fixes
+
+- Revert “Cache the custom secrets backend so the same instance gets re-used” (#25556)
```
jwitz, line 86:
```suggestion
- Revert “Cache the custom secrets backend so the same instance gets re-used” ([#25556](https://github.com/apache/airflow/pull/25556))
```

astro/runtime-release-notes.md
```diff
@@ -101,9 +68,25 @@ For a complete list of commits, see the [Apache Airflow 2.4.0 release notes](htt
...
+- Remove TaskFail duplicates check (#26714)
```
jwitz, line 87:
```suggestion
```
This is a check for Runtime only and has no end user impact afaik, I think we can remove it?

astro/runtime-release-notes.md
```diff
@@ -101,9 +68,25 @@ For a complete list of commits, see the [Apache Airflow 2.4.0 release notes](htt
...
+- Fix faulty executor config serialization logic (#26191)
```
jwitz, line 88:
```suggestion
```
Same with this, we should either check with a SME on actual user impact and remove the commit # or remove the note entirely.

astro/runtime-release-notes.md
```diff
@@ -286,6 +269,39 @@ Astro Runtime 5.0.0 provides support for [Airflow 2.3.0](https://airflow.apache.
 
 For more information on Airflow 2.3, see ["Apache Airflow 2.3 — Everything You Need to Know"](https://www.astronomer.io/blog/apache-airflow-2-3-everything-you-need-to-know) by Astronomer.
 
+## Astro Runtime 4.2.7
```
jwitz, line 272: I'm cool with this for now but we should have a docs discussion about whether we want release notes sorted chronologically or by version #.

jwitz (approved):
3 minor comments, otherwise LGTM!

astro/runtime-release-notes.md
jedcunningham: What do you mean this is a "check for Runtime only"?

This didn't "break" anything afaik, it was just unnecessary and incorrect migration work. Do I have that right @dstandish?

astro/runtime-release-notes.md
jedcunningham: This definitely should stay, as it's a fix for KubernetesExecutor executor_config. See the linked issues.

astro/runtime-release-notes.md
jwitz: @jedcunningham I'm confused, what's the difference between commits from `airflow-private` and OSS?

jedcunningham: Generally, nothing. I'm happy to have a chat about it if you'd like the understand the process a bit better. Short version is these are just commits slated for newer Airflow versions that are being applied to older ones.

PR #1390, opened 2022-10-26
Tiny concept guide update: replace dummy with empty operator in Run SQL

TJaniF:
Replacing the DummyOperator with the EmptyOperator in the Using Airflow to Execute SQL concept guide after it was raised on Slack. :)

### DIFF
--- a/learn/airflow-sql.md
+++ b/learn/airflow-sql.md
@@ -237,7 +237,7 @@ First, create a DAG that pulls COVID data from an [API endpoint](https://covidtr
 
 ```python
 from airflow import DAG
-from airflow.operators.dummy_operator import DummyOperator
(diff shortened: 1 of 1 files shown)


kentdanas (approved):

paolaperaza (approved):
Approving and merging while we get you codeowner permissions @kentdanas !!

paolaperaza:
Actually -- @TJaniF did we wanna fix _all_ instances of this? After a quick search in our repo, looks like there's a few more:

image

I forget what version of Airflow this came in so feel free to ignore some of these if you think we need to. Will keep my ✅ and let you take this the rest of the way.

TJaniF:
@paolaperaza: Good idea! I did it in a second PR #1397 (in hindsight I could have just added it here 😅 )

PR #1397, opened 2022-10-27
Replacing dummy with empty operators

TJaniF:
Replacing all but one instance of the word "dummy" in the text too. Also, the EmptyOperator is not on the registry, so I linked to the OS docs for it in one place.

I removed the dummy trigger_rule since it has been deprecated in favor of 'always'. (see: https://github.com/apache/airflow/pull/17144)


jwitz (approved):
LGTM!

PR #1431, opened 2022-11-03
Update decorators concept guide

kentdanas:
Update list of available decorators to include new decorators released in Airflow 2.3 and 2.4


TJaniF (approved):
LGTM! 

One question that isn't about the update: on line 10 and 23 there is a reference to "the `astro` library" I assume this refers to the Astro Python SDK? I found that a little confusing. The sdk is also referred to as `astro` again in the steps on line 303 to 308.

kentdanas:
Good catch @TJaniF! Those references were outdated from back when the whole package was just called "astro". We updated the name to make it less confusing :) I think I got everything updated here.

PR #1443, opened 2022-11-08
Create "Transfer a Deployment" topic

jwitz:
Release notes will be in another PR, but the language will be very similar to whatever's approved here!

### DIFF
--- a/astro/configure-deployment-resources.md
+++ b/astro/configure-deployment-resources.md
@@ -73,6 +73,20 @@ Alert emails assigned to a Deployment are used by Astronomer support to notify r
 
 In addition to alert emails for your Deployments, Astronomer recommends subscribing to the [Astro status page](https://status.astronomer.io). When you subscribe, you'll receive email notifications about system-wide incidents in real time.
 
+## Transfer a Deployment to another Workspace 
+
+To transfer a Deployment from one Workspace to another, the Workspaces must be in the same Organization and Astro cluster. Transferring a Workspace can be helpful when your organization needs to change user access to a Deployment. Transferring a Deployment moves all DAGs, task history, connections, API keys, and other Astro configurations. Running tasks are not interrupted and tasks will continue to be scheduled.
+
+Only the users who are members of the target Workspace can access the Deployment after it is transferred. To transfer a Deployment, you must be a Workspace Admin in the original Workspace and a Workspace Admin or Editor in the target Workspace.
+
+1. In the Cloud UI, select a Workspace and then select a Deployment.
+2. Click the **Options** menu of the Deployment you want to transfer, and select **Transfer Deployment**. 
+
+    ![Transfer Deployment in options menu](/img/docs/transfer-deployment.png)
+
+3. Select the target Workspace where you want to transfer the Deployment. 
+4. Click **Transfer Deployment**.
+
 ## Delete a Deployment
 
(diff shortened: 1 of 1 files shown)


astro/configure-deployment-resources.md
```diff
@@ -73,6 +73,20 @@ Alert emails assigned to a Deployment are used by Astronomer support to notify r
...
 In addition to alert emails for your Deployments, Astronomer recommends subscribing to the [Astro status page](https://status.astronomer.io). When you subscribe, you'll receive email notifications about system-wide incidents in real time.
 
+## Transfer a Deployment to another Workspace 
+
+To transfer a Deployment from one Workspace to another, the Workspaces must be in the same Organization and Astro cluster. Transferring a Workspace can be helpful when your organization needs to change user access to a Deployment. Transferring a Deployment moves all DAGs, task history, connections, API keys, and other Astro configurations. Running tasks are not interrupted and tasks will continue to be scheduled.
+
+Only the users who are members of the target Workspace can access the Deployment after it is transferred. To transfer a Deployment, you must be a Workspace Admin in the original Workspace and a Workspace Admin or Editor in the target Workspace.
+
+1. In the Cloud UI, select a Workspace and then select a Deployment.
+2. Click the **Options** menu of the Deployment you want to transfer, and select **Transfer Deployment**.
```
kwcanuck, line 83:
```suggestion
2. Click the **Options** menu of the Deployment you want to transfer, and select **Transfer Deployment**. 
```
It appears that the button is actually named Deployment Actions:

image

I prefer _**Options** menu_, but I'll leave it to you if you want to update the procedure. Maybe get the UI label changed?

kwcanuck (approved):

astro/configure-deployment-resources.md
```diff
@@ -73,6 +73,20 @@ Alert emails assigned to a Deployment are used by Astronomer support to notify r
 
 In addition to alert emails for your Deployments, Astronomer recommends subscribing to the [Astro status page](https://status.astronomer.io). When you subscribe, you'll receive email notifications about system-wide incidents in real time.
 
+## Transfer a Deployment to another Workspace 
+
+To transfer a Deployment from one Workspace to another, the Workspaces must be in the same Organization and Astro cluster. Transferring a Workspace can be helpful when your organization needs to change user access to a Deployment. Transferring a Deployment moves all DAGs, task history, connections, API keys, and other Astro configurations. Running tasks are not interrupted and tasks will continue to be scheduled.
+
+Only the users who are members of the target Workspace can access the Deployment after it is transferred. To transfer a Deployment, you must be a Workspace Admin in the original Workspace and a Workspace Admin or Editor in the target Workspace.
```
staufferl16, line 80: Editors, not just Admins, of the original workspace can initiate a transfer as well.  We're using the same permissions that we use for deployment deletion.
Let me know if this raises any concerns.  Otherwise, I just wanted to point out this nuance.

jwitz, line 80:
```suggestion
Only the users who are members of the target Workspace can access the Deployment after it is transferred. To transfer a Deployment, you must be a Workspace Admin or Editor in both the original Workspace and the target Workspace.
```

astro/configure-deployment-resources.md
jwitz: The only way you can see that label is through an accessibility reader, I'm not sure we want to label it as such until the text is visible to all users in the UI.

kwcanuck: @jwitz from the _Google developer documentation style guide_:

image

The guidance from our selected style guide is to use the text displayed in the `aria-label` field when defining the button name. Your call.

kwcanuck: @jwitz my recommendation to remove "of the Deployment you want to transfer" from step 2 remains. The user has already been instructed to select a Deployment in step 1.

PR #1453, opened 2022-11-08
Update Soda guide to new tutorial format

kentdanas:
@TJaniF mind checking this over to make sure I got it all right? Shouldn't be anything new here, just rearranged the existing content a bit.

### DIFF
--- a/learn/soda-data-quality.md
+++ b/learn/soda-data-quality.md
@@ -65,7 +65,7 @@ Configure a new Astro project to run Airflow locally.
 
 Create a configuration file to connect to your database backend. The following example uses the template from the Soda documentation to create the configuration file for Snowflake.
 
-```YAML
+```yaml
 # the first line names the datasource "MY_DATASOURCE"
 data_source MY_DATASOURCE:
   type: snowflake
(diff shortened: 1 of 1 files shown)


learn/soda-data-quality.md
```diff
@@ -25,113 +29,43 @@ To get the most out of this tutorial, make sure you have an understanding of:
...
+    ```
 
-### Step 1: Create the configuration file
+## Step 2: Create the configuration file
 
-You need to create a configuration file to connect to Snowflake. The following example uses the template from the Soda documentation to create the configuration file.
+Create a configuration file to connect to your database backend. The following example uses the template from the Soda documentation to create the configuration file for Snowflake.
 
-```YAML
+```yaml
```
kentdanas, line 68: general YAML question - I assume here that the user is supposed to put their username/password in double quotes but the other values without quotes? Asking because I noticed that anything without quotes doesn't get syntax highlighting in the deploy preview. Which looks odd....but maybe it's just our code styling?

TJaniF: Yes, this is the syntax that worked for me (username pw with quotes, the rest without). Though I think the requirement to have the quotes might have been dropped by now, I cannot find the example in their docs anymore and instead find example configurations without quotes. If my instance where I tested it ever starts (it has been updating packages for 15min now 😅) I can test if it works without quotes.

TJaniF (approved):
LGTM! Few tiny comments.

jwitz (approved):
LGTM!

PR #1499, opened 2022-11-16
we should mention smtp server URL too

nandlalyadav57:
some of the users might miss SMTP URL configuration by referring to our website
an example ticket:
https://astronomer.zendesk.com/agent/tickets/14150


nandlalyadav57:
closed those requests are no longer needed to edit the doc as we have secret info mentioned in helm section of the doc

PR #1507, opened 2022-11-16
Fix DAG Deploy CI/CD Scripts

sunkickr:
This PR updates the DAG Deploy CI/CD pipelines

- Fixes a few typos with the branch deploy CI/CD pipeline
- Updates the code that determine if only DAGs changed to make it more readable

These new CI/CD scripts need to be QA tested


kwcanuck (approved):

astro/ci-cd.md
```diff
@@ -123,13 +134,13 @@ for (( i=0; i<${#lines[@]}; i++ )) ; do
 done
```
jlaneve, line 134: we should be able to get rid of this, right?

sreenusuuda:
@sunkickr 

Below code will not work if dags directory is in another directory

for file in $files; do
  if [[ $file != "dags/"* ]]; then
    echo "$file is not a dag, triggering a full image build"
    dags_only=0
    break
  fi
  
Eg: If my repo contains dags_deployment and then dags directory inside dags_deployment.

I guess we should update if condition like this (adding * as a prefix):
if [[ $file != *"dags/"* ]]; then

Can you share your comments?

jwitz (approved):

astro/ci-cd.md
sunkickr: yes

astro/ci-cd.md
```diff
@@ -479,28 +470,25 @@ This setup assumes the following prerequisites:
     jobs:
       deployment-type:
         runs-on: ubuntu-latest
-          steps:
+        outputs:
+          DAGS_ONLY: ${{ steps.deployment-type.outputs.DAGS_ONLY }}
+        steps:
           # Determine if only DAGs have changes 
         - name: Get Deployment Type
```
sreenusuuda, line 477: @sunkickr : Check out is missing here. Can you please add below code?

- name: checkout repo
      uses: actions/checkout@v3
      with:
        fetch-depth: 2

sreenusuuda: @sunkickr we need one space as prefix at fetch-deptch:2 
like this:
with:
 fetch-depth: 2

astro/ci-cd.md
```diff
@@ -409,33 +407,28 @@ To automate code deploys to a Deployment using [GitHub Actions](https://github.c
...
-            local lines=($OUTPUT)
-            local i
-            for (( i=0; i<${#lines[@]}; i++ )) ; do
-                if [[ "${lines[$i]}" == *"dags/"* ]]
-                then
-                    DAGS_DEPLOY=TRUE
-                else
-                    REGULAR_DEPLOY=TRUE
-                fi
+            files=$(git diff main... --name-only)
```
sreenusuuda, line 410: @sunkickr  I see this in multi branch example and image based deploy example, git diff --name-only HEAD^..HEAD.

Is it ok having git diff main... --name-only instead of git diff --name-only HEAD^..HEAD

sunkickr: what is the difference between  `git diff --name-only HEAD^..HEAD` and `git diff main... --name-only`

astro/ci-cd.md
sunkickr: okay

astro/ci-cd.md
sreenusuuda: @sunkickr  No difference, My understanding was wrong.

PR #1515, opened 2022-11-17
Remove old "create cross account role" script

dnarain-astro:
The new script completely replaces the old script, so removed the old script from the doc (running them in sequence will result in an "role already exists" error)


dnarain-astro:
Found another little nit to change

Issue #1466, opened 2022-11-10
Add Snowflake section to "Enable data lineage" doc (& other updates)


paolaperaza:
### Context

In discussing with @denimalpaca and @virajmparekh about an upcoming workshop with a customer about Open Lineage, Airflow/Astro, and Snowflake -- we recognized that:

- There is nothing about Snowflake and OL in our current docs
- There is no information about what lineage data we collect with the Snowflake Operator
- Our integration guides seem hidden in our overall doc, when it should be the most prominent

### Ask

- [ ] Review the introduction and overview to `Enable data lineage`
- [ ] Consider shrinking the data lineage architecture image. It takes up too much page real estate.
- [ ] Move `Retrieve your OpenLineage API Key` to an H2
- [ ] Consider moving the integration guides OUT of tabs and into H2s so that they're easier to see and link to
- [ ] Remove the `Astronomer` section/tab, as it's only there for older versions of Astro Runtime
- [ ] Add more links to OL docs as needed
- [ ] Create a new `Snowflake` tab or section based on @denimalpaca's doc (https://www.notion.so/astronomerio/Lineage-Collection-in-Snowflake-8ea974cbff1540f4b315367450f80b6b)
- [ ] Include a matrix of open lineage data we collect with Airflow/Astro (from Benji also)

@kwcanuck @jwitz I think we'll need my/other folks' guidance here so hold up before getting started on this issue

jwitz:
Based on discussions: 

- Remove "Architecture" info / tighten it up to be more feature-focused
- Add Snowflake tab to explain that experience 
- Possibly break tabs out into separate information

PR #1539, opened 2022-11-22
Add snowflake-specific data lineage documentation

denimalpaca:
Related to: #1466

### DIFF
--- a/astro/data-lineage-and-snowflake.md
+++ b/astro/data-lineage-and-snowflake.md
@@ -1,0 +1,63 @@
+---
+sidebar_label: 'Data lineage and Snowflake'
+title: "Data lineage and Snowflake"
+id: data-lineage-and-snowflake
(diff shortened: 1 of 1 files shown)


denimalpaca:
let me know if we also want to include a table of some kind that relates all the OL facets to what can be displayed for Snowflake.

jwitz:
Hey Benji, thanks for taking the time to write out this guide! We've created a new home for Snowflake-related lineage information on Astro in this PR: https://github.com/astronomer/docs/pull/1473. Because this current PR doesn't account for those changes, I'm going to close this out for now.

However, happy to add anything to the new topic that you think we're missing from this PR. The main thing that stands out is the "Unsupported queries" info, but if that's an Airflow issue vs. an Astro issue then I don't think we need to document it (at least in Astro docs)

denimalpaca:
The "Unsupported queries" is really an OpenLineage thing, but it'll show up in Astro (or to be more precise, _not_ show up).

PR #1560, opened 2022-11-29
IGNORE THIS: Create mock for documentation around GCP reduced access

dnarain-astro:
I'm just creating a PR so I can have a deployment up to let engineering know what the end user view of this feature will look like.  Comments welcome, but approval/merging are not happening for quite some time.


jwitz:
@dnarain-astro In the future can we use Notion for scratch versions of documents that will eventually be published? Notion accepts Markdown, so you can copy-paste the text from this file there and make the changes as needed in Notion.

We're trying to limit the use of preview deployments exclusively to GitHub review because we don't want links floating around that are subject to change.

dnarain-astro:
Sure! FYI, this isn't a scratch document, its just being written early.
Engineering wanted to know what the end user view of this feature would be
like, and nothing says that like the docs.

I'll be creating a PR and requesting a review in a couple of weeks.

kwcanuck:
Closing for now. If this is required in future, it can be reopened.

PR #1562, opened 2022-11-30
Polish Astro environment variable documentation 

jwitz:
Looking at plausible, this doc is one of the more popular ones users look at. I noticed some deficiencies in one of the section and decided to tidy it up a bit.


astro/environment-variables.md
```diff
@@ -113,67 +113,55 @@ On Astro, environment variables are applied and overridden in the following orde
 
 For example, if you set `AIRFLOW__CORE__PARALLELISM` with one value in the Cloud UI and you set the same environment variable with another value in your `Dockerfile`, the value set in the Cloud UI takes precedence.
 
+## Call environment variables in a DAG
+
+You can call environment variables storing your Airflow variables and connections from DAG code in the following ways:
```
kwcanuck, line 118:
```suggestion
You can call environment variables storing your Airflow variables and connections from DAG code in the following ways: 
```
I find this statement very confusing and I'm not sure what you're trying to convey. Maybe:

You can use the following methods to call environment variables in your DAG code:

astro/environment-variables.md
```diff
@@ -113,67 +113,55 @@ On Astro, environment variables are applied and overridden in the following orde
...
 
-As mentioned above, Airflow connections and variables are stored in Airflow's metadata database. Adding them outside of task definitions and operators requires an additional connection to Airflow's metadata database, which is called every time the scheduler parses a DAG. This parsing process is defined by the `process_poll_interval` environment variable, which is set to 1 second by default.
+Airflow connections and variables are stored in Airflow's metadata database. Calling them outside of task definitions and operators requires an additional connection to Airflow's metadata database which is used every time the scheduler parses a DAG.
 
-By adding connections and variables as environment variables, you can refer to them more easily in your code and lower the amount of open connections, thus preventing a strain on your Database and resources.
+By adding connections and variables as environment variables, you can refer to them more easily in your code and lower the amount of open connections, which improves the performance of your database and resources.
```
kwcanuck, line 133:
```suggestion
By adding connections and variables as environment variables, you can lower the amount of open connections, and improve the performance of your database and resources.
```
image

astro/environment-variables.md
```diff
@@ -113,67 +113,55 @@ On Astro, environment variables are applied and overridden in the following orde
...
-ENV AIRFLOW_CONN_MY_PROD_DB=my-conn-type://login:password@host:5432/schema
-```
-
-For more information on how to generate a Connection URI, see the [Apache Airflow documentation](https://airflow.apache.org/docs/stable/howto/connection/index.html#generating-connection-uri).
+To store this connection as an environment variable, you create an environment variable with the key `AIRFLOW_CONN_MY_PROD_DB` and the value `my-conn-type://login:password@host:5432/schema`,
 
 :::info
 
-Airflow connections set with environment variables do not appear in the Airflow UI and cannot be modified there.
+Airflow connections set with environment variables do not appear in the Airflow UI. They can only be seen and updated from the Cloud UI or your Dockerfile.
```
kwcanuck, line 151:
```suggestion
Airflow connections set with environment variables do not appear in the Airflow UI. They can only be seen and updated in the Cloud UI or your Dockerfile. 
```

astro/environment-variables.md
```diff
@@ -113,67 +113,55 @@ On Astro, environment variables are applied and overridden in the following orde
...
+Airflow connections and variables are stored in Airflow's metadata database. Calling them outside of task definitions and operators requires an additional connection to Airflow's metadata database which is used every time the scheduler parses a DAG.
 
-By adding connections and variables as environment variables, you can refer to them more easily in your code and lower the amount of open connections, thus preventing a strain on your Database and resources.
+By adding connections and variables as environment variables, you can refer to them more easily in your code and lower the amount of open connections, which improves the performance of your database and resources.
 
 ### Airflow connections
 
-The environment variable naming convention for Airflow connections is:
+Use the Airflow [connection URI format](https://airflow.apache.org/docs/apache-airflow/stable/howto/connection.html#uri-format) to store connections as environment variables. The environment variable naming convention for Airflow connections is:
```
kwcanuck, line 137:
```suggestion
Use the Airflow [connection URI format](https://airflow.apache.org/docs/apache-airflow/stable/howto/connection.html#uri-format) to store connections as environment variables. The naming convention for Airflow environment variable connections is:
```

kwcanuck (approved):

astro/environment-variables.md
```diff
@@ -113,67 +113,55 @@ On Astro, environment variables are applied and overridden in the following orde
 
 For example, if you set `AIRFLOW__CORE__PARALLELISM` with one value in the Cloud UI and you set the same environment variable with another value in your `Dockerfile`, the value set in the Cloud UI takes precedence.
 
+## Call environment variables in a DAG
+
+You can call environment variables storing your Airflow variables and connections from DAG code in the following ways: 
+
+- `Variable.get('<environment-variable-key>')`
+- `os.getenv('<environment-variable-key>', '<default_value>')`
```
paolaperaza, line 121: Hm I actually think this section doesn't quite makes sense _before_ the Airflow connections and variables section

- For one, the intro is saying `storing your Airflow variables and connections` even though you haven't introduced that concept yet
- The example here is both "how to call an env var as an env var" and then "how to call variables stored as env vars"

We could split these sections up, but I'd kind of rather first have sections about _creating_ these things and then a consolidated section on calling them in your DAGs

paolaperaza: (I agree the header shouldn't be called "query" though, that's a bit yikes)

paolaperaza (commented):
One comment here @jwitz -- I think let's keep the OG structure if you agree based on my reasoning? Let me know if not and I can take a look at it again.

+1 to all of @kwcanuck's suggestions!

astro/environment-variables.md
jwitz: This section should serve as a more general "How to call _any_ environment variable value in a DAG", regardless of whether it's a connection or Airflow variable or something else. For that reason, I think it makes sense to include before the details about storing connections and variables. 

Rob's edits removed mentions of Airflow connections/ variables, which should make this more clear.

PR #1571, opened 2022-12-02
Fix AC 2.3 Policy to EOL in March 2023

paolaperaza:
I spoke to @astro-rft earlier this week and we realized that we had a miscommunication and did not document the fact that while AC 2.3 was reclassified as a stable release (as we modified in our docs), we ALSO decided to give that version an extra 6 months of support to give folks some wiggle room.

I've updated our AC policy doc such that:

- March 2023 is now the EOL date for AC 2.3
- We specify that this release is stable but has an extra 6 months, and clarify what it was before

@jwitz If this looks good to you, want to apply it to other Software docs + Nebula?


jwitz (approved):
LGTM , will hotfix this to the other doc versions

PR #1594, opened 2022-12-06
Update astro-deployment-inspect.md

magdagultekin:
`--key information.status` doesn't return a cluster id, `--key configutation.cluster_id` does. 

Additionally, this doc seems outdated in general because in the version 1.8.2 I don't see a section `information` - it's all under `metadata` (i.e., to get the status, I need to run `astro deployment inspect <deployment-id> --key metadata.status`).


jwitz (approved):

jwitz:
Thanks @magdagultekin, I'll work with @sunkickr  to make sure that we're always using the latest schema on this doc. I think it's also worth mentioning when we change the schema in release notes.

PR #1615, opened 2022-12-11
[2.5] Adding the sensor decorator and other small updates

TJaniF:
Part 1 of the 2.5 guide updates includes:

- chapter on the sensor decorator in the sensor guide
- adding the sensor decorator to the decorators list
- Replacing screenshots of the Datasets view to include the filter by recent update (Note: currently the highlighting of a Dataset does not seem to be working, when viewing the details of one dataset the graph disappears, that is why I cropped the second screenshot in the UI guide). 
- Adding the one_done trigger rule


learn/what-is-a-sensor.md
```diff
@@ -100,6 +100,63 @@ dag = partner()
...
+    @task
+    def print_shibe_picture_url(url):
+        print(url)
+
+
+    print_shibe_picture_url(check_shibe_availability())
+```
+
+In the DAG the `@task.sensor` decorator turns the `check_shibe_availability()` function into a sensor which will check if the provided API returns the status code 200. If yes the sensor task is marked as successful, if any other status code is returned, the sensor will poke again after the `poke_interval` has passed.
+The optional `xcom_value` parameter of the `PokeReturnValue` defines what data will be pushed to [XCom](airflow-passing-data-between-tasks.md) once the condition provided to the `is_done` parameter is True. In the example above, the information returned by the API is pushed to XCom to be accessed by the downstream task `print_shibe_picture_url`.
```
jwitz, line 158:
```suggestion
Here, `@task.sensor` decorates the `check_shibe_availability()` function, which checks if a given API returns a 200 status code. If the API returns a 200 status code, the sensor task is marked as successful. If any other status code is returned, the sensor pokes again after the `poke_interval` has passed.

The optional `xcom_value` parameter in `PokeReturnValue` defines what data will be pushed to [XCom](airflow-passing-data-between-tasks.md) once the `is_done=true`. You can use this data in any downstream tasks.
```

jwitz (approved):

kentdanas (approved):

PR #1616, opened 2022-12-11
[2.5] Mapping updates

TJaniF:
Part 2 of the 2.5 updates. I separated out the changes around dynamically mapping over task groups because I assume review will take longer on this topic. 

I mostly added examples to the task group guide and made a reference to it in the dynamic task mapping guide. The reason that there are no screenshots of dynamically mapped task groups is that the UI does not show them yet. Brent is working on that, so will add screenshots as soon as it gets implemented (see https://github.com/apache/airflow/pull/28208).


learn/dynamic-tasks.md
```diff
@@ -7,7 +7,7 @@ id: dynamic-tasks
 
 With the release of [Airflow 2.3](https://airflow.apache.org/blog/airflow-2.3.0/), you can write DAGs that dynamically generate parallel tasks at runtime. This feature, known as dynamic task mapping, is a paradigm shift for DAG design in Airflow.
 
-Prior to Airflow 2.3, tasks could only be generated dynamically at the time that the DAG was parsed, meaning you had to change your DAG code if you needed to adjust tasks based on some external factor. With dynamic task mapping, you can easily write DAGs that create tasks based on your current runtime environment.
+Prior to Airflow 2.3, tasks could only be generated dynamically at the time that the DAG was parsed, meaning you had to change your DAG code if you needed to adjust tasks based on some external factor. With dynamic task mapping, you can easily write DAGs that create tasks based on your current runtime environment. In Airflow 2.5 the possibility to [dynamically map over task groups](task-groups.md#generate-task-groups-dynamically-at-runtime) was added.
```
jwitz, line 10:
```suggestion
Prior to Airflow 2.3, tasks could only be generated dynamically at the time that the DAG was parsed, meaning you had to change your DAG code if you needed to adjust tasks based on some external factor. With dynamic task mapping, you can easily write DAGs that create tasks based on your current runtime environment.
```
This feels tacked on to the end of the paragraph which is mainly about how things worked before 2.3 and how much easier they are now.

jwitz (changes requested):
@TJaniF What do you think about moving the main content here to the dynamic tasks guide? IMO someone would be more likely to want this information coming from the context of wanting to work with dynamic tasks, rather than wanting to work with task groups. 

I also see that we're repeating some information that's in the dynamic task guide, such as explaining each mapping method. That makes me think this would be much more at home in the other doc.

kentdanas:
Hey @jwitz thanks for looking at this. I think I had told Tamara to add something to the task groups guide - I do want to include a section to that guide because it's very widely viewed and this is cool functionality that we want to promote. 

But, I made a few changes:

- Updated the task groups guide to include just a short section for dynamically mapping over task groups that shows the DAG Tamara wrote with the syntax and cross-links to the dynamic tasks guide. Cross-linking should be enough for covering the concepts.
- Removed the section on looping over task groups from the task groups guide since this is no longer really relevant with task mapping
- Added a short section to the dynamic tasks guide on mapping over task groups showing the syntax and indicating that mapping over multiple parameters is possible in the same way it is over single tasks. Since the syntax is basically the same, I don't think we need to show all of the examples.

Let me know what you think.

learn/dynamic-tasks.md
```diff
@@ -320,6 +319,33 @@ The add_nums task will have three mapped instances with the following results:
...
+        def add_42(num):
+            return num + 42
+        
+        print_num(my_num) >> add_42(my_num)
+
+    # creating 6 mapped task group instances of the task group group1
+    tg1_object = tg1.expand(my_num=[19, 23, 42, 8, 7, 108])
+```
+
+You can also dynamically map over multiple task group input parameters in the same manner as you would in regular tasks using a cross-product, `zip` function, or sets of keyword arguments. For more on this, see the [Mapping over multiple parameters](#mapping-over-multiple-parameters) section.
```
jwitz, line 347:
```suggestion
You can also dynamically map over multiple task group input parameters as you would for regular tasks using a cross-product, `zip` function, or sets of keyword arguments. For more on this, see [Mapping over multiple parameters](#mapping-over-multiple-parameters).
```

learn/task-groups.md
```diff
@@ -130,37 +130,76 @@ task_group_example = task_group_example()
...
 
-## Dynamically generate task groups
+## Generate task groups dynamically at runtime
 
-Task groups can be dynamically generated to make use of patterns within your code. In an ETL DAG, you might have similar downstream tasks that can be processed independently, such as when you call different API endpoints for data that needs to be processed and stored in the same way. For this use case, you can dynamically generate task groups by API endpoint. You can expand generated task groups in the Airflow UI to see specific tasks. 
-
-In the following code, iteration is used to create multiple task groups. While the tasks and dependencies remain the same across task groups, you can change which parameters are passed in to each Task Group based on the `group_id`:
+As of Airflow 2.5, you can use [dynamic task mapping](dynamic-tasks.md) syntax with the `@task_group` decorator to dynamically map over task groups. An implementation is shown in the DAG below.
```
jwitz, line 142:
```suggestion
As of Airflow 2.5, you can use [dynamic task mapping](dynamic-tasks.md) with the `@task_group` decorator to dynamically map over task groups. The following DAG shows how you can dynamically maps over a task group with different inputs for a given parameter.
```

jwitz (approved):
Looks good! Only thing I'd add is a new UI visual for the task groups generated by at least one of the DAGs. It looks like we're no longer using `looped_task_groups.png`, but something similar for `group1` would be helpful because task groups are such a visual concept

kentdanas:
We don't have a new UI visual because for now the UI doesn't update to show mapped task groups. Sounds like Brent is working on adding this, so when it's included in a future release we'll update with a new screenshot.

PR #1631, opened 2022-12-13
Release Notes RSS Feeds

katiebrady-astro:
This PR adds a node script to automatically generate RSS feeds for release notes! It will run with each site build (either during `yarn start` or `yarn build`) and build an XML file for each of the relative markdown files. The best part: nothing needs to change with how release notes are written! 😄 

A little orange RSS button has also been added to Astro, Astro CLI and Astro Runtime.
image


kwcanuck (approved):
Interesting addition. A few comments:

- When I attempted to use the RSS button, an error message was returned - this is probably the expected behavior prior to deployment. 

- Should hover text be provided to identify the purpose of the button?

- The image seems small and pixelated. It could be my monitor settings.

- I'm not sure the button is in the correct location - it seems odd in its current position. 

- Are RSS feeds still popular? There seems to be some debate about this.

juliana-astro (approved):

juliana-astro:
> 

• I was able to pull up the RSS feed and it looks good to me (pulling the updates and text correctly)

• The button itself was maybe a bit pixelated, but not different from the image I see on other sites.

• RSS feeds on Release Notes have been a recurring customer request. An RSS feed can also be used to set up a Slack channel for an organization.

jwitz:
LGTM! You can merge whenever you're ready @katiebrady-astro

PR #1641, opened 2022-12-14
Add inforamtion about new deploy action

sunkickr:
[Deploy action](https://github.com/astronomer/deploy-action) is a custom Github action that we have developed to make creating CI/CD pipelines on Github Actions easier.

Once this action has been thoroughly tested and published to the GitHub Actions marketplace we will release these docs changes to update the templates to use the new actions

### DIFF
--- a/astro/ci-cd.md
+++ b/astro/ci-cd.md
@@ -37,16 +37,33 @@ Use the Astronomer CI/CD templates to automate deploying code to Astro with popu
 
 Templates allow you to easily configure automated workflows using popular CI/CD tools. Each template can be implemented as-is to produce a simple CI/CD pipeline. Astronomer recommends reconfiguring the templates to work with your own directory structures, tools, and best practices.
 
-Astro supports the following CI/CD workflows:
-
-- Image-only workflows: All files in your Astro project are built into a Docker image and pushed to Astro in a single step.
-- DAG-based workflows: The [DAG-only deploy feature](deploy-code.md#deploy-dags-only) is used to deploy DAGs in your Astro project separate from the Docker image that is built for all other project files.
-
-The following templates are available to implement your CI/CD workflows: 
+There are two workflow types, image-only workflows and DAG-based workflows, which are described in detail below. The following templates are available for each workflow type: 
 
 - Single branch: Deploys a single branch from your version control tool to Astro. This is the default template for all CI/CD tools. 
-- Multiple branch:  Deploys multiple branches to separate Deployments on Astro.
-- Custom image:  Deploys an Astro project with a customized Runtime image and additional build arguments.
+- Multiple branch: Deploys multiple branches to separate Deployments on Astro.
+- Custom image: Deploys an Astro project with a customized Runtime image and additional build arguments.
+
+### Image-only workflows  
+
+The image-only workflow builds a Docker image and pushes it to Astro whenever you update any file in your Astro project. This type of template is simple to set up and works well for development workflows that include complex Docker customization or logic.
+
+CI/CD templates that use image-only workflows do the following:
+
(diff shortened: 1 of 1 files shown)


astro/ci-cd.md
```diff
@@ -361,7 +263,7 @@ To complete this setup, you need:
 
 ### GitHub Actions (DAG-based deploy)
 
-The following templates are examples of how to implement DAG-only deploys in GitHub Actions. These templates can be modified to run on other CI/CD tools.
+The following templates are examples of how to implement DAG-only deploys in GitHub Actions. They use the [Deploy Action's](https://github.com/astronomer/deploy-action) `dag-deploy-enabled` option to implement a dag-based deploy workflow.
```
jwitz, line 266:
```suggestion
The following templates are examples of how to implement DAG-only deploys in GitHub Actions. They use the [Deploy Actions](https://github.com/astronomer/deploy-action) `dag-deploy-enabled` option to implement a DAG-based deploy workflow.
```

astro/ci-cd.md
```diff
@@ -48,95 +48,9 @@ The following templates are available to implement your CI/CD workflows:
...
-    astro deploy
-fi
-```
-
-:::info
-
-All CI/CD pipelines that use the DAG-based deployment method require [Astro CLI v1.7 or later](cli/release-notes.md) to deploy. All CI/CD templates in this document automatically install the latest version of the Astro CLI. If your CI/CD pipeline does not automatically install the latest version of the Astro CLI, make sure to specify version 1.7 or later.
-
-:::
+The following templates use Astronomer's [Deploy Action](https://github.com/astronomer/deploy-action) to deploy code to Astronomer. The Deploy Action is the easiest way to create CI/CD pipelines with Astro. Take a look at the [Deploy Action readme](https://github.com/astronomer/deploy-action#readme) to learn more about its many input options.
```
jwitz, line 53:
```suggestion
The following templates use Astronomer's [Deploy Action](https://github.com/astronomer/deploy-action) to deploy code to Astronomer. The Deploy Action is the easiest way to create CI/CD pipelines for Astro. See the [Deploy Action README](https://github.com/astronomer/deploy-action#readme) to learn more about using and customizing this action.
```

jwitz (changes requested):
@sunkickr lovin the simplified templates. Can you revert the change where you moved the explanations of each type of workflow until after the GH actions templates? Those topics are still important for explaining the GH action templates even if they've been abstracted a bit.

sunkickr:
@jwitz I think we should at least move the actually scripts to the second section. I don't want to overwhelm users with those scripts if they are just going to use the deploy actions. Especially the dag deploy script

jwitz:
@sunkickr I think this is a doc where users mainly jump to sections they're interested in, so I'm not worried about adding too much to the front of the doc as long as it's contained in a skippable topic. We need the explanations of how the scripts work before the templates themselves, because those explanations have useful info for both GitHub Actions and all the other templates.

If you're worried about the shell scripts overwhelming users, why don't we just remove them? It was more important to have those scripts when we weren't able to use `astro deploy` natively on CI/CD, but that's no longer the case. The shell scripting isn't much more descriptive than some of the actions, so I don't think we lose much by removing it.

sunkickr:
@jwitz I don't think we should remove the shell scripts. The shell scripts would be easiest to use if someone was trying to create CI/CD pipeline for a tool we don't cover. It makes sense to describe what the scripts are doing first are the shell scripts being used to do that?

sunkickr:
I moved the workflow descriptions back to the top of the page but kept the scripts in the custom script section

jwitz:
@sunkickr I think your other CI/CD PR will require a few updates to this templates doc. Want to close this one and we can work there on any changes? I want to make sure we're being consistent and not repeating information across both docs.

jwitz:
Going to close out this PR and move any unique content into https://github.com/astronomer/docs/pull/1697

PR #1655, opened 2022-12-19
Update operator-extra-link-tutorial.md

magdagultekin:
Connection generated in `Step 4` is called `random_user_api_conn`, not `randomuser_api`.


TJaniF (approved):
Good catch, thank you! 
It was also wrong in one of the screenshots, I added that to this PR.

PR #1663, opened 2022-12-20
Update airflow-kafka.md

magdagultekin:
https://astronomer.slack.com/archives/C015V2JFKT5/p1671472147598409


TJaniF (approved):
👍 (turns out having 4 vscode windows open with almost identical code is not advisable 😅 )

PR #1677, opened 2022-12-26
fix connection id argument is misnamed

vadzimk:
According to the constructor signature, the connection id argument is not mongo_conn_id but conn_id

<pre>
class MongoHook(BaseHook):
    def __init__(self, conn_id: str = default_conn_name, *args, **kwargs) -> None:
        super().__init__()
        self.mongo_conn_id = conn_id
<pre>

### DIFF
--- a/learn/airflow-mongodb.md
+++ b/learn/airflow-mongodb.md
@@ -110,7 +110,7 @@ from airflow.providers.mongo.hooks.mongo import MongoHook
 from pendulum import datetime
 
 def uploadtomongo(ti, **context):
-    hook = MongoHook(mongo_conn_id='mongo_default')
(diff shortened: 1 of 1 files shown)


kentdanas:
Since it's the only param in the hook it doesn't actually matter what the argument is called. Additionally, the documentation indicates the param is called `mongo_conn_id` https://github.com/apache/airflow/blob/d20fa951e517f239783aae4fb051fd7416d17843/airflow/providers/mongo/hooks/mongo.py#L44

which is a bit inconsistent, as you point out, but I'm inclined to leave the code in the tutorial as-is since it works and is consistent with Mongo's tutorial and with the docs.

PR #1682, opened 2022-12-29
Clarify restrictions on cross-region PrivateLink Endpoints

stuart23:


astro/connect-aws.md
```diff
@@ -100,7 +100,7 @@ If Astronomer creates a new transit gateway in your AWS account for Astro, keep
 
 <TabItem value="AWS PrivateLink">
 
-Use AWS PrivateLink to create private connections from Astro to your AWS services without exposing your data to the public internet.
+Use AWS PrivateLink to create private connections from Astro to your AWS services without exposing your data to the public internet.  PrivateLink endpoints can only be used to expose services in that were created in the same region as the Astro cluster.
```
kwcanuck, line 103:
```suggestion
Use AWS PrivateLink to create private connections from Astro to your AWS services without exposing your data to the public internet.  PrivateLink endpoints can only be used to expose services that were created in the same region as the Astro cluster.
```

astro/connect-aws.md
```diff
@@ -112,6 +112,8 @@ Astro clusters are pre-configured with the following AWS PrivateLink endpoint se
 
 To request additional endpoints, or assistance connecting to other AWS services, contact [Astronomer support](https://cloud.astronomer.io/support).
 
+To access a service in a different region using PrivateLink endpoints, an inter-region VPC Peering Connection can be used to an intermediate VPC. The intermediate VPC can be created in the same region as the targetted service, and the PrivateLink endpoint can then be created in this VPC.
```
kwcanuck, line 115:
```suggestion
To access a service in a different region using PrivateLink endpoints, you can use an inter-Region VPC peering connection. You create the inter-Region VPC peering connection in the same region as the targeted service, and then you create the PrivateLink endpoint in the VPC.
```

stuart23:
@kwcanuck I don't think we properly captured the idea about the intermediate VPC. I drew this diagram to help explain the concept - maybe you can help word it better? If you think the bottom half of the diagram is useful, I can prettify it and we can add it to the docs (chart is here: https://lucid.app/lucidchart/45278dc1-b6d9-4257-9d2e-ebcd0ebee503/edit?viewport_loc=-363%2C-166%2C3072%2C1479%2C0_0&invitationId=inv_03307a46-d9b6-44a8-a0e0-cd109d6ba35b)

image

kwcanuck:
@stuart23 this is your PR and I'm not familiar with the content. I performed a review to allow it to move forward. The changes I recommended were intended to improve readability. If you want to provide alternate or new text, I would be happy to review it.

jwitz:
Looking through the PR, I see that the two added sentences are contradictory to each other. We say you can "only" connect to services in the same region, but then we provide a way to do the opposite of that.

What about "To access a service in a different region using PrivateLink endpoints, you **must** use an inter-Region VPC peering connection. You create an intermediate PrivateLink endpoint in the same region as the targeted service, then connect to that endpoint through an inter-Region VPC peering connection."

@stuart23 Do you have relevant documentation for how someone might set up that intermediate VPC? If you want I can try to make that bottom chart into a [Mermaid flowchart](https://mermaid.js.org/syntax/flowchart.html), which could clarify  what you mean. Otherwise, I think the AWS docs have a similar diagram/ explanation that we could use in a pinch: https://docs.aws.amazon.com/vpc/latest/peering/what-is-vpc-peering.html

stuart23:
Thanks @jwitz - We don't really want to give them the instructions to setup this architecture because that happens in the customers account, and is beyond the scope of our responsibility model. Based off a recent support request, we found that the architecture illustrated below is the only way to used cross-region PrivateLink. As you can see, there is a lot that happens outside of the Astro account for which we are not responsible. I don't think we should include the diagram, but we do have it available for if a customer requires help with this.
image

kwcanuck:
@jwitz do you have additional changes or questions. or can this PR be merged?

jwitz:
@kwcanuck Doesn't look the core suggestion of my comment was addressed

> What about: "To access a service in a different region using PrivateLink endpoints, you must use an inter-Region VPC peering connection. You create an intermediate PrivateLink endpoint in the same region as the targeted service, then connect to that endpoint through an inter-Region VPC peering connection."

If you want to pass on that it's fine, but just lmk either way

kwcanuck:
@jwitz the text has been updated.

astro/connect-aws.md
```diff
@@ -100,7 +100,7 @@ If Astronomer creates a new transit gateway in your AWS account for Astro, keep
 
 <TabItem value="AWS PrivateLink">
 
-Use AWS PrivateLink to create private connections from Astro to your AWS services without exposing your data to the public internet.
+Use AWS PrivateLink to create private connections from Astro to your AWS services without exposing your data to the public internet.  PrivateLink endpoints can only be used to expose endpoint services that were created in the same region as the Astro cluster.
```
jwitz, line 103: The added sentence here is no longer true..

kwcanuck: @jwitz fixed. This is easier when direct edits are provided.

stuart23: Could do with some help with this please. The above statement is true, there is just a work around that we can do if the upstream service is in a separate region. We are still exposing the service with the network load balancer in the same region as Astro, however that network load balancer is able to front a service that is in a different region.

kwcanuck: @stuart23 if the workaround is completed by Astronomer support, we should probably add a statement similar to the following:

_If your AWS services are located in a different region than Astro, contact Astronomer support._

If this isn't accurate, feel free to provide alternate wording and I'll work with you to refine it.

Thanks!

stuart23: That sounds fine to me

kwcanuck: @stuart23 the statement has been added. Thanks!

PR #1685, opened 2022-12-29
Fix outdated Airflow 1 imports

BasPH:
This PR fixes several Python package import issues in the code samples:

- Replace Airflow 1 imports by Airflow 2 imports because Airflow 1 is seriously outdated e.g. `from airflow.operators.bash_operator import BashOperator` --> `from airflow.operators.bash import BashOperator `
- Fix undesired imports such as `from airflow.utils.dates import datetime` --> `from datetime import datetime`
- Remove a few unused imports
- Fix a few outdated imports such as `S3toSnowflakeTransferOperator` --> `S3toSnowflakeOperator`
- Also applied import grouping to the changed code samples so that imports are nicely sorted & grouped according to PEP8 convention


kentdanas (approved):
Thanks @BasPH!

PR #1695, opened 2023-01-03
Tiny changes to the Sagemaker tutorial

TJaniF:
Fixed the step numbering and added the info that you should add `sagemaker.amazonaws.com` to the trust policy (because I had to go back after an error to do this).


kentdanas (approved):
Interesting on the trust policy error.....I don't think I had to do that, but the policies are pretty confusing so maybe I just missed it 🤷‍♀️ 

Thanks for the updates!

TJaniF:
@kentdanas yes, the policies are definitely one of the trickier parts of the tutorial. 

At least the error message I got was helpful:

```airflow.exceptions.AirflowException: SageMaker job failed because ClientError: TrainingHostAgent Initialization failed:InvalidInputException: SageMaker is unable to perform: sts:AssumeRole on the role. Check if you have a trust relationship for 'sagemaker.amazonaws.com' to assume the role.``` 😅

Issue #1683, opened 2022-12-29
Missing Dockerfile code highlighting


BasPH:
**Affected Document:**
All docs that use `` ```docker(file) `` code highlighting are not highlighted. For example https://docs.astronomer.io/astro/develop-project:

![image](https://user-images.githubusercontent.com/6249654/210008860-1f2b49f6-c018-4d2c-b8a1-de88b191b111.png)

**Problem Type (Priority):**

- [ ] Error (Critical): This document has existing information that's incorrect and prevents readers from completing the documented setup (Add the `error/critical` label to this issue)
- [ ] Misleading (Important): This document could benefit from additional context, guidance, or clarity (Add the `error/FYI` label to this issue)
- [x] Typo (Low): Any issue related to spelling/ punctuation/ formatting (Add the `error/typo` label to this issue)

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**
Fix `` ```docker(file) `` code highlighting.

**Additional Notes:**
Found this related PR but it hasn't solved Dockerfile code highlighting https://github.com/astronomer/docs/pull/563.

BasPH:
It seems dockerfile highlighting was added to prismjs in [v1.4.0](https://github.com/PrismJS/prism/commit/951e973eaf6f61f9bbba513ed075638aa1c76b65), and we're currently running [v1.28.0](https://github.com/astronomer/docs/blob/30536388f7eba5443b757ed4cdaa2ca5b5f5406c/yarn.lock#L8105), so should be good.

Not sure what the issue is though...

BasPH:
Think I found the reason. We're using the `prism-react-renderer` package which according to their [readme](https://github.com/FormidableLabs/prism-react-renderer#faq) "only includes an [arbitrary subset of the languages](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js) that Prism supports", and this doesn't include `docker` or `dockerfile`.

According to the instructions, we'd have to include `prismjs/components/prism-docker`.

I'm way out of my comfort zone with front-end stuff so maybe @jwitz can help here?

PR #1699, opened 2023-01-04
Update Docusaurus prism to render Dockerfile syntax

jwitz:
Resolves https://github.com/astronomer/docs/issues/1683


jwitz:
Confirmed that any code fences marked `docker` render styling for Dockerfile commands: 
Screen Shot 2023-01-04 at 11 12 33 AM

Updated `dockerfile` fences because those don't seem to work

PR #1721, opened 2023-01-09
Correcting GCP Secrets Manager AIRFLOW__SECRETS__BACKEND_KWARGS

ReadytoRocc:
Correcting the GCP Secrets Manager `AIRFLOW__SECRETS__BACKEND_KWARGS` example to use `project_id` instead of `project-id`.


kwcanuck (approved):
Thanks for this. Is the the `airflow-connections` and `airflow-variables` syntax correct?

PR #1742, opened 2023-01-13
Move dynamic tasks to DAGs section

kentdanas:
At some point we might want to think about alphabetizing these or ordering in some other way....but for now just want to get this one out of 'Advanced'.


TJaniF (approved):
LGTM! (and makes me happy 😊 )

PR #1756, opened 2023-01-19
Update SDK concept guide comparison structure

### LINKED ISSUE #1512: Break down before/ after DAG in Astro Python SDK guide by function
I was reading through our Learn guide on the Astro Python SDK and think the current compare/ contrast structure could be further broken down into comparing individual functions. Right now we present readers with a long bulleted list of benefits and differences, whereas breaking the comparisons down to individual functions could let users jump to the parts they're most interested in optimizing. For example, comparing just `combine_tables` with `extract_data` in a title called "Extracting data". We could then remove the giant "before" dag code block and instead focus mainly on the Astro Python SDK implementation .

kentdanas:
Resolves #1512 

@jwitz let me know what you think. No net-new content here, just rearranged the comparison part into different sections as suggested in the issue.

### DIFF
--- a/learn/astro-python-sdk-etl.md
+++ b/learn/astro-python-sdk-etl.md
@@ -62,9 +62,9 @@ Using the Astro Python SDK requires configuring a few things in your Airflow pro
 
 For a guided experience to get started, see the [Astro Python SDK tutorial](astro-python-sdk.md).
 
-## Example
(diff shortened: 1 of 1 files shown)


jwitz (approved):
Love it! Thanks for working on this @kentdanas

PR #1791, opened 2023-01-25
Use Groovy highlighting for Jenkinsfiles

BasPH:
We have a few example Jenkinsfiles in the docs, would be nice to have syntax highlighting.

Also found an invalid Jenkinsfile with too many brackets and fixed that in the process.

Before:
image

After:
image

### DIFF
--- a/astro/ci-cd.md
+++ b/astro/ci-cd.md
@@ -664,32 +664,33 @@ To automate code deploys to a single Deployment using [Jenkins](https://www.jenk
 
 2. At the root of your Git repository, add a [Jenkinsfile](https://www.jenkins.io/doc/book/pipeline/jenkinsfile/) that includes the following script:
 
-    <pre><code parentName="pre">{`pipeline {
-       agent any
-         stages {
-           stage('Deploy to Astronomer') {
-             when {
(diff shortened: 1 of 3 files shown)


jwitz (changes requested):
@BasPH Apologies for the premature review but wanted to add: At the moment I would prioritize hosting the latest version of the Astro CLI over syntax highlighting. If there's a way to keep both the highlighting and the environment variables then I'm all for it.

BasPH:
I'm not following, what do you mean by "hosting the latest version of the Astro CLI"? Those seem like unrelated things?

jwitz:
@BasPH We autoload the latest CLI version into Jenkins scripts using `${siteVariables.cliVersion}`, which is absent in your edit. AFAIK this type of variable can only work when you use `<pre><code parentName="pre">` tags, so I don't think we can continue to use that variable with groovy highlighting

BasPH:
Aah my bad, must have been a bad copy-paste in that first code example.

I did a bit of research but couldn't find a simple solution. Closing this PR and will create a new one to fix the brackets & indentation that initially led me to add the highlighting.

PR #1806, opened 2023-01-27
Quick fix

fritz-astronomer:
An Astro user should only see `astronomer.io` not `cloud.astronomer.io`


jwitz (approved):
Thank you @fritz-astronomer !

PR #1810, opened 2023-01-27
Update rerunning-dags.md

RNHTTR:
Fix typo. Should also update the tree view image with a grid view image in the "rerun tasks" section


kentdanas (approved):
Awesome, thanks @RNHTTR!

Issue #1278, opened 2022-09-28
Add Taskflow section to branching concept doc


kentdanas:
**Affected Document:** Branching in Airflow concepts doc

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):** Should include a section on how to branch with Taskflow relevant to Airflow 2.3 update. Issue ported from guides repo https://github.com/astronomer/airflow-guides/issues/394

**Additional Notes:**

kentdanas:
Related to #1800

Issue #1800, opened 2023-01-26
Add tabs for task decorator and non-task decorator example DAGs in guides


kentdanas:
**Affected Document:**

**Problem Type (Priority):**

- [ ] Error (Critical): This document has existing information that's incorrect and prevents readers from completing the documented setup (Add the `error/critical` label to this issue)
- [x] Misleading (Important): This document could benefit from additional context, guidance, or clarity (Add the `error/FYI` label to this issue)
- [ ] Typo (Low): Any issue related to spelling/ punctuation/ formatting (Add the `error/typo` label to this issue)

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):** Where relevant, our learn guides should have two tabs for example DAGs - one showing the DAG using task decorators, and another showing with traditional operators. This will be most relevant for DAGs using the PythonOperator.

**Additional Notes:**

PR #1838, opened 2023-02-03
Adding decorator examples to the Branching guide

TJaniF:
Resolves: #1278 
Related to: #1800

The main question is how to word the text when talking about a decorator and operator that are equivalent in their function together. I went with mentioning both together in most cases to be explicit that users can chose freely between the two. But I think the argument could also be made to only mention the traditional operator in the first sentence of each section and have the rest of the text "decorator first".


kentdanas (approved):
This looks good to me. I'll defer to @jwitz on the copy question. Seems like mentioning both together is easiest to highlight their equivalency and not cause confusion, but I guess it does sound a bit clunkier.

learn/airflow-branch-operator.md
```diff
@@ -93,37 +158,97 @@ with DAG(
...
+
+:::info
+
+More examples using the `@task.branch` decorator, can be found on the [Astronomer Registry](https://registry.astronomer.io/dags/example-branch-python-dop-operator-3).
+
+:::
+
+## @task.short_circuit and ShortCircuitOperator
+
+Another option for implementing conditional logic in your DAGs is the `@task.short_circuit` decorator and [ShortCircuitOperator](https://registry.astronomer.io/providers/apache-airflow/modules/shortcircuitoperator). This operator takes a Python callable that returns `True` or `False` based on logic implemented for your use case. If `True` is returned, the DAG will continue, and if `False` is returned, all downstream tasks will be skipped.
```
jwitz, line 180:
```suggestion
Another option for implementing conditional logic in your DAGs is the `@task.short_circuit` decorator and [ShortCircuitOperator](https://registry.astronomer.io/providers/apache-airflow/modules/shortcircuitoperator). This operator takes a Python callable that returns `True` or `False` based on logic implemented for your use case. If `True` is returned, the DAG continues, and if `False` is returned, all downstream tasks are skipped.
```

learn/airflow-branch-operator.md
```diff
@@ -93,37 +158,97 @@ with DAG(
...
+
+The `@task.short_circuit` and `ShortCircuitOperator` is best used in cases where you know that part of your DAG runs only occasionally. For example, maybe your DAG runs daily, but some tasks should only run on Sundays. Or maybe your DAG orchestrates a machine learning model, and tasks that publish the model should only be run if a certain accuracy is reached after training. This type of logic can also be implemented with `@task.branch`, but that operator requires a task ID to be returned. Using the `@task.short_circuit` decorator or `ShortCircuitOperator` can be cleaner in cases where the conditional logic equates to "run or not" as opposed to "run this or that".
```
jwitz, line 182:
```suggestion 
`@task.short_circuit` and `ShortCircuitOperator` are useful when you know that some tasks in your DAG should run only occasionally. For example, maybe your DAG runs daily, but some tasks should only run on Sundays. Or maybe your DAG orchestrates a machine learning model, and tasks that publish the model should only be run if a certain accuracy is reached after training. This type of logic can also be implemented with `@task.branch`, but that operator requires a task ID to be returned. Using the `@task.short_circuit` decorator or `ShortCircuitOperator` can be cleaner in cases where the conditional logic equates to "run or not" as opposed to "run this or that".
```

learn/airflow-branch-operator.md
```diff
@@ -139,26 +264,28 @@ with DAG(
...
     chain(cond_false, *ds_false)
-
 ```
 
-:::info
+</TabItem>
+</Tabs>
 
-The full example code in this section, as well as other examples using the `ShortCircuitOperator`, can be found on the [Astronomer Registry](https://registry.astronomer.io/dags/example-short-circuit-operator).
+In this DAG there are two short circuits, one which will always return `True` and one which will return `False`. When you run the DAG, you can see that tasks downstream of the `True` condition operator ran, while tasks downstream of the `False` condition operator were skipped.
```
jwitz, line 272:
```suggestion
In this DAG there are two short circuits, one which always returns `True` and one which always returns `False`. When you run the DAG, you can see that tasks downstream of the `True` condition operator ran, while tasks downstream of the `False` condition operator were skipped.
```

learn/airflow-branch-operator.md
```diff
@@ -9,60 +9,123 @@ id: airflow-branch-operator
...
+When designing your data pipelines, you may encounter use cases that require more complex task flows than "Task A > Task B > Task C". For example, you may have a use case where you need to decide between multiple tasks to execute based on the results of an upstream task. Or you may have a case where part of your pipeline should only run under certain external conditions. Fortunately, Airflow has multiple options for building conditional logic and/or branching into your DAGs.
 
-In this guide, you'll learn how you can use the `BranchPythonOperator` and `ShortCircuitOperator`, other available branching operators, and additional resources to implement conditional logic in your Airflow DAGs.
+In this guide, you'll learn how you can use `@task.branch` (`BranchPythonOperator`) and `@task.short_circuit` (`ShortCircuitOperator`), other available branching operators, and additional resources to implement conditional logic in your Airflow DAGs.
```
jwitz, line 17:
```suggestion
In this guide, you'll learn how you can use `@task.branch` (BranchPythonOperator) and `@task.short_circuit` (ShortCircuitOperator), other available branching operators, and additional resources to implement conditional logic in your Airflow DAGs.
```
We don't need code ticks around full operator names. This should be fixed throughout the PR.

learn/airflow-branch-operator.md
```diff
@@ -9,60 +9,123 @@ id: airflow-branch-operator
...
 
-## BranchPythonOperator
+## @task.branch and BranchPythonOperator
 
-One of the simplest ways to implement branching in Airflow is to use the [BranchPythonOperator](https://registry.astronomer.io/providers/apache-airflow/modules/branchpythonoperator). Like the [`PythonOperator`](https://registry.astronomer.io/providers/apache-airflow/modules/pythonoperator), the `BranchPythonOperator` takes a Python function as an input. However, the `BranchPythonOperator's` input function must return a list of task IDs that the DAG should proceed with based on some logic. 
+One of the simplest ways to implement branching in Airflow is to use the `@task.branch` decorator or the [BranchPythonOperator](https://registry.astronomer.io/providers/apache-airflow/modules/branchpythonoperator). Like the `@task` decorator and the [`PythonOperator`](https://registry.astronomer.io/providers/apache-airflow/modules/pythonoperator), `@task.branch` and the `BranchPythonOperator` take a Python function as their input. However, the input function to `@task.branch` and to the `BranchPythonOperator` must return a list of valid task IDs that the DAG should proceed with based on some logic.
```
jwitz, line 29:
```suggestion
One of the simplest ways to implement branching in Airflow is to use the `@task.branch` decorator, which is a decorated version of the [BranchPythonOperator](https://registry.astronomer.io/providers/apache-airflow/modules/branchpythonoperator). `@task.branch` accepts any Python function as an input as long as the function returns a list of valid IDs for tasks that the DAG should run after the function completes. 
```
I think you only have to mention `<decorator> = <traditional operator>` once at the first mention of the decorator, after which you only need to refer to the decorator.

learn/airflow-branch-operator.md
```diff
@@ -9,60 +9,123 @@ id: airflow-branch-operator
...
 ## Assumed knowledge
 
 To get the most out of this guide, you should have an understanding of:
 
 - Airflow operators. See [Operators 101](what-is-an-operator.md).
 - Dependencies in Airflow. See [Managing Dependencies in Apache Airflow](managing-dependencies.md).
+- Using Airflow decorators. See [Introduction to Airflow decorators](airflow-decorators.md).
 
-## BranchPythonOperator
+## @task.branch and BranchPythonOperator
```
jwitz, line 27:
```suggestion
## `@task.branch` (BranchPythonOperator)
```

jwitz (approved):
Looks good to me! As for the copy question, I think we only need to mention the decorator and the traditional operator together on the first reference. After that, we only need to reference the decorator. This both reinforces us recommending the TaskFlow API as default and makes the whole doc easier to read.

TJaniF (commented):
Realized I had not submitted my last comment and just merged without 😅 . Thanks for your review @jwitz ! I agree only referencing the decorator makes it much easier to read :)

PR #1878, opened 2023-02-10
Fix links to the public Airflow docs after they were restructured

BasPH:
The Airflow public docs were restructured in https://github.com/apache/airflow/pull/27235, released with Airflow 2.5.1.

Redirect rules were added but don't always function correctly, e.g. https://airflow.apache.org/docs/apache-airflow/stable/concepts/dags.html#airflowignore redirects to https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html (correct page but incorrect section).

This PR points all links to https://airflow.apache.org/docs/apache-airflow/stable/concepts* to the updated location. An added benefit of this PR is that even if a redirect is set correctly, links now point to the correct page, removing redirect latency.


jwitz (approved):
This is hugely helpful. Thank you @BasPH !

PR #1892, opened 2023-02-14
Redshift end_date removal

frankbreetz:
If you include the end_date before today you can't manually trigger the DAG which is very frustrating when you are testing a DAG and don't know the cause


kentdanas (approved):
Looks great, thanks for this improvement @frankbreetz!

PR #1903, opened 2023-02-16
update hosted supported regions

chomipi88:
https://github.com/astronomer/astro/issues/9426


kwcanuck (approved):

jwitz:
@chomipi88 Was this change made on Tuesday? I'm going to add a release note for it

chomipi88:
@jwitz yesterday we updated the regions as part of product packaging updates. no impact to any customers as we don't have any hosted clusters in the region we dropped support for.

jwitz:
Got it, I'll still write a note for the ones that we added.

PR #1911, opened 2023-02-20
Learn code spring syntax cleaning: Basics

TJaniF:
I originally intended this PR to be the alternative code tabs for examples using the PythonOperator / @task decorator (Issue 1800). While adding alternative tabs and swapping `with DAG()` syntax with `@dag` to make the examples TaskFlow first, I started fixing small code format issues like indention or inconsistent use of `'` and `"`. 

Since I made more edits than anticipated I thought it makes sense to split up this PR by Learn chapter. This PR  is covering Basics and the decorators guide (because I originally started there) :) 

Things I standardized on:

- TaskFlow API first: if there are tabs, the TF one is the default, the only exception is the first example in the decorators guide.
- For DAGs that don't use decorated tasks (the only TF item is `@dag`) I did not use tabs. ~~The exception was in the Intro to DAGs guide and in the Scheduling guide. My reasoning is that there the topic is DAGs/DAG parameters, therefore people might be specifically searching for `with DAG()` syntax.~~
- using pendulum over datetime
- using `"` unless there is a reason to use `'`
- ~~flake8 linted formatting (there might be the odd trailing whitespace since I did not run all code snippets through a linter, but the visual format and indentions should be flake8/black conform :) )~~
- moved the code of full DAGs to code-samples and black formatted them

I also changed the copy a bit in the Intro to DAGs and Scheduling guide. :)
Oh and I added a screenshot of a DAG in the intro to DAGs guide and updated a screenshot in the dependencies guide. 

Part of: #1800


learn/airflow-decorators.md
```diff
@@ -28,6 +31,57 @@ Airflow decorators were introduced as part of the TaskFlow API, which also handl
 
 Using decorators to define your Python functions as tasks is easy. Let's take a before and after example. In the "traditional" DAG below, there is a basic ETL flow with tasks to get data from an API, process the data, and store it.
```
kentdanas, line 32: I think the descriptions before and after the code examples here need to be updated a bit to reference the tabs. Also since this is a "before/after", I think it makes sense to have the "Traditional Syntax" DAG be the left most/default tab so that's what shows up first.

learn/dags.md
```diff
@@ -46,84 +49,177 @@ Most DAGs follow this general flow within a Python script:
 
 The following example DAG loads data from Amazon S3 to Snowflake, runs a Snowflake query, and then sends an email.
 
+<Tabs
```
kentdanas, line 52: For this one, I don't think we need a separate TaskFlow API example if it's only the DAG instantiation that is changing. That's actually not really part of the TaskFlow API anyway, it's just a decorator. Also we don't have separate tabs for the other examples you updated to use `@dag` instead of the regular DAG function syntax. 

In general, I think I would leave the separate tabs to examples where the operators can make use of taskflow, and the rest it's fine to leave them using the standard DAG function.

learn/scheduling-in-airflow.md
```diff
@@ -262,7 +265,36 @@ class UnevenIntervalsTimetablePlugin(AirflowPlugin):
 
 Because timetables are plugins, you'll need to restart the Airflow Scheduler and Webserver after adding or updating them.
 
-In the DAG, you can import the custom timetable plugin and use it to schedule the DAG by setting the `timetable` parameter:
+In the DAG, you can import the custom timetable plugin and use it to schedule the DAG by setting the `schedule` parameter (in pre-2.4 Airflow you will need to use the `timetable` parameter):
+
+<Tabs
```
kentdanas, line 270: I'm not sure this tab is needed since the `schedule` parameter and syntax doesn't change at all. I think we likely only need the tabs in cases where it changes the syntax (like different versions of operators where the params, XComs, etc. are treated a bit differently).

learn/what-is-a-hook.md
```diff
@@ -87,74 +89,228 @@ The following example DAG completes the following steps:
 - A second decorated Python operator completes a simple sum check using the results from the first task.
```
kentdanas, line 89: Need to update the text in these bullets to be more general since there are two examples now. e.g. instead of saying "A decorator Python operator" maybe just put "A Python operator"

kentdanas (changes requested):
Thanks for tackling this @TJaniF! Requesting changes because for a few of these I think it's okay to leave them without separate tabs, and I think we should evaluate whether we need to update `DAG` instantiation to `@dag` in every case (I think both are equally valid and we don't need to update to favor `@dag` unless we're stressing decorators specifically, but happy to hear other opinions). Let me know if you want to chat through any of this!

TJaniF (commented):
Thank you for the review @kentdanas ! I changed the text and remove tabs where you suggested. I also switch from in-markdown code to codeblocks for all full DAGs, which made linting a lot easier, should have done that from the start 👀

learn/airflow-decorators.md
```diff
@@ -26,86 +35,30 @@ In general, whether to use decorators is a matter of developer preference and st
...
-import logging
-from datetime import datetime
-
-import requests
-from airflow import DAG
-from airflow.operators.python import PythonOperator
+Using decorators to define your Python functions as tasks is easy. Let's take a before and after example. Under the **Traditional Syntax** tab below, there is a basic ETL DAG with tasks to get data from an API, process the data, and store it. Click on the **Decorators** tab to see the same DAG written using Airflow decorators.
 
-API = "https://api.coingecko.com/api/v3/simple/price?ids=bitcoin&vs_currencies=usd&include_market_cap=true&include_24hr_vol=true&include_24hr_change=true&include_last_updated_at=true"
+The decorated version of the DAG eliminates the need to explicitly instantiate the PythonOperator, has much less code and is easier to read. Notice that it also doesn't require using `ti.xcom_pull` and `ti.xcom_push` to pass data between tasks. This is all handled by the TaskFlow API when you define your task dependencies with `store_data(process_data(extract_bitcoin_price()))`.
```
kentdanas, line 40: Maybe this paragraph should go below the code snippets?

learn/scheduling-in-airflow.md
```diff
@@ -9,6 +9,9 @@ id: scheduling-in-airflow
   <meta name="og:description" content="Get to know Airflow scheduling concepts and different ways to schedule a DAG. Learn how timetables in Airflow 2.2 bring new flexibility to DAG scheduling." />
 </head>
 
+import Tabs from '@theme/Tabs';
```
kentdanas, line 12: I think we don't need these imports anymore since you removed the tabs in this one?

kentdanas (approved):
Looks great! A couple more very minor comments, otherwise good to move to copy review

learn/scheduling-in-airflow.md
TJaniF: Haha, good catch!

TJaniF (commented):
Thanks @kentdanas ! :)
@jwitz: I think this one is ready for you. In hindsight this should have been several PRs, I apologize. I started adding tabs and then just could not keep myself from fixing small things in the code and surrounding text. 😅

jwitz (approved):
Wow! Incredible effort here @TJaniF , I love that we're starting to adhere to a DAG style guide beyond just nice formatting. Do we have an internal resource for a "DAG style guide"? If not, I'd love for us to take some of the rule you mention in the PR and put them on Notion so that I can start editing for these in the future.

My only other note: "Traditional Syntax" sounds good, but why did we switch from "Traditional operators"? I feel like that's a more specific term explaining what the difference is between the two tabs. If you decide to stick with "syntax", let's make that lowercase in the tabs so that it reads "Traditional syntax"

TJaniF:
Thank you @jwitz ! 
I am not aware of any internal DAG style guide, I'd love to get that started (personally I don't have too strong opinions other than heavily favouring having a consistent style and being nitpicky about over-indented code 😅  but I can start writing something up and maybe we can get others to weigh in. For example using `pendulum` over `datetime` was something I saw mentioned on slack as a new unofficial best practice). 
I put what I did in this PR into a Notion Doc and sent you an invite :)

The reason I switched from `Traditional operators` to `Traditional syntax` is because the difference includes `with DAG()` vs `@dag` and `TaskGroup` vs `@task_group, which aren't really operators and I wanted to avoid any confusion there, so I chose a less specific term. Agree with making it lowercase :)

PR #1914, opened 2023-02-21
Learn code spring syntax cleaning: DAGs.

TJaniF:
Part 2! 😄 

This PR does essential the same as  #1911 but for all remaining guides in the DAGs category.

- moved all full DAGs into code-samples and referenced them as code blocks
- where PythonOperator or @task decorators were used I provided tabs with both syntax versions where possible (the one code snippet where that does not work was task group mapping, that only works with the @task_group decorator). 
- switched from datetime to pendulum
- changed wording around code that got new tabs in a few places

Other:
- fixed airflow_passing_data_between_tasks_s3.py, the DAG was breaking for cat facts that had a comma 🐱 

Part of: #1800


learn/airflow-passing-data-between-tasks.md
```diff
@@ -82,12 +85,35 @@ You can see that these limits aren't very big. And even if you think your data m
 
 In this section, you'll review a DAG that uses XCom to pass data between tasks. The DAG uses XComs to analyze cat facts that are retrieved from an API. To implement this use case, the first task makes a request to the [cat facts API](http://catfact.ninja/fact) and pulls the `fact` parameter from the results. The second task takes the results from the first task and performs an analysis. This is a valid use case for XCom, because the data being passed between the tasks is a short string.
 
+<Tabs
+    defaultValue="taskflow"
+    groupId= "xcom-example"
+    values={[
+        {label: 'TaskFlow API', value: 'taskflow'},
+        {label: 'Traditional Syntax', value: 'traditional'},
```
kentdanas, line 93: Let's change the tab here to "Traditional syntax" (lowercase "s") to be consistent with the previous PR.

learn/dag-best-practices.md
```diff
@@ -9,6 +9,11 @@ id: dag-best-practices
   <meta name="og:description" content="Keep up to date with the best practices for developing efficient, secure, and scalable DAGs using Airflow. Learn about DAG design and data orchestration." />
 </head>
 
+import CodeBlock from '@theme/CodeBlock';
```
kentdanas, line 12: Since we're switching to tabs in this PR - do you think it would be better to present the "good/bad" examples in this guide in tabs as opposed to one following the other? Seems to me like that might be cleaner.

learn/dynamic-tasks.md
```diff
@@ -5,6 +5,13 @@ description: "How to dynamically create tasks at runtime in your Airflow DAGs."
 id: dynamic-tasks
```
kentdanas, line 5: Wondering if it would make sense to put the subsections of "Mapping over the result of another operator" section into tabs instead. This would make the guide shorter and in my opinion easier to read/navigate. The only downside I see is we wouldn't have the subsections in the right-hand menu, but I'm not sure that would be a big deal. Thoughts?

learn/rerunning-dags.md
```diff
@@ -5,6 +5,9 @@ description: "How to catchup, backfill, and clear task instances in Airflow."
 id: rerunning-dags
 ---
 
+import Tabs from '@theme/Tabs';
```
kentdanas, line 8: Looks like we don't use tabs on this one, can we remove this import?

kentdanas (approved):
Looks great @TJaniF! A couple of minor comments, including some that would make the PR "bigger" (I know you were trying not to make it too overwhelming) but I think could still make sense to include here. Let me know if you want to chat about any of those suggestions - otherwise this is good from my side!

learn/dag-best-practices.md
TJaniF: Looks a lot better :)

learn/dynamic-tasks.md
TJaniF: I think it is better with tabs, a lot easier to compare the syntax version this way :)

TJaniF (commented):
Thank you @kentdanas ! Great suggestions, needed some small changes in the top-level code section of the best practices guide (there was no good practices version of the DAG and I needed to change the example a little to focus in on top level code without having loop vs dynamic task mapping too). I think it looks cleaner now. :) 

@jwitz Part 2 ready for you 😄

learn/airflow-passing-data-between-tasks.md
```diff
@@ -82,12 +85,35 @@ You can see that these limits aren't very big. And even if you think your data m
...
+        {label: 'Traditional syntax', value: 'traditional'},
+    ]}>
+
+<TabItem value="taskflow">
+
+<CodeBlock language="python">{airflow_passing_data_between_tasks_taskflow}</CodeBlock>
+
+You can use the [TaskFlow API](https://airflow.apache.org/docs/apache-airflow/stable/tutorial_taskflow_api.html) to push and pull values to and from XCom. To push a value to XCom return it at the end of your task as with traditional operators. To retrieve a value from XCom provide the object created by the upstream task as an input to your downstream task.
+
+Using the TaskFlow API usually requires less code to pass data between tasks than working with the traditional syntax.
```
jwitz, line 102: Any reason we have this after the code block? I think it makes more sense to have before and to have a sentence introducing the DAG.

learn/cross-dag-dependencies.md
```diff
@@ -167,89 +147,26 @@ In Airflow 2.2 and later, a deferrable version of the ExternalTaskSensor is avai
...
+</TabItem>
+
+<TabItem value="traditional">
+
+<CodeBlock language="python">{external_task_sensor_example_traditional}</CodeBlock>
+
+</TabItem>
+</Tabs>
 
 In this DAG
```
jwitz, line 171:
```suggestion
In this DAG:
```

learn/dag-best-practices.md
```diff
@@ -88,134 +97,61 @@ In the context of Airflow, top-level code refers to any code that isn't part of
...
-
-    for grocery_item in result:
-        query = PostgresOperator(
-            task_id='query_{0}'.format(result),
-            postgres_conn_id='postgres_default',
-            sql="INSERT INTO purchase_order VALUES (value1, value2, value3);"
-        )
-```
+In the **Bad practice** example the connection to the other database is made outside of an operator instantiation as top-level code. When the scheduler parses this DAG, it will use the `hook` and `result` variables to query the `grocery_list` table. This query is run on every Scheduler heartbeat, which can cause performance issues. 
+
```
jwitz, line 103:
```suggestion
```

learn/dag-best-practices.md
```diff
@@ -88,134 +97,61 @@ In the context of Airflow, top-level code refers to any code that isn't part of
...
 
 ### Treat your DAG file like a config file
 
 Including code that isn't part of your DAG or operator instantiations in your DAG file makes the DAG harder to read, maintain, and update. When possible, leave all of the heavy lifting to the hooks and operators that you instantiate within the file. If your DAGs need to access additional code such as a SQL script or a Python function, consider keeping that code in a separate file that can be read into a DAG run.
 
-The following example DAG demonstrates what you shouldn't do. A SQL query is provided directly in the `PostgresOperator` `sql` param:
+The following example DAGs demonstrate the difference between providing a SQL query directly to the PostgresOperator as shown in the tab **Bad practice** and referencing a separate SQL file under the tab **Good practice**. In the good practice DAG, the DAG-level configuration includes `template_searchpath` and the PostgresOperator specifies a `covid_state_query.sql` file that contains the same query as was directly provided to the PostgresOperator in the bad practice DAG.
```
jwitz, line 133:
```suggestion
The following example DAGs demonstrate the difference between the bad and good practices of including code in your DAGs. In the **Bad practice** DAG, a SQL query is provided directly to the PostgresOpertator, which unnecessarily exposes code in your DAG. In the **Good practice** DAG, the DAG-level configuration includes `template_searchpath` and the PostgresOperator specifies a `covid_state_query.sql` file that contains the query to execute. 
```

learn/dynamic-tasks.md
```diff
@@ -80,101 +118,103 @@ You can use the output of an upstream operator as the input data for a dynamical
...
 In this section you'll learn how to pass mapping information to a downstream task for each of the following scenarios:
 
-- Both tasks are defined using the TaskFlow API.
-- The upstream task is defined using the TaskFlow API and the downstream task is defined using a traditional operator.
-- The upstream task is defined using a traditional operator and the downstream task is defined using the TaskFlow API.
-- Both tasks are defined using traditional operators.
+- **TaskFlow over TraskFlow**: Both tasks are defined using the TaskFlow API.
+- **TaskFlow over Traditional operator**: The upstream task is defined using a traditional operator and the downstream task is defined using the TaskFlow API.
+- **Traditional operator over TaskFlow**: The upstream task is defined using the TaskFlow API and the downstream task is defined using a traditional operator.
+- **Traditional operator over Traditional operator**: Both tasks are defined using traditional operators.
```
jwitz, line 124:
```suggestion
- **TaskFlow over traditional operator**: The upstream task is defined using a traditional operator and the downstream task is defined using the TaskFlow API.
- **Traditional operator over TaskFlow**: The upstream task is defined using the TaskFlow API and the downstream task is defined using a traditional operator.
- **Traditional operator over traditional operator**: Both tasks are defined using traditional operators.
```

learn/dynamic-tasks.md
```diff
@@ -80,101 +118,103 @@ You can use the output of an upstream operator as the input data for a dynamical
...
+
+<Tabs
+    defaultValue="two-flow"
+    groupId= "mapping-over-results"
```
jwitz, line 128:
```suggestion
    groupId= "mapping-over-the-result-of-another-operator"
```

jwitz: Shoooooot, I totally forgot to edit this in the previous PRs, but the groupId should always be identical to the topic that the tabs are in. This is so our website can use the group Id to structure a link for sharing specific tabs.

jwitz (commented):
Just a few comments, realized that these PRs need to have a lot of the groupId's rewritten so that our site's tabbing functionality works. Lmk if you want to split this work!

TJaniF:
Thank you @jwitz :) 
GroupIDs should now be fixed! Let me know if you want to have another look or want me to merge this

jwitz (approved):
This now LGTM!

PR #1921, opened 2023-02-22
Upgrade Docusaurus, fix tab behavior

jwitz:
Annoyingly the new Docusaurus tab feature is just a worse version of the one we have now, refactored our code so we have the good stuff. Maybe we can contribute our header functionality to the open source in the future


jwitz:
@katiebrady-astro I had to do a little work to keep our tab-setting functionality going (which is unfortunately still more practical than Docusaurus' implementation, see https://docusaurus.io/changelog/2.3.0). Can you check the `set-tabs` script to make sure that nothing broke as a result of the upgrade there? Thank you!

jwitz:
Tested the tabbing behavior a bit and everything seems to be consistent with before, merging now

PR #1928, opened 2023-02-24
Clarify that new Azure AD setup is only for Teams

jwitz:


software/integrate-auth-system.md
```diff
@@ -129,7 +129,9 @@ Follow these steps to configure Azure AD as your OIDC provider.
 
 ![authentication.png](/img/software/azure-authentication.png)
 
-#### Create a client secret
+#### Create a client secret (Optional)
+
+Complete this setup only if you are importing groups of users to Astronomer Software as [Teams](import-idp-groups.md).
```
jwitz, line 134:
```suggestion
Complete this setup only if you want to import Azure AD groups to Astronomer Software as [Teams](import-idp-groups.md).
```

odaneau-astro (approved):

software/integrate-auth-system.md
```diff
@@ -152,6 +154,7 @@ Follow these steps to configure Azure AD as your OIDC provider.
     - In **Customize token properties by type** area, expand **ID**, **Access**, and **SAML** and then select **Group ID** for each type.
     
 7. Click **Add**.
+8. Encrypt the secret value you copied as a Kubernetes Secret on your Astronomer installation. See [Store and encrypt identity provider secrets](#store-and-encrypt-identity-provider-secrets).
```
kwcanuck, line 157:
```suggestion
8. Encrypt the secret values you copied in step 5 as Kubernetes Secrets. See [Store and encrypt identity provider secrets](#store-and-encrypt-identity-provider-secrets).
```
If you want to keep _on your Astronomer installation_, add _Software_ - _on your Astronomer Software installation_. I'm not sure how secrets can be kept _on_ a software installation. Maybe _with your_?

jwitz: They can be kept on the Astronomer namespace, which is where you install Astronomer on Kubernetes. I can update the wording, but this is accurate in Kubernetes terms.

kwcanuck: @jwitz it reads better with the edit. Also, I'm wondering why you removed the reference to the step where the values were copied.  In multi-step procedures, it helps users who scan locate related procedure steps quickly. After six steps, they might not recall what values they copied or where. These references can be a maintenance issue, but it's often worth it. 

You'll probably want to update the procedure numbering as it's no longer correct:

image

Thankfully, the CSS renders it correctly.

PR #1949, opened 2023-02-28
add note on scaling default worker queue to zero

jaketf:
If we don't want to document this publicly I understand, but figured it was at least worth discussing in a PR.

cc @paolaperaza 

context in:
https://github.com/astronomer/astro/pull/10044
https://github.com/astronomer/airflow-operator/pull/1404

### DIFF
--- a/astro/configure-worker-queues.md
+++ b/astro/configure-worker-queues.md
@@ -54,12 +54,9 @@ If you don’t change any settings in the default worker queue:
 
 You can change all settings of the default worker queue except for its name.
 
-#### Scaling default worker queue to zero
-If you configure your default worker queue to have a minimum workers of zero there is a subtle additional side effect:
-the Astronomer monitoring DAG will be configured to run on separate queue that is listened to only by a tiny worker sidecar
-in the scheduler pod. The monitoring DAG will still serve it's purpose of making sure tasks are able to run on your deployment.
-The reason for this subtle difference, is that the default worker queue cannot effectively scale to zero if the monitoring DAG
-is creating tasks every 5 mins.
+#### Scaling the default worker queue to zero
+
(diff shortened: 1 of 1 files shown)


astro/configure-worker-queues.md
```diff
@@ -54,6 +54,13 @@ If you don’t change any settings in the default worker queue:
 
 You can change all settings of the default worker queue except for its name.
 
+#### Scaling default worker queue to zero
+If you configure your default worker queue to have a minimum workers of zero there is a subtle additional side effect:
+the Astronomer monitoring DAG will be configured to run on separate queue that is listened to only by a tiny worker sidecar
+in the scheduler pod. The monitoring DAG will still serve it's purpose of making sure tasks are able to run on your deployment.
+The reason for this subtle difference, is that the default worker queue cannot effectively scale to zero if the monitoring DAG
+is creating tasks every 5 mins.
```
jwitz, line 62:
```suggestion
#### Scaling the default worker queue to zero

If you configure your default worker queue to have a **Minimum Worker Count** of 0, the `astronomer_monitoring_dag` runs on a separate, non-configurable worker queue that always has an available worker for running the DAG. You will not be charged for any task runs that occur on this queue.
```

jwitz (commented):
Sorry for the late review on this, I think this is great info to have @jaketf but will defer to @paolaperaza who has a little more product strategy knowledge here. I've also written an issue for us to include more information about the monitoring DAG in our docs (it's currently only in release notes)

astro/configure-worker-queues.md
jaketf: @jwitz might be worth mentioning that this non-configurable queue is "resource usage" or "cost" optimized if we're not going to leak the implementation detail of it being on the scheduler pod as a sidecar.

jwitz:
@jaketf I added a release note for this based on your PR, and Ash had some pushback: https://github.com/astronomer/docs/pull/2066#discussion_r1151018554

Am I missing some nuance here? Also, do we have an estimation on when this will make it to GA?

jwitz:
Hey @jaketf never saw this get pushed to prod, so closing out for now. Feel free to reopen/ edit if it's been implemented!

PR #1965, opened 2023-03-03
Create external link checking GitHub action

jwitz:
Lychee seems to have the support and options we need to make a truly useful external broken link checker. I also think it makes sense to run this as a cron job rather than as a PR check - it's rare to ship a broken link in new content and we don't want ancient links to prevent PRs from merging.

### DIFF
--- a/.github/workflows/linkCheck.yml
+++ b/.github/workflows/linkCheck.yml
@@ -18,7 +18,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
         with: 
-          args: --scheme 'https' --verbose --no-progress './**/*.md' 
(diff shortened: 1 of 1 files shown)


.github/workflows/linkCheck.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v1.6.1
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        with: 
+          args: --scheme 'https' --verbose --no-progress './**/*.md'
```
jwitz, line 21:
```suggestion
          args: --scheme 'https' --verbose --no-progress './astro/*.md' './software/*.md'  './learn/*.md'
```

jwitz:
Since this is a cron and won't block publication, just gonna merge and see how it goes!

Issue #1954, opened 2023-03-01
Mention Podman prerequisites in "Install the CLI"


jwitz:
**Affected Document:** https://docs.astronomer.io/astro/cli/install-cli

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**

We now have [docs](https://docs.astronomer.io/astro/cli/configure-cli#run-the-astro-cli-using-podman) that explain how to set up Podman as your primary container engine for running Airflow locally. However, our install docs only include the prerequisites for using Docker. 

We should add a note in the "Install" prerequisites (for all OSes) that explain there are different prerequisites for running the CLI with Podman. We can then link out to our new Podman docs to explain the prerequisites/ setup steps.

PR #1967, opened 2023-03-03
Add Podman Astro CLI  pre-req notes to install instructions

lzdanski:
- Added note to each OS install instructions to reference pre-reqs and config instructions for Podman ([Preview](https://deploy-preview-1967--gallant-neumann-56599d.netlify.app/astro/cli/install-cli?tab=mac#install-the-astro-cli))
- Resolves https://github.com/astronomer/docs/issues/1954


astro/cli/install-cli.md
```diff
@@ -35,6 +35,8 @@ Install the Astro CLI on a Mac operating system with a single command.
 - [Homebrew](https://brew.sh/)
 - [Docker Desktop](https://docs.docker.com/get-docker/) (v18.09 or higher).
 
+By default, the Astro CLI uses Docker as its container management engine. However, if your organization uses Podman to run and manage containers, you can configure the Astro CLI to use it instead. See [Run the Astro CLI with Podman](/astro/cli/configure-cli?tab=mac#run-the-astro-cli-using-podman) for prerequisites and configuration steps.
```
jwitz, line 38:
```suggestion
By default, the Astro CLI uses Docker as its container management engine. However, if your team uses Podman to run and manage containers, you can configure the Astro CLI to use it instead. See [Run the Astro CLI with Podman](/astro/cli/configure-cli?tab=mac#run-the-astro-cli-using-podman) for prerequisites and configuration steps.
```
Because "Organization" is an RBAC component on Astro, I try to use an alternative word when discussing orgs in the general sense.

Also it looks like you have found another way to structure relative links 😄 I would usually go `cli/configure-cli.md?tab=mac#run-the-astro-cli-using-podman`. I could totally see us switching to your method if it works for all links, but I do think we should stick to one format so that it's easier for other contributors to know how to structure their links.

jwitz (approved):
LGTM! Just one minor comment that applies to all of the notes. Congrats on your first PR in `docs`  🥳

PR #1972, opened 2023-03-06
Fix pre-installed providers packages list

BasPH:
This PR:

- Links to the correct PyPi URL for the Snowflake providers package
- Removes 2 links (StatsD & Virtualenv) that aren't providers packages

In this list: https://docs.astronomer.io/astro/runtime-image-architecture#provider-packages


jwitz (approved):
Thank you Bas!

Issue #1922, opened 2023-02-22
Be more clear about not supporting Azure Single Server


danielhoherd:
**Affected Document:**

Probably https://docs.astronomer.io/software/install-azure

**Problem Type (Priority):**

- [ ] Error (Critical): This document has existing information that's incorrect and prevents readers from completing the documented setup (Add the `error/critical` label to this issue)
- [X] Misleading (Important): This document could benefit from additional context, guidance, or clarity (Add the `error/FYI` label to this issue)
- [ ] Typo (Low): Any issue related to spelling/ punctuation/ formatting (Add the `error/typo` label to this issue)

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**

A blurb saying that we do not support Azure Single Server would suffice. 

**Additional Notes:**

A customer reported an error that I investigated for a long time before CRE told me that Azure Single Server is an unsupported configuration. Ticket with that discussion here: https://github.com/astronomer/astronomer/issues/1756

CC @rbankston

PR #1982, opened 2023-03-07
Add admonition about Azure Single Server

lzdanski:
* User had issues with setting up Azure with Astronomer Software, but after a long troubleshooting process, realized they were using Azure Single Server which is not supported. 
* Added admonition to list of prerequisites in Install Azure page (https://docs.astronomer.io/software/install-azure)


jwitz:
This LGTM! Tagging @danielhoherd for review since he wrote the original issue. We should also close out https://github.com/astronomer/docs/issues/1922 when this merges

jwitz (approved):

danielhoherd (approved):
Short and sweet. LGTM.

PR #1989, opened 2023-03-08
Small changes updating the Astro SDK tutorial

TJaniF:
While revisiting the tutorial to test changes in #1931 I noticed some outdated information:

- The instructions for env vars to set in order to permit for serialization of SDK objects have changed, now 2.5 only needs one env vars adjusting the normal XCom backend, 2.4 and earlier still needs 3 setting a custom XCom backend.
- The AWS connection of type `S3` does not exist in newer aws provider versions, the connection type now used for S3 is called `Amazon Web Services` and has a direct field for the key and secret.
- The Snowflake connection form does not have a `host` field anymore, instead account and region have to be provided in their own fields. 
- Reformatted the SQL code to be more compact with less side-scroll

- The example DAG was written back when pickling was necessary to use the SDK. This is why returning a Pandas series did not use to be an issue. The newer XCom env vars handle SDK specific objects like Tables but not a pandas series apparently (I tried both, the adjusted Airflow XCom backend and the `astro.custom_backend.astro_custom_backend.AstroCustomXcomBackend`. This is why I added `.to_json()` on the return value of the @dataframe task :)


kentdanas (approved):
Thanks for this @TJaniF - I'm preemptively approving, but I think we should ask in the SDK channel about the serialization thing. I don't think the `.to_json()` _should_ be required based on my understanding of how the SDK works and since I didn't have to do this with dataframes using the local metadb backend I want to make sure we're not overlooking a bug or something else we should be instructing users on.

learn/astro-python-sdk.md
```diff
@@ -110,22 +113,22 @@ Now that you have a staging table in Snowflake and some example data ready to lo
...
-    - Login:
-    - Password:
-    - Account:
+    - Login: `<your_login>`
+    - Password: `<your_password>`
+    - Account: `<your_account>`. This is the account name found in the URL where you can log into Snowflake, it has the format xy12345.
+    - Warehouse: `ASTRO_SDK_DW`
     - Database: `ASTRO_SDK_DB`
-    - Region: (something like `us-east-1` or `us-central1.gcp`)
+    - Region: `<your_region>` (something like `us-east-1` or `us-central1.gcp`)
```
jwitz, line 130:
```suggestion
    - Region: `<your_region>` (for example, `us-east-1` or `us-central1.gcp`)
```

learn/astro-python-sdk.md
```diff
@@ -84,19 +84,22 @@ Now that you have a staging table in Snowflake and some example data ready to lo
...
-    If you're using Airflow 2.4 or earlier, additionally set the following environment variable to use a required custom XCom backend:
+    If you're using Airflow 2.4 or earlier, set the following environment variables instead to use a required custom XCom backend:
     
     ```
+    AIRFLOW__ASTRO_SDK__XCOM_STORAGE_CONN_ID=<your_aws_conn>
+    AIRFLOW__ASTRO_SDK__XCOM_STORAGE_URL='s3://<your-bucket>/xcom/'
     AIRFLOW__CORE__XCOM_BACKEND='astro.custom_backend.astro_custom_backend.AstroCustomXcomBackend'
     ```
 
+    Learn more about how the Astro SDK uses XCom backends in the [Astro SDK documentation](https://astro-sdk-python.readthedocs.io/en/stable/guides/xcom_backend.html).
```
jwitz, line 101:
```suggestion
    Learn more about how the Astro SDK uses XCom backends using the [Astro SDK documentation](https://astro-sdk-python.readthedocs.io/en/stable/guides/xcom_backend.html).
```

jwitz (approved):
LGTM!

Issue #1985, opened 2023-03-07
Update recommendations on `/include` + Helper Functions


paolaperaza:
Originally reported by @ReadytoRocc: https://astronomer.slack.com/archives/C015V2JFKT5/p1677507794199359

### Context

Astro docs currently have misleading recommendations for how customers are supposed to add utility files (such as CSVs, python scripts, SQL code files) to their Astro project.

We tell them to do 2 things:

- Add a `helper-function` directory: https://docs.astronomer.io/astro/develop-project#add-dag-helper-functions
- Add utility files to `/include`: https://docs.astronomer.io/astro/develop-project#add-dag-utility-files

These sections should be the same and give users the latest product feedback.

### Ask

- [ ] Simplify to one section called `Add Utility Files` and specify that there are two categories of utility files:
       - Files that are needed by and references your DAG code (e.g. python scripts, SQL code files)
       - Files that are needed to be built in your Docker image (e.g  custom operators, test data, etc)
- [ ] Clarify our recommendations:
        - Create a folder called `/utils` in your DAGs directory for things your DAG need
        - Add anything else to your `/include` folder

There's some complexity here, so make sure to get a review from: @cmarteepants @sunkickr @jlaneve @ReadytoRocc

jwitz:
As I'm working through this issue I have a nit: I don't think `include` is for utility files - we should have a separate section explaining that include is for local testing only.

PR #1990, opened 2023-03-08
Various improvements to Develop a Project

jwitz:
Resolves https://github.com/astronomer/docs/issues/1985

I took a crack at improving this doc across the board. A few significant changes:

- Removed "Apply changes to your project" because those steps are already included in every mentioned file's documentation, having a centralized doc with only this information feels redundant - especially because we already have steps on starting and restarting projects
- Moved some trickier configs into an "Advanced configuration" section so it's more clear to new users that they don't need to worry about compose overrides or custom Docker images.
- Created new topics for "Utility files", "Plugins", and "Custom data and utilities for testing". The plugin content is net new, and the others are reworked, simplified versions of existing topics
- Moved "Specify files to ignore" out of the DAGs setup, as its not totally related besides the setup existing in the DAGs folder

### DIFF
--- a/astro/develop-project.md
+++ b/astro/develop-project.md
@@ -56,7 +56,7 @@ The command builds your Astro project into a Docker image and creates the follow
 - **Scheduler:** The Airflow component responsible for monitoring and triggering tasks
 - **Triggerer:** The Airflow component responsible for running triggers and signaling tasks to resume when their conditions have been met. The triggerer is used exclusively for tasks that are run with [deferrable operators](https://docs.astronomer.io/learn/deferrable-operators).
 
-After the project builds, you can access the Airflow UI by going to `http://localhost:8080/` and logging in with `admin` for both your username and password. You can also access your Postgres database at `localhost:5432/postgres`. As long as your Airflow environment is running, any changes you make in your `dags`, `plugins`, and `include` directories are automatically applied without needing to restart the environment. 
+After the project builds, you can access the Airflow UI at `http://localhost:8080/`. You can also access your Postgres database at `localhost:5432/postgres`. As long as your Airflow environment is running, any changes you make in your `dags`, `plugins`, and `include` directories are automatically applied without needing to restart the environment. 
 
 ### Restart your local environment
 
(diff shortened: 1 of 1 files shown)


fritz-astronomer (approved):

astro/develop-project.md
```diff
@@ -75,55 +91,17 @@ Use the `astro run <dag-id>` command to run and debug a DAG from the command lin
...
 
-### Add DAG helper functions
+If you're using [DAG-only deploys](https://docs.astronomer.io/astro/deploy-code#deploy-dags-only) on Astro, changes to this folder are deployed when you run `astro deploy --dags` and do not require restarting your Deployment. 
 
-Use the `include` folder to store additional utilities required by your DAGs. For example, templated SQL scripts or a custom helper function that's used in an operator and that you can reference in multiple DAGs.
+## Add plugins 
 
-Astronomer recommends storing the `include` folder inside the `dags` directory of your Astro project. If you're using [DAG-only deploys](deploy-code.md#deploy-dags-only), this allows you to deploy DAG code without an image restart by running `astro deploy --dags` and makes sure that your DAGs can access your utility files when you deploy them.
+Airflow plugins enable custom features in Airflow, such as new pages in the Airflow UI or custom alerting and monitoring tools. To use an Airflow plugin, add your plugin files to the `plugins` folder of your Astro project. See [Airflow plugins](https://docs.astronomer.io/learn/using-airflow-plugins) for examples of creating and using Airflow plugins.
```
paolaperaza, line 102:
```suggestion
If you need to build a custom view in the Airflow UI or build an application on top of the Airflow metadata database, you can use Airflow plugins. To use an Airflow plugin, add your plugin files to the `plugins` folder of your Astro project. A small number of Astro users need this feature.

To learn more about Airflow plugins and how to build them, see [Airflow Plugins](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/plugins.html) in Airflow documentation or the [Airflow plugins](https://docs.astronomer.io/learn/using-airflow-plugins) guide.
```

I think it's ok to explicitly state that most folks don't need this, and frame a "don't worry about it" tone? Also -- these are rare enough that it should be closer to the bottom of the doc

astro/develop-project.md
```diff
@@ -139,6 +117,12 @@ When you add Airflow objects to the Airflow UI of a local environment or to your
 
 Astronomer recommends using the `airflow_settings.yaml` file so that you don’t have to manually redefine these values in the Airflow UI every time you restart your project. To ensure the security of your data, Astronomer recommends [configuring a secrets backend](secrets-backend.md).
 
+## Add custom data and utilities for testing (Local development only)
+
+Use the `include` folder of your Astro project to store files for testing locally, such as custom operators or test data. These files are not included when you deploy your project to Astro.
+
+If you're running Airflow locally, apply your changes by refreshing the Airflow UI.
```
paolaperaza, line 124: The difference between `/include` and `/utils` is still confusing to me 😞  Should we combine these sections to explicitly define the difference?

What is include for local dev only?

astro/develop-project.md
```diff
@@ -139,6 +117,12 @@ When you add Airflow objects to the Airflow UI of a local environment or to your
...
+
 ### Configure `airflow_settings.yaml` (local development only)
```
paolaperaza, line 126:
```suggestion
### Configure `airflow_settings.yaml` (Local development only)
```

astro/develop-project.md
```diff
@@ -282,7 +257,33 @@ my_project
     └── prod.env
 ```
 
-## Run commands on build
+## Advanced configuration
```
paolaperaza, line 260: Plugins should be here. Or TBH -- we prob just shouldn't have the `/plugins` folder in the default Astro directory cc: @sunkickr

paolaperaza (changes requested):
This looks good, @jwitz. Requesting changes for two main reasons:

- I still think our framing for `/include` vs `/utils` is confusing, especially for a first-time user
- Let's modify the framing of Airflow plugins and move them to the bottom of the doc

I like the "advanced configuration" section. Also, we're removing the section on helper functions (that's cool), but I still think we should probs have a scaffolding of an Astro project. Or at least, a more explicit link here? https://docs.astronomer.io/astro/create-project

paolaperaza (approved):
@jwitz I re-worked a bit of the `utils` folder part.. let's see where we land with `/include` and modify as needed

PR #1992, opened 2023-03-08
Migrate to Astro from MWAA/GCC

fritz-astronomer:
Looking to create a document that ADEs can follow along with customers (similar to the [software install](https://docs.astronomer.io/software/install-aws) docs that solutions used to do with customers) during our new offering for a 3h lift-n-shift migration

rough draft here:
https://www.notion.so/astronomerio/3-Hour-Migration-Agenda-Template-96a8ed25d60448c29a467bb6d3b2dd0b#d53fb1a82b0a4e6aaeb1ea3385c72900


jwitz:
@fritz-astronomer Based on our discussion in Slack, I broke out this content into separate docs for each tool you might migrate from. Because you said that a lot of content would be the same between tools, I created a new doc called `_migration-setup.mdx` that will be imported into each doc and starts from Step 2. 

I intentionally didn't edit any of the core content past the prerequisites because I know you're still working on it, but let me know if you want me to get started on that!

iamakashladdha:
@fritz-astronomer, review comments for MWAA, same apply for GCC

1. Step 6: Copy core Airflow code
Mapping of Airflow Version 2.0.2 with Runtime Version 3.0.4 needs to be removed
Airflow Version for 4.2.9 Runtime needs to be corrected as 2.2.5 in the Column
Reference Link for Mapping needs to be Updated from Release notes to [this](https://docs.astronomer.io/astro/runtime-version-lifecycle-policy#stable-releases)
A mapping of Astronomer Runtime releases to Airflow versions is available [in the Release Notes](https://docs.astronomer.io/astro/release-notes).

2. Step 3: Create Astro Deployment
INFO for Astronomer runtime needs to be corrected as above (#1), It would make sense to keep this Runtime Info only once in the Step3, It may be removed from Step6
For Scheduler Size Info Dialog box do you want to remove (Small/Medium and Large) and rather keep the respective CPU and Memory for the AU’s mentioned

3. [DAGs-only Deployments](https://deploy-preview-1992--gallant-neumann-56599d.netlify.app/astro/migrate-mwaa#dags-only-deployments) - You may want to change the term and it occurrences in [this](https://deploy-preview-1992--gallant-neumann-56599d.netlify.app/astro/migrate-mwaa#dags-only-deployments) section it to DAGs-only Deploy

4. Step 6: Copy core Airflow code
Airflow to Astronomer Runtime Mapping can be removed from here (as mentioned above)
Since Astro Deployment is already created in the Step#3, so here we should just tell them to Edit Docker File to Match the Runtime Version selected while creating the Deployment in Step#3.
Edit Dockerfile to match Astro Runtime version to customer's Source environment Airflow version this isn’t needed, since if you cannot downgrade a deployment anyway using docker, they just need to use what they created in Step#3.

5. Step 2: Create Astro Workspace
For Manager Astro Users and Manage Astro User Permissions, Do you want to refer to respective workspace section links instead of the entire page which has organization level access as well?

6. Need to double check on the Hyperlinks, as some are pointing to the Test Version (https://deploy-preview-1992--gallant-neumann-56599d.netlify.app/) while some are to main ([docs.astronomer.io](http://docs.astronomer.io/)), Example - https://deploy-preview-1992--gallant-neumann-56599d.netlify.app/astro/manage-workspaces . Not sure if they get auto fixed when merged with main.

magdagultekin:
Two little things for MWAA:
1. In the Validate, Cut-over and Tune Process section, the pictures refer to env vars, not DAGs cutover tab. 
2. In Step 5 I think you wanted to link another section but it looks like this instead:

> You can view more detailed instructions and information at the [create-project.md#step-1-create-an-astro-project](Create Project) documentation.

fritz-astronomer:
> Need to double check on the Hyperlinks, as some are pointing to the Test Version (https://deploy-preview-1992--gallant-neumann-56599d.netlify.app/) while some are to main ([docs.astronomer.io](http://docs.astronomer.io/)), Example - https://deploy-preview-1992--gallant-neumann-56599d.netlify.app/astro/manage-workspaces . Not sure if they get auto fixed when merged with main.
@iamakashladdha the links will change to `docs.astronomer.io` when it's deployed 👍 
@jwitz - this was actually a question for you, though, I couldn't figure out how to link to a category? like
https://docs.astronomer.io/astro/category/connect-astro
^ this page rather than the deep link. Do you know how?

fritz-astronomer:
> * For Scheduler Size Info Dialog box do you want to remove (Small/Medium and Large) and rather keep the respective CPU and Memory for the AU’s mentioned
> *

fritz-astronomer:
@iamakashladdha -

> Mapping of Airflow Version 2.0.2 with Runtime Version 3.0.4 needs to be removed

"Needs to be removed" -Those are versions that MWAA/GCC use 😢  I'm not entirely sure how to handle that. I don't wanna update customers, if we can just move them directly.  I'd hope that we can just support a version in the 3 and 4 series 😢 

> 2. Step 3: Create Astro Deployment ... For Scheduler Size Info Dialog box do you want to remove (Small/Medium and Large) and rather keep the respective CPU and Memory for the AU’s mentioned

 I was intentionally using Small/Medium/Large to match GCC and MWAA's terminology, that I imagine they'll be familiar with. Thoughts?
 
 ---
 Thanks Akash and Magda for the proofreads! Yall are awesome

fritz-astronomer:
Also @jwitz - Is there a way to hide the two non-completed migration pages from navigation ("from oss" and "from astro")? I'd like to ship this PR and then we can work on those later

iamakashladdha:
@fritz-astronomer  for Step3 Create Astro Deployment, 

Do you want to get away with Scheduler Size (Small, Medium and Large) or include another column for what does AU translates in terms of Memory and CPU?
Per MWAA Environment Class, 
Small Instance would be 1 CPU for Scheduler
Medium would be 2 CPU 
and Large would be 4 CPU
Per above if you would want to have CPU mapping then it would imply
having 10 AU for Small
20 AU for Medium
40 AU for Large (since in Astro we can go max to 24 AU only for Scheduler, so may wanna keep Large to 24 then)

Thoughts?

iamakashladdha:
@fritz-astronomer 
> Mapping of Airflow Version 2.0.2 with Runtime Version 3.0.4 needs to be removed

> "Needs to be removed" -Those are versions that  MWAA/GCC use 😢 I'm not entirely sure how to handle that. I don't wanna update customers, if we can just move them directly. I'd hope that we can just support a version in the 3 and 4 series 😢

Akash - In that case you keep mapping for the [lowest Astro-supported version ](https://docs.astronomer.io/astro/runtime-version-lifecycle-policy#stable-releases)i.e. Astro Runtime 4.2.9 with Airflow 2.2.5. They make a choice and move the nearest version they wanna move to Astro if they don't get an exact match. Astro supported version are documented and it is public so they should decide on that

> 

> Step 3: Create Astro Deployment ... For Scheduler Size Info Dialog box do you want to remove (Small/Medium and Large) and rather keep the respective CPU and Memory for the AU’s mentioned

> I was intentionally using Small/Medium/Large to match GCC and MWAA's terminology, that I imagine they'll be familiar with. Thoughts?

Akash - For this one, i suggest mentioning CPU and Memory as well for the AU units mentioned and they can take a call to keep Scheduler AU to the nearest possible value and once the deployment is created and dags are deployed they can monitor the scheduler AU using deployment analytics and optimize the value.

Having said that MWAA small/medium/large mapping is incorrect with AU, I have posted the numbers in my above comment.
For GCC, [users define the Scheduler CPU and Memory in core and GB respectively](https://cloud.google.com/composer/docs/composer-2/scale-environments#workloads-configuration), i didn't find a reference for small/medium and large ( i can be mistaken), thus it would make sense to keep CPU and Memory for AU units

fritz-astronomer:
re: version - I am **really** hoping we can get a 3 series supported :( I can add a "this is unsupported" tag for 3.x in that table?

You are right @iamakashladdha - I did size down the small/medium/large. I genuinely think those are WAY oversized.
GCC _kinda_ has a small/medium/large which is pretty close to MWAA - but you're right that it's mostly a slider. You only see that size mapping at the beginning. Would it be 👍 if I add CPU, GPU and a note our recommendation is sized down intentionally??

image
image
image

astro/migration-partials/directory-setup.mdx
```diff
@@ -1,5 +1,19 @@
-1. Using the same name as the chosen environment, create a new directory for your project
-2. Initialize a new Astro Project in this directory with `astro dev init`
-3. Initialize a new git repository in this directory with `git init`, using the git CLI and/or your Source Control Tool’s UI
```
fritz-astronomer: No git?

iamakashladdha:
Thanks @fritz-astronomer I am good with the [comment](https://github.com/astronomer/docs/pull/1992#issuecomment-1473961233) mentioned above. Please proceed.

astro/migration-partials/deployments.mdx
```diff
@@ -82,26 +82,29 @@ Before you create your Deployment, copy the following information from your sour
...
 
-    | Environment size | AUs |
-    | ---------------- | --- |
-    | Small            | 5   |
-    | Medium           | 10  |
-    | Large            | 15  |
+    | Environment size | AUs | CPU / Memory Equivalents | Note                                       |
+    |------------------|-----|--------------------------|--------------------------------------------|
+    | Small            | 5   | .5vCPU, 1.88GiB          |                                            |
+    | Medium           | 10  | 1vCPU, 3.75GiB           | smaller than source Airflow recommendation |
```
jwitz, line 106: @fritz-astronomer Can we include a column that defines what we mean by "small/medium/large" here? Like how many DAGs is that per environment?

astro/migration-partials/prereqs.mdx
```diff
@@ -0,0 +1,20 @@
...
+- Optional. You have [configured your identity provider (IdP)](configure-idp.md).
+
+On your local machine, make sure you have the following:
+
+- An [Astro account](log-in-to-astro.md).
+- The [Astro CLI](cli/install-cli.md).
+
+On the cloud service from which you're migrating, ensure that you have:
+
+- A source Airflow environment on Airflow 2 or later.
```
jwitz, line 14:
```suggestion
- A source Airflow environment on Airflow 2.2 or later.
```
@fritz-astronomer I'm concerned that we don't provide any steps for how to migrate an environment that's on a lower version than this. Do we recommend completing all required upgrades in your source environment before migrating? If so, we should say that.

astro/migration-partials/deployments.mdx
```diff
@@ -0,0 +1,158 @@
...
+
+    - `<true-or-false>` / `<variable-name>` / `<variable-value>`: Set system-level environment variables for your Deployment and specify whether they should be secret. Repeat this configuration in the file for any additional variables you need to set.
+    - `<deployment-name>`: Enter the same name as your source Airflow environment.
+    - `<deployment-description>`: Optional. Enter a description for your Deployment.
+    - `<runtime-version>`: Select the Runtime version that's based on the Airflow version that you use in your source Airflow environment. See the following table to determine which version of Runtime to use. Where exact version matches are not available, the nearest Runtime version is provided with its supported Airflow version in parentheses.
+
+    | Airflow Version | Runtime Version       | Note                                                                    |
+    |-----------------|-----------------------|-------------------------------------------------------------------------|
+    | 2.0             | 3.0.4 (Airflow 2.1.1) | Unsupported. The minimum supported version is Runtime 4.x / Airflow 2.2 |
```
jwitz, line 97:
```suggestion
```
@fritz-astronomer If it's unsupported, why would we mention it?

astro/migration-partials/deployments.mdx
fritz-astronomer: @jwitz - Small/Medium/Large are lifted directly from MWAA and GCC - so customers should be familiar. With that - do you feel it's still helpful?

astro/migration-partials/prereqs.mdx
fritz-astronomer: Agreed :/ I'd really like us to support all 2.x (and I think right now we are just going to do the best we can with all 2.x, so I wouldn't mention 2.2 specifically here) - but we do mention above that they at least need to get to Airflow 2.x or use Professional Services or update themselves before starting.

astro/migration-partials/deployments.mdx
fritz-astronomer: We will be commiting to migrate customers from all 2.x in GCC/MWAA - which both have a 2.0.2 version. We'd put them in the closest available version, so I was trying to be as transparent as possible. I also want to try to get us to expand our supported versions :/

astro/migration-partials/deployments.mdx
jwitz: Oh good to know! I might add a note about that before the table starts. 

Still confused about "Smaller than source Airflow recommendation". How does that information help users make a decision related to their AU usage?

astro/migration-partials/prereqs.mdx
jwitz: Hm, based on that conversation it seems like we want them to upgrade on MWAA/ GCC before triggering the migration?

fritz-astronomer: If possible - they should upgrade from 1.x to 2.x in MWAA/GCC (though that is difficult, honestly). If not, they should use our professional services, and this guide doesn't apply

astro/migration-partials/deployments.mdx
fritz-astronomer: following up in slack

astro/migration-partials/deployments.mdx
jwitz: Oh I see - yeah I'll add an asterisk note about this. Because Runtime 3.0.4 is unsupported, is the plan to deal with whatever upgrade pains arise from the jump from 2.x to Runtime 3.0.4 during migration, then upgrade in place until they're on a supported version? Should all customers on these versions expect to have their environments upgraded, and if so should we give them heads up about provider package versioning etc?

astro/migration-partials/prereqs.mdx
```diff
@@ -0,0 +1,20 @@
...
+- Optional. You have [configured your identity provider (IdP)](configure-idp.md).
+
+On your local machine, make sure you have the following:
+
+- An [Astro account](log-in-to-astro.md).
+- The [Astro CLI](cli/install-cli.md).
+
+On the cloud service from which you're migrating, ensure that you have:
+
+- A source Airflow environment on Airflow 2.2 or later.
```
jwitz, line 14:
```suggestion
- A source Airflow environment on Airflow 2 or later.
```

astro/migration-partials/deployments.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+   - **Name**: Enter the same name as your source Airflow environment.
+   - **Astro Runtime**: Select the Runtime version that's based on the Airflow version that you use in your source Airflow environment. See the following table to determine which version of Runtime to use. Where exact version matches are not available, the nearest Runtime version is provided with its supported Airflow version in parentheses.
+
+   | Airflow Version | Runtime Version       |
+   | --------------- | --------------------- |
+   | 2.0             | 3.0.4 (Airflow 2.1.1) |
+   | 2.2             | 4.2.9 (Airflow 2.2.5) |
+   | 2.4             | 6.3.0 (Airflow 2.4.3) |
+
```
jwitz, line 35:
```suggestion
   | Airflow Version | Runtime Version       |
   | --------------- | --------------------- |
   | 2.0             | 3.0.4 (Airflow 2.1.1)* |
   | 2.2             | 4.2.9 (Airflow 2.2.5) |
   | 2.4             | 6.3.0 (Airflow 2.4.3) |
   
   >*The earliest supported Airflow version on Astro Runtime is 2.1.1. There are no known risks for upgrading directly from Airflow 2.0 to Airflow 2.1.1 during migration. 
```

fritz-astronomer:
I did my pass (as best I could at the end of the day on a friday 😵‍💫 ) and ready to go 👍

jwitz (approved):

fritz-astronomer:
lgtm @jwitz 👍 
(I can't approve my own PR 😂 lame)

PR #1996, opened 2023-03-09
Add notes about different workspace layout strategies

fritz-astronomer:
we previously only alluded to "per env", but honestly I don't see people do that much.

Added:
- single workspace
- per env
- per team
- per application


jwitz (approved):
This is great @fritz-astronomer ! I only have some minor wording adjustments. Maybe we could elaborate on which strategies work for given team sizes/ contexts? But even if we don't include that info, this PR is good to merge

PR #2004, opened 2023-03-14
Fix for import code in Learn / Azure Container Instances

fletchjeff:
The import is currently `airflow.providers.microsoft.azure.operators.azure_container_instances` but should be `airflow.providers.microsoft.azure.operators.container_instances`


TJaniF (approved):
LGTM! Thank you for catching and fixing this :)

PR #2008, opened 2023-03-14
Product Email Updates Sign Up Form

katiebrady-astro:
This PR creates a new component, `newsletterForm`, that users can fill out to signup for Astro product updates.


jwitz:
@katiebrady-astro I love that this is configurable in Docusaurus :D 

Forgot if this was just to test one place over the other, but I think it would be good to only have one of these instances per page. Personally I prefer it in the footer of the doc itself rather than in the footer of the site
Screenshot 2023-04-27 at 10 02 57 AM

katiebrady-astro:
> @katiebrady-astro I love that this is configurable in Docusaurus :D
> 
> Forgot if this was just to test one place over the other, but I think it would be good to only have one of these instances per page. Personally I prefer it in the footer of the doc itself rather than in the footer of the site Screenshot 2023-04-27 at 10 02 57 AM

Oh whoops haha, yup meant to only have one of these active. Removed from the site footer and kept in the DocItem footer! 🙌

PR #2022, opened 2023-03-17
CSS Bug fix for Sidebar Nav and Bottom Links

katiebrady-astro:
This PR fixes an issue where the bottom links in the sidebar navigation were moving with the height of the page based on page content + window height. The bottom nav is now sticky to the bottom with an absolute position. The main sidebar itself is also always at least `100vh` to prevent content jumping around as much.

Previous:
image
image (1)

Fix with this PR:

Screenshot 2023-03-17 at 5 13 54 PM
Screenshot 2023-03-17 at 5 14 07 PM


docusaurus.config.js
```diff
@@ -102,6 +102,10 @@ module.exports = {
         }
       }
     },
+    newsletterForm: {
+      title: "Sign Up for Monthly Updates",
+      buttonText: "Submit"
+    },
```
paolaperaza, line 108: Is this new sign up form included in this PR, @katiebrady-astro ? cc: @juliana-astro

src/theme/DocSidebar/Desktop/Content/styles.module.css
```diff
@@ -19,9 +19,11 @@
 .menu__listBottom {
   font-family: "Strawford", "Open Sans", sans-serif;
   left: 1rem;
-  position: relative;
-  bottom: 2rem;
-  padding: 0;
+  position: absolute;
+  bottom: 0;
+  background: white;
```
jwitz, line 24: FYI I hotfixed this line out, was there another reason we added the fixed background color that we need to now work around? It seems to work fine without this

Issue #2010, opened 2023-03-15
Runtime upgrade docs should have considerations for provider version management


jwitz:
See https://astronomer.slack.com/archives/C015V2JFKT5/p1678829581812409

Affected doc:  https://docs.astronomer.io/astro/upgrade-runtime

PR #2043, opened 2023-03-22
Add Runtime upgrade step for provider package management

jwitz:
Resolves https://github.com/astronomer/docs/issues/2010

### DIFF
--- a/astro/upgrade-runtime.md
+++ b/astro/upgrade-runtime.md
@@ -61,7 +61,9 @@ If you're upgrading a local Airflow environment, you don't need an Astro Deploym
 
 ## (Optional) Pin provider package versions
 
-Major Astro Runtime upgrades can include major upgrades to built-in provider packages. For the most stable upgrade path, Astronomer recommends pinning all provider package versions from your current Runtime version before upgrading. To check the version of all provider packages installed in your Runtime version, run:
+Major Astro Runtime upgrades can include major upgrades to built-in provider packages. These package upgrades can sometimes include breaking changes for your DAGs. See the [Apache Airflow documentation](https://airflow.apache.org/docs/apache-airflow-providers/packages-ref.html) for a list of all available provider packages and their release notes. 
+
+For the most stable upgrade path, Astronomer recommends pinning all provider package versions from your current Runtime version before upgrading. To check the version of all provider packages installed in your Runtime version, run:
 
 ```sh
 docker run --rm quay.io/astronomer/astro-runtime:<current-runtime-version> pip freeze | grep apache-airflow-providers
--- a/astro/upgrade-runtime.md
+++ b/astro/upgrade-runtime.md
@@ -70,6 +70,22 @@ If you're upgrading a local Airflow environment, you don't need an Astro Deploym
 
     You must always specify the major, minor, and patch version of any given Astro Runtime version.
 
+## (Optional) Upgrade or pin provider package versions
+
+Minor Astro Runtime upgrades can include major provider package upgrades. To check the version of any provider package installed in the Runtime version you're upgrading to, run:
+
+```sh
+docker run --rm quay.io/astronomer/astro-runtime:<new-runtime-version> pip freeze | grep apache-airflow-providers
+```
+
+Run the same command for your existing Runtime version and compare provider packages between versions. If you see a package being upgraded that you're not yet ready to upgrade, pin the package version from your previous Runtime version in your Astro project `requirements.txt` file. For example, Runtime 7.4.1 has uses version 4.0.0 of `apache-airflow-providers-databricks`. To keep this version of the Databricks provider package when you upgrade to a later version of Runtime, you add the following line to your `requirements.txt` file:
+
+```text
+apache-airflow-providers-databricks==4.0.0
+```
+
+Similarly, review any currently pinned provider package versions in your `requirements.txt` file and determine whether you want to upgrade them alongside your Runtime upgrade.
+
 ## Test Astro Runtime locally
 
 Astronomer recommends testing new versions of Astro Runtime locally before upgrading a Deployment on Astro.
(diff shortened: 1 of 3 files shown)


astro/upgrade-runtime.md
```diff
@@ -81,6 +97,8 @@ Astronomer recommends testing new versions of Astro Runtime locally before upgra
 
     ![Runtime Version banner - Local](/img/docs/image-tag-airflow-ui-local.png)
 
+5. (Optional) Run DAGs locally to ensure that all of your code works as expected. If you encounter errors after your upgrade, it's possible that the Astro Runtime upgrade included a breaking provider package change. If you encounter one of these breaking changes, follow the steps in [Upgrade or pin provider package versions](#optional-upgrade-or-pin-provider-package-versions) to check your provider package versions and, if required, pin the provider package version from your previous Runtime version in your `requirements.txt` file.
```
jwitz, line 100:
```suggestion
5. (Optional) Run DAGs locally to ensure that all of your code works as expected. If you encounter errors after your upgrade, it's possible that your new Astro Runtime version includes a breaking provider package change. If you encounter one of these breaking changes, follow the steps in [Upgrade or pin provider package versions](#optional-upgrade-or-pin-provider-package-versions) to check your provider package versions and, if required, pin the provider package version from your previous Runtime version in your `requirements.txt` file.
```

astro/upgrade-runtime.md
```diff
@@ -70,6 +70,22 @@ If you're upgrading a local Airflow environment, you don't need an Astro Deploym
 
     You must always specify the major, minor, and patch version of any given Astro Runtime version.
 
+## (Optional) Upgrade or pin provider package versions
+
+Minor Astro Runtime upgrades can include major provider package upgrades. To check the version of any provider package installed in the Runtime version you're upgrading to, run:
```
jedcunningham, line 75: We now only do major provider upgrades in minor release when there are security fixes.

astro/upgrade-runtime.md
```diff
@@ -70,6 +70,22 @@ If you're upgrading a local Airflow environment, you don't need an Astro Deploym
...
+
+```sh
+docker run --rm quay.io/astronomer/astro-runtime:<new-runtime-version> pip freeze | grep apache-airflow-providers
+```
+
+Run the same command for your existing Runtime version and compare provider packages between versions. If you see a package being upgraded that you're not yet ready to upgrade, pin the package version from your previous Runtime version in your Astro project `requirements.txt` file. For example, Runtime 7.4.1 has uses version 4.0.0 of `apache-airflow-providers-databricks`. To keep this version of the Databricks provider package when you upgrade to a later version of Runtime, you add the following line to your `requirements.txt` file:
```
lzdanski, line 81:
```suggestion
Run the same command for your existing Runtime version and compare provider packages between versions. If you see a package being upgraded that you're not yet ready to upgrade, pin the package version from your previous Runtime version in your Astro project `requirements.txt` file. For example, Runtime 7.4.1 uses version 4.0.0 of `apache-airflow-providers-databricks`. To keep this version of the Databricks provider package when you upgrade to a later version of Runtime, you add the following line to your `requirements.txt` file:
```

lzdanski (approved):
Typo - otherwise, looks good!

astro/upgrade-runtime.md
```diff
@@ -70,6 +70,22 @@ If you're upgrading a local Airflow environment, you don't need an Astro Deploym
 
     You must always specify the major, minor, and patch version of any given Astro Runtime version.
 
+## (Optional) Upgrade or pin provider package versions
```
kentdanas, line 73: Not sure if this is reasonable or not (since it might exist in multiple places), but I'm wondering if it would make sense somewhere in here to link to release notes pages for the providers so users have a place to go to understand what the changes are. If our guidance is just "if it breaks, go back to the one that was working" then it's fine as is, but I think if it were me I would want to understand what had changed so I could figure out a path to upgrading (assuming we don't want people to get stuck in one version forever). If there's a straight forward way to make that link here I think it would be helpful, but also understand if that's not really possible.

astro/upgrade-runtime.md
```diff
@@ -70,6 +70,22 @@ If you're upgrading a local Airflow environment, you don't need an Astro Deploym
...
+
+Minor Astro Runtime upgrades can include major provider package upgrades. To check the version of any provider package installed in the Runtime version you're upgrading to, run:
+
+```sh
+docker run --rm quay.io/astronomer/astro-runtime:<new-runtime-version> pip freeze | grep apache-airflow-providers
```
collinmcnulty, line 78: I think this is not a good way to get the providers out of a runtime. If we're going to tell people to do this we should just publish the list.

jwitz: What's not good about it? I'm happy to publish a list, but we don't list provider upgrades in the Runtime changelog and don't have a way of showing them natively because the repo is private. We'd need a tool to extract the image versions and publish them somewhere (maybe a job for the Astro CLI?)

collinmcnulty: I don't think this is an issue we can resolve easily with just a docs change. I think we need to have a more in-depth discussion about it holistically.

astro/upgrade-runtime.md
```diff
@@ -59,6 +59,20 @@ If you're upgrading a local Airflow environment, you don't need an Astro Deploym
 
 :::
 
+## (Optional) Pin provider package versions
+
+Major Astro Runtime upgrades can include major upgrades to built-in provider packages. For the most stable upgrade path, Astronomer recommends pinning all provider package versions from your current Runtime version before upgrading. To check the version of all provider packages installed in your Runtime version, run:
```
jwitz, line 64:
```suggestion
Major Astro Runtime upgrades can include major upgrades to built-in provider packages. These package upgrades can sometimes include breaking changes for your DAGs. See the [Apache Airflow documentation](https://airflow.apache.org/docs/apache-airflow-providers/packages-ref.html) for a list of all available provider packages and their release notes. 

For the most stable upgrade path, Astronomer recommends pinning all provider package versions from your current Runtime version before upgrading. To check the version of all provider packages installed in your Runtime version, run:
```

PR #2044, opened 2023-03-22
Add pubDate to RSS channels and items

katiebrady-astro:
This PR for our RSS feeds adds `<pubDate>` attributes to the feed channels equal to last build time as well as to feed items equal to their publish date. Also adds `<guid>` to each feed item. This should help aggregators stay up-to-date. 🙂

### DIFF
--- a/createRSSFeeds.mjs
+++ b/createRSSFeeds.mjs
@@ -12,6 +12,7 @@ let astroCLI = await fs.readFile('./astro/cli/release-notes.md', 'utf8');
 let software = await fs.readFile('./software/release-notes.md', 'utf8');
 let softwareRuntime = await fs.readFile('./astro/runtime-release-notes.md', 'utf8');
 
+let currentDate = new Date();
 
 // function to get all elements between two elements
 function nextUntil(elem, selector, filter) {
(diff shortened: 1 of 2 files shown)


nandlalyadav57 (approved):

jwitz:
@katiebrady-astro If we make an edit to a release notes page after we publish, what effect would that have on `pubDate`?

katiebrady-astro:
> @katiebrady-astro If we make an edit to a release notes page after we publish, what effect would that have on `pubDate`?

Hey @jwitz, the `pubDate` for the feed itself will update everytime the script runs to build the RSS XML files. But the `pubDate` for the individual release notes will stay static with whatever they're listed as for that release (e.g. `Release date: March 23, 2023` or for Astro Release Notes their `<h2>` title since that's also the pubDate). Most aggregators, including Slack's, will periodically look at the last published item's `pubDate` and update if that's changed/new.

Hope that makes sense/answers the question! 🙂

jwitz (approved):
For sure, as long as folks don't get notifications whenever we fix an earlier dated release note, then this LGTM!

PR #2048, opened 2023-03-23
Update what-is-a-sensor.md

fritz-astronomer:
customer pointed out import path change for `SqlSensor`

### DIFF
--- a/learn/what-is-a-sensor.md
+++ b/learn/what-is-a-sensor.md
@@ -67,6 +67,12 @@ To review the available Airflow sensors, go to the [Astronomer Registry](https:/
 
 The following example DAG shows how you might use the `SqlSensor` sensor: 
 
+:::info
+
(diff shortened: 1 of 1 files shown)


kentdanas (changes requested):
@fritz-astronomer thanks for pointing this out. Can we just update the example DAG instead of leaving a note? We always want the docs to reflect the most up to date Airflow version, so no need to keep the older import path shown.

fritz-astronomer:
> @fritz-astronomer thanks for pointing this out. Can we just update the example DAG instead of leaving a note? We always want the docs to reflect the most up to date Airflow version, so no need to keep the older import path shown.

ah, that makes sense. I was going for safety - assuming people reading might not be on the newest, but that makes sense that we always want to target the newest 🤔 I'll update when I get a chance

TJaniF:
Had a minute and am procrastinating something else so I did the import change in the code. Thanks @fritz-astronomer for catching this!

Issue #2013, opened 2023-03-15
New Astro Quick Start Guide


lzdanski:
**Request type:**

- [x] **New tutorial:**: Add the `docs-content` and `area/learn` labels to this issue.
- [ ] **New Airflow concept:**: Add the `docs-content` and `area/learn` label to this issue.
- [ ] **Addition to an existing guide**: Add the `area-learn` and `enhancement` labels to this issue.

**Use case or feature that should be covered:**
- Update content that allows users to finish their first `astro deploy` step. 
- New topic likely to replace [Create an Astro project](https://docs.astronomer.io/astro/create-project)
- See notes in [Getting started friction log](https://www.notion.so/astronomerio/Friction-Logs-Astro-f54df0627ead41938dc4ee27ed1b0a02?pvs=4)

**Who the guide is intended for:**
e.g. users who are new to Airflow, users who are migrating from a legacy scheduler, etc.
- Users who are new to Astro and potentially also new to airflow

**Are you the right person to review a draft of the guide for technical accuracy? If not, who is?**

PR #2101, opened 2023-04-03
Airflow quickstart Learn page

kentdanas:
Drafted up a page for Learn that will link to our new Airflow quickstart repo: https://github.com/astronomer/airflow-quickstart

I intentionally did not include a ton of details on this page because the quickstart Readme is quite detailed and I don't want to maintain the information in more than one place as much as possible. My main goal with this page is to bring more attention to the quickstart as a possible learning path, link to related resources, and to help people distinguish whether they want this or another resource (e.g. a more instructional tutorial). 

Happy for any feedback on the best way to do this. And where to put it! Under 'get started with Airflow' seemed the most logical place to me to have a sidebar label, but I think it would also be ideal to have this on the Learn homepage somewhere as its own box.


learn/airflow-quickstart.md
```diff
@@ -0,0 +1,29 @@
...
+title: "Airflow Quickstart"
+sidebar_label: "Quickstart"
+id: airflow-quickstart
+---
+
+A great way to explore Airflow features and use cases is using Astronomer's [Airflow Quickstart](https://github.com/astronomer/airflow-quickstart).
+
+This quickstart uses open source tools Airflow, [DuckDB](https://duckdb.org/), [Streamlit](https://streamlit.io/), and the [Astro Python SDK](https://astro-sdk-python.readthedocs.io/en/stable/index.html) to implement an ELT pipeline that creates an interactive dashboard showing temperature changes over time for your city.
+
+![Interactive dashboard](/img/guides/quickstart_streamlit_result_1.png)
```
TJaniF, line 11: Small nitpick: the sentence references the latter half of the streamlit app that shows city data and the screenshot is from the first half showing the global climate data. Not sure if people might find that confusing. 

Maybe "interactive dashboard showing global and local temperature changes over time."? 

Because I agree that the picture shown should be from the start of the dashboard, it would look odd showing the scrolled down part 😄

TJaniF (approved):
Looks great! I only had one small nitpick. :)

For the location I think we discussed this a bit when talking about the Quickstart and I liked the idea of having a top level sidebar item called "Get started" (maybe below "Overview")
and then 3 pages in there: Quickstart, Get started tutorial and Academy.

kentdanas:
> 

Ah yes, I remember that now. I do like the idea of pulling it out into a new section. Would likely be more discoverable than having it nested under tutorials. @paolaperaza let us know what you think.

learn/airflow-quickstart.md
```diff
@@ -0,0 +1,29 @@
+---
+title: "Airflow Quickstart"
+sidebar_label: "Quickstart"
+id: airflow-quickstart
+---
+
+A great way to explore Airflow features and use cases is using Astronomer's [Airflow Quickstart](https://github.com/astronomer/airflow-quickstart).
+
+This quickstart uses open source tools Airflow, [DuckDB](https://duckdb.org/), [Streamlit](https://streamlit.io/), and the [Astro Python SDK](https://astro-sdk-python.readthedocs.io/en/stable/index.html) to implement an ELT pipeline that creates an interactive dashboard showing global and local temperature changes over time.
```
lzdanski, line 9:
```suggestion
This quickstart uses open source tools including Airflow, [DuckDB](https://duckdb.org/), [Streamlit](https://streamlit.io/), and the [Astro Python SDK](https://astro-sdk-python.readthedocs.io/en/stable/index.html) to implement an ELT pipeline that creates an interactive dashboard showing global and local temperature changes over time.
```

lzdanski: Or could be - `...uses open source tools, Airflow, DuckDB, and the Astro Python SDK,...` but it needs something there grammatically.

lzdanski (changes requested):
This looks fantastic, @kentdanas! I had one small sentence nitpick but I'm definitely going to be using this.

kentdanas:
@TJaniF @lzdanski thanks for your reviews! I updated the Learn sidebar to create a new "get started" category that contains our intro tutorials and this quickstart. Couple questions for you in:

1. What do we think of this layout? Overall I like putting these resources right at the top rather than having them nested further down and I think this will make the learning journey more clear. This is also what we had discussed doing previously, but want to make sure we're still on the same page.
2. Any better suggestions for the label on the tutorial subsection? (I don't love having "get started" twice). Or maybe we don't need that anymore since there are only a few things in this section?
3. [not a question, but a comment] I know we had discussed linking to academy here as well, but I'm going to hold on that for now because I'm not totally sure what Paola envisions there (unless she told you @lzdanski)

lzdanski:
> 1. What do we think of this layout? Overall I like putting these resources right at the top rather than having them nested further down and I think this will make the learning journey more clear. This is also what we had discussed doing previously, but want to make sure we're still on the same page.

I really like this layout because it sets up a consistent navigation structure between both the Astro tab and the Learn tab once these two PR's are merged. 

1. https://github.com/astronomer/docs/issues/2013 > has _some_ redundancy with the "Create and run your first DAG", but we were careful to keep the new Astro Quickstart really focused on Astro vs learning about Airflow.
2. https://github.com/astronomer/docs/issues/2111 > Moves _Migrate to Astro_ into the Quickstart section for Astro and uses a 2 level nested sidebar navigation, like your Learn: Get Started changes in this PR create.

> 2. Any better suggestions for the label on the tutorial subsection? (I don't love having "get started" twice). Or maybe we don't need that anymore since there are only a few things in this section?

I think something like **Airflow Tutorials** would be both specific and also clarifies that these aren't exactly quickstarts, they're a longer tour through the function of Airflow and that Airflow/Astro relationship. 

> [not a question, but a comment] I know we had discussed linking to academy here as well, but I'm going to hold on that for now because I'm not totally sure what Paola envisions there (unless she told you @lzdanski)

I'm not sure what she has in mind for Academy, but one of my big pieces of feedback and observations in my [Getting started friction log](https://www.notion.so/astronomerio/Friction-Logs-Astro-f54df0627ead41938dc4ee27ed1b0a02?pvs=4) was that it's really difficult to navigate to Academy by clicking around in the Docs + on the main Astro sites. Paola has asked me to dig into places in the docs it might make sense to merge/link content between the two, so this might be TBD?

lzdanski (approved):

kentdanas:
> I think something like **Airflow Tutorials** would be both specific and also clarifies that these aren't exactly quickstarts, they're a longer tour through the function of Airflow and that Airflow/Astro relationship.

@lzdanski hmm my only concern here is we already have **Airflow tutorials** as a high level section on the nav bar, so I don't think we want to duplicate that. Maybe **Introductory tutorials**?

lzdanski:
I think **Introductory tutorials** but also - do we have just **Tutorials** anywhere else?

kentdanas:
> I think **Introductory tutorials** but also - do we have just **Tutorials** anywhere else?

I don't think so...Since it's a subsection maybe we don't need to distinguish it any more from the general "Airflow tutorials" section. I'd say let's go with "Tutorials" for now, and we can always update it later.

sidebarsLearn.js
```diff
@@ -1,6 +1,30 @@
...
+      link: {
+        type: 'generated-index',
+        title: 'Get started',
+        description: 'Get started with Airflow.',
+      },
+      items: [
+        'airflow-quickstart',
+        {
+          type: 'category',
+          label: 'Get started tutorials',
```
kentdanas, line 16:
```suggestion
          label: 'Tutorials',
```

TJaniF:
@kentdanas I love the layout! (very curious to see if we'll get more traffic on the get started tutorial too after this is merged simply because it makes is so much easier to find). 

For the Get started tutorial tab. Maybe something like "Step-by-step" or "Your first DAG"? The first one is a bit out there and maybe promises too much, the second one would assume people know vaguely what a DAG is, so maybe not ideal.
"First steps"? Just some ideas, I think Tutorial works well too under the "Get started" section :)

lzdanski:
For any disambiguation things - We're calling the new Astro Quickstart "Run your first DAG " as the title because it covers creating the sample DAGs with `astro dev init` and running them on Astro.

PR #2112, opened 2023-04-05
Move migration docs to getting started 

lzdanski:
Move docs from `Admin` section to `Getting Started`

https://github.com/astronomer/docs/issues/2111


jwitz:
Catching up on this.. Is there any concern with having a guide that requires working with an Astronomer Data Engineer in our getting started material? At the moment our starting docs assume minimal friction through a trial and follow a logical process from starting a trial to running your first DAG, but the migration guides are complex processes that repeat a lot of the steps from other getting started docs. I'm worried that by catering to two different user personas, we end up making it confusing for both.

paolaperaza (approved):
Assuming this is building correctly, this LGTM @lzdanski ! Sent message in Slack with my thoughts based on conversation with Elan 😊

PR #2117, opened 2023-04-07
Add repeated mapping to the dynamic task guide

TJaniF:
Thanks @astronautyates for the suggestion! 

This PR adds a short section on what happens when a task is dynamically mapped over the output of another dynamically mapped task. :)

### DIFF
--- a/learn/dynamic-tasks.md
+++ b/learn/dynamic-tasks.md
@@ -442,6 +442,16 @@ It is possible to dynamically map an Airflow task over the output of another dyn
 
 The following example shows three dynamically mapped tasks.
 
+<Tabs
+    defaultValue="taskflow"
+    groupId="repeated-mapping"
+    values={[
+        {label: 'TaskFlow API', value: 'taskflow'},
+        {label: 'Traditional syntax', value: 'traditional'},
+    ]}>
+
+<TabItem value="taskflow">
+
 ```python
 @task
 def multiply_by_2(num):
@@ -463,6 +473,38 @@ multiply_by_100.expand(
     )
 )
 ```
+</TabItem>
+<TabItem value="taskflow">
+
+```python
+def multiply_by_2_func(num):
+    return [num * 2]
+
+def add_10_func(num):
+    return [num + 10]
+
+def multiply_by_100_func(num):
+    return num * 100
+
+multiply_by_2 = PythonOperator.partial(
+    task_id="multiply_by_2",
+    python_callable=multiply_by_2_func
+).expand(op_args=[[1], [2], [3]])
+
+add_10 = PythonOperator.partial(
+    task_id="add_10",
+    python_callable=add_10_func
+).expand(op_args=multiply_by_2.output)
+
+multiply_by_100 = PythonOperator.partial(
+    task_id="multiply_by_100",
+    python_callable=multiply_by_100_func
+).expand(op_args=add_10.output)
+
+multiply_by_2 >> add_10 >> multiply_by_100
+```
+</TabItem>
+</Tabs>
 
 In the example above the `multiply_by_2` task is dynamically mapped over a list of 3 elements (`[1, 2, 3]`). The task has 3 dynamically mapped task instances returning the following values:
 
@@ -476,7 +518,7 @@ The `add_10` task is dynamically mapped over the output of the `multiply_by_2` t
 - Map index 1: `14` (4+10)
 - Map index 2: `16` (6+10)
 
-The `multiply_by_10` task is dynamically mapped over the output of the `add_10` task. It also has 3 dynamically mapped task instances returning the values:
+The `multiply_by_100` task is dynamically mapped over the output of the `add_10` task. It also has 3 dynamically mapped task instances returning the values:
 
 - Map index 0: `1200` (12*100)
 - Map index 1: `1400` (14*100)
--- a/learn/dynamic-tasks.md
+++ b/learn/dynamic-tasks.md
@@ -436,6 +436,54 @@ The add_nums task will have three mapped instances with the following results:
 - Map index 1: `1202` (2+1000+200)
 - Map index 2: `2300` (1000+1000+300)
 
+## Repeated mapping
+
+It is possible to dynamically map an Airflow task over the output of another dynamically mapped task. In this situation the each mapped task will create one mapped task instance for every mapped task instance of the first mapped task.
+
+The following example shows three dynamically mapped tasks.
+
+```python
+@task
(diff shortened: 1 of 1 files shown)


learn/dynamic-tasks.md
```diff
@@ -436,6 +436,96 @@ The add_nums task will have three mapped instances with the following results:
...
+</TabItem>
+</Tabs>
+
+In the example above the `multiply_by_2` task is dynamically mapped over a list of 3 elements (`[1, 2, 3]`). The task has 3 dynamically mapped task instances returning the following values:
+
+- Map index 0: `2` (1*2)
+- Map index 1: `4` (2*2)
+- Map index 2: `6` (3*2)
+
+The `add_10` task is dynamically mapped over the output of the `multiply_by_2` task. It has 3 dynamically mapped task instances (one for each mapped instance of the previous task) returning the values:
```
kentdanas, line 515:
```suggestion
The `add_10` task is dynamically mapped over the output of the `multiply_by_2` task. It has 3 mapped task instances (one for each mapped instance of the previous task) returning the values:
```

learn/dynamic-tasks.md
```diff
@@ -436,6 +436,96 @@ The add_nums task will have three mapped instances with the following results:
...
+
+- Map index 0: `12` (2+10)
+- Map index 1: `14` (4+10)
+- Map index 2: `16` (6+10)
+
+The `multiply_by_100` task is dynamically mapped over the output of the `add_10` task. It also has 3 dynamically mapped task instances returning the values:
```
kentdanas, line 521:
```suggestion
The `multiply_by_100` task is dynamically mapped over the output of the `add_10` task. It also has 3 mapped task instances returning the values:
```

kentdanas (approved):
Looks good, just a few minor comments! If we can let's get this merged before the workshop on Weds.

TJaniF (commented):
Thank you @kentdanas !

@jwitz ready for you, it would be great if we can get this in before Wednesday :)

learn/dynamic-tasks.md
```diff
@@ -436,6 +436,96 @@ The add_nums task will have three mapped instances with the following results:
 - Map index 1: `1202` (2+1000+200)
 - Map index 2: `2300` (1000+1000+300)
 
+## Repeated mapping
+
+It is possible to dynamically map an Airflow task over the output of another dynamically mapped task. This results in one mapped task instance for every mapped task instance of the upstream task.
```
jwitz, line 441:
```suggestion
You can dynamically map an Airflow task over the output of another dynamically mapped task. This results in one mapped task instance for every mapped task instance of the upstream task.
```

learn/dynamic-tasks.md
```diff
@@ -436,6 +436,96 @@ The add_nums task will have three mapped instances with the following results:
...
+</TabItem>
+</Tabs>
+
+In the example above, the `multiply_by_2` task is dynamically mapped over a list of 3 elements (`[1, 2, 3]`). The task has 3 mapped task instances returning the following values:
+
+- Map index 0: `2` (1*2)
+- Map index 1: `4` (2*2)
+- Map index 2: `6` (3*2)
+
+The `add_10` task is dynamically mapped over the output of the `multiply_by_2` task. It has 3 mapped task instances (one for each mapped instance of the previous task) returning the values:
```
jwitz, line 515:
```suggestion
The `add_10` task is dynamically mapped over the output of the `multiply_by_2` task. It has 3 mapped task instances (one for each mapped instance of the previous task) which contain the following values:
```

learn/dynamic-tasks.md
```diff
@@ -436,6 +436,96 @@ The add_nums task will have three mapped instances with the following results:
...
+    task_id="multiply_by_100",
+    python_callable=multiply_by_100_func
+).expand(op_args=add_10.output)
+
+multiply_by_2 >> add_10 >> multiply_by_100
+```
+</TabItem>
+</Tabs>
+
+In the example above, the `multiply_by_2` task is dynamically mapped over a list of 3 elements (`[1, 2, 3]`). The task has 3 mapped task instances returning the following values:
```
jwitz, line 509:
```suggestion
In the example above, the `multiply_by_2` task is dynamically mapped over a list of three elements (`[1, 2, 3]`). The task has three mapped task instances containing the following values:
```

learn/dynamic-tasks.md
```diff
@@ -436,6 +436,96 @@ The add_nums task will have three mapped instances with the following results:
 - Map index 1: `1202` (2+1000+200)
 - Map index 2: `2300` (1000+1000+300)
 
+## Repeated mapping
```
jwitz, line 439: For my own understanding: is this the same thing as recursive mapping?

learn/dynamic-tasks.md
jwitz: Feel free to change the word back if my understanding is wrong.

jwitz (approved):
LGTM! Just a few minor updates and some questions I was curious about

learn/dynamic-tasks.md
TJaniF: Hmm, I think with dynamic task mapping what I'd call recursive mapping is if each mapped task instance would be the base of an entire set of more mapped tasks. so 3 instances, next task maps 3 = 9 and so forth and we would be using the same task, which keeps using its own mapped version as an input until it hits a base case. Which I am pretty sure is currently not possible but would be very interesting (and potentially system crashing 😅 ).

But yeah that is a great question, I think if that exponential pattern becomes possible recursive mapping would be a possible term for it 🤔 . My programming terminology is a little too blurry here to be sure if that would be ok, since usually recursion refers to a function calling itself, so maybe it depends on how the code is resolved under the hood, whether we could truly call that recursion.

Btw I stole the term repeated mapping from the Airflow docs :)

learn/dynamic-tasks.md
TJaniF: That is a good question. I would say the value is returned by the function that the task uses, which pushes it to XCom with the return_value key, which is the can then be retrieved from the mapped task instance by calling xcom_pull, so I think saying that it is contained by the task instance works. :) 

(looking at the source code XCom are evaluated in a lazy way that I don't fully understand yet, so this might not be entirely true, and maybe some of the values are only created/returned if they are actually accessed, but I think  contain works for now :) )

TJaniF (commented):
Thank you @jwitz ! Those questions are great and my tldr answer is that I really need to learn how to less blurrily talk about some Python code structures 😅

PR #2120, opened 2023-04-10
Add new Env Var

lzdanski:
* https://github.com/astronomer/docs/issues/2119
* Add `INSTANCE_TYPE` to env var doc. 
Screenshot 2023-04-10 at 1 14 57 PM


lzdanski:
@rbankston - Paola and Jake are OOO - can you take a quick look to make sure that this definition is correct? I added some more detail but I might have misinterpreted your original draft.

jwitz (approved):

PR #2136, opened 2023-04-14
Add traditional operator override example

fritz-astronomer:
For KubeExec - we only have taskflow


jwitz (approved):
@fritz-astronomer Do we need to call the trad operator as an upstream/ downstream dependency at the end of the DAG? Otherwise LGTM!

PR #2137, opened 2023-04-14
Update KubeExec Limits

fritz-astronomer:
relates to https://github.com/astronomer/astro/pull/12286

### DIFF
--- a/astro/executors.md
+++ b/astro/executors.md
@@ -106,7 +106,7 @@ On Astro, you can configure Kubernetes executor in the following ways:
 - Customize individual tasks, including CPU and memory requests, using a `pod_override` configuration. 
 - Change the worker node type on which your Pods run.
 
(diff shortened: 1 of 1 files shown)


jwitz (approved):
LGTM

Why is it "Up to"? I thought the point of Pods is that they spin up with the same amount of resources and don't scale

fritz-astronomer:
Pods can scale and that's pretty normal and recommended in Kubernetes. They are created with resource `requests` and `limits` which are both optional to define. Limits - obviously, you can't scale past. In Astro, we happen to have the `request` and `limit` of memory be the same, which means we do always book exactly that much. My initial work in the PR linked was for a request/limit that would allow some room to grow, but we did just change it to match nebula where you do only ask for `384Mi` and have the same request as your limit so you can't grow past that

fritz-astronomer:
Moving this to a draft PR until the linked work is merged and released

PR #2151, opened 2023-04-18
Correct default value for poke_interval of sensor

pankajkoti:
The default poke_interval value is 60 seconds, however, the what-is-sensor.md wrongly mentions it as 30 seconds. This commit corrects the default value mentioned.


TJaniF (approved):
Thank you @pankajkoti, great catch! :)

pankajkoti:
It was not my catch, it was Marc's catch :)  Thanks, @TJaniF for the quick review and merge!

PR #2154, opened 2023-04-18
Software: Add required annotation for azure load-balancer

rob-1126:
This annotation is apparently required for all currently supported versions of Software on AKS  (e.g. > 1.22) but has never been included in our docs.

### DIFF
--- a/software/install-azure-standard.md
+++ b/software/install-azure-standard.md
@@ -324,6 +324,7 @@ nginx:
   ingressAnnotations:
     # required for azure load balancer post Kubernetes 1.22
     service.beta.kubernetes.io/azure-load-balancer-health-probe-request-path: "/healthz"
+
 astronomer:
   houston:
     config:
--- a/software/install-azure-standard.md
+++ b/software/install-azure-standard.md
@@ -321,8 +321,9 @@ nginx:
   # IP address the nginx ingress should bind to
   loadBalancerIP: ~
   # Dict of arbitrary annotations to add to the nginx ingress. For full configuration options, see https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-annotations/
-  ingressAnnotations: {}
-
+  ingressAnnotations:
+    # required for azure load balancer post Kubernetes 1.22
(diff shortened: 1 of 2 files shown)


software/install-azure-standard.md
```diff
@@ -321,7 +321,9 @@ nginx:
   # IP address the nginx ingress should bind to
   loadBalancerIP: ~
   # Dict of arbitrary annotations to add to the nginx ingress. For full configuration options, see https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-annotations/
-  ingressAnnotations: {}
+  ingressAnnotations:
+    # required for azure load balancer post Kubernetes 1.22
```
jwitz, line 325:
```suggestion
    # required for azure load balancer post Kubernetes 1.24
```
Is our upgrade note wrong here? https://docs.astronomer.io/software/upgrade-astronomer#kubernetes-124-on-azure

software_versioned_docs/version-0.30/install-azure-standard.md
```diff
@@ -311,8 +311,10 @@ nginx:
   # IP address the nginx ingress should bind to
   loadBalancerIP: ~
   # Dict of arbitrary annotations to add to the nginx ingress. For full configuration options, see https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-annotations/
-  ingressAnnotations: {}
-
+  ingressAnnotations:
+    # required for azure load balancer post Kubernetes 1.22
```
jwitz, line 315:
```suggestion
    # required only if using Kubernetes 1.24 or later
```

software/install-azure-standard.md
rob-1126: I think they are right. I was told post 1.22 (aka 1.23+) but google seems to say its 1.24+. Lets roll with 1.24+ and we can fix it if we hear different.

rob-1126:
@jwitz Lets roll in with 1.24+, we can bump it down to 1.23+ if we get confirmation. Let's get this to press :)

PR #2170, opened 2023-04-21
fix delete action for deployment previews

sunkickr:
delete deployment preview script is wrong. triggering on pull request when should be triggered when branch is deleted


lzdanski (approved):
LGTM

Issue #2171, opened 2023-04-21
Fix broken link in Astro Software docs


lzdanski:
**Affected Document:**
* Link for Azure is broken https://docs.astronomer.io/software/deploy-nfs#provision-an-nfs-volume
**Problem Type (Priority):**

- [ ] Error (Critical): This document has existing information that's incorrect and prevents readers from completing the documented setup (Add the `error/critical` label to this issue)
- [ ] Misleading (Important): This document could benefit from additional context, guidance, or clarity (Add the `error/FYI` label to this issue)
- [x] Typo (Low): Any issue related to spelling/ punctuation/ formatting (Add the `error/typo` label to this issue)

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**

Update URL to point to https://docs.astronomer.io/software/deploy-nfs#create-and-mount-an-azure-file-share

**Additional Notes:**

Conversation on slack: https://astronomer.slack.com/archives/C015V2JFKT5/p1682104219062349

PR #2172, opened 2023-04-21
Update link to match header

lzdanski:
Broken link in Software doc from mismatched rel link and header text.

https://github.com/astronomer/docs/issues/2171


jwitz (approved):
Thanks @lzdanski !

PR #2178, opened 2023-04-24
[Software] Add docs for cluster provisioning

jwitz:


software/cluster-resource-provisioning.md
```diff
@@ -0,0 +1,33 @@
...
+    astronomer:
+      houston:
+        config:
+          deployments:
+            # This is a multiplying factor as a percentage of the limits. Defaults to 1
+            overProvisioningFactorMem: 1
+            overProvisioningFactorCPU: 1
+    ```
+
+    For example, if you set `overProvisioningFactorMem: 1.2` and `overProvisioningFactorCPU: 0.5`, an AU will be equivalent to 0.12 CPU and 192.5 MB of memory.
```
danielhoherd, line 30: Is this correct? I thought overProvisioning was setting the requests to a percentage of the limits, and thus could not go above 1.0. CC @rbankston @rob-1126 @rishkarajgi

rbankston: I thought it could only be a floating point number between 0.001 and 1 since multiplication was done on the limits to get the new requests. I believe if you set a number above 1 it would produce an invalid yaml file since the requests would be above the limits,

jwitz:
```suggestion
    For example, if you set `overProvisioningFactorMem: 0.75` and `overProvisioningFactorCPU: 0.5`, an AU will be equivalent to 0.075 CPU and 192.5 MB of memory.
```

jwitz: @rbankston So you can only under-provision Deployments? Why is the key called `overProvisioning`? I might be missing something in these definitions.

software/cluster-resource-provisioning.md
```diff
@@ -0,0 +1,33 @@
...
+description: Configure your Astronomer cluster to over- or underprovision Deployment resources.
+---
+
+:::danger Development workloads only
+
+Changing cluster resource provisioning is only recommended for development clusters and can result in task failures. Astronomer can't provide support or troubleshoot issues related to  over- or underprovisioning Deployments through this configuration.
+
+:::
+
+By default, Deployments can request CPU and memory for Airflow components in terms Astronomer Units (AU). One AU is equivalent to 0.1 CPU and 385 MB of memory. You can only request whole numbers of AUs, which can sometimes result in you over- or underprovisioning your Deployments.
```
jwitz, line 14:
```suggestion
By default, Deployments can request CPU and memory for Airflow components in terms Astronomer Units (AU). One AU is equivalent to 0.1 CPU and 385 MB of memory. You can only request whole numbers of AUs, which can sometimes result in you overprovisioning your Deployments.
```

software/cluster-resource-provisioning.md
```diff
@@ -0,0 +1,33 @@
...
+
+To change this behavior, you can change the amount of CPU and memory that an AU represents, allowing you to more accurately provision resources based on the requirements for your Deployments. 
+
+1. Add the following configuration to your `config.yaml` file. Replace the values for `overProvisioningFactorMem` and `overProvisioningFactorCPU` with the factor by which you want to increase or decrease the amount of CPU and memory in an AU.
```
jwitz, line 18:
```suggestion
1. Add the following configuration to your `config.yaml` file. Replace the values for `overProvisioningFactorMem` and `overProvisioningFactorCPU` with the factor by which you want to increase or decrease the amount of CPU and memory that an AU represents.
```

software/cluster-resource-provisioning.md
jwitz: Ah I see what you mean though, in that by reducing requests, you are actually running **more** work on a node than before.

software/cluster-resource-provisioning.md
```diff
@@ -0,0 +1,33 @@
...
+
+:::danger Development workloads only
+
+Changing cluster resource provisioning is only recommended for development clusters and can result in task failures. Astronomer can't provide support or troubleshoot issues related to  over- or underprovisioning Deployments through this configuration.
+
+:::
+
+By default, Deployments can request CPU and memory for Airflow components in terms Astronomer Units (AU). One AU is equivalent to 0.1 CPU and 385 MB of memory. You can only request whole numbers of AUs, which can sometimes result in you over- or underprovisioning your Deployments.
+
+To change this behavior, you can change the amount of CPU and memory that an AU represents, allowing you to more accurately provision resources based on the requirements for your Deployments.
```
jwitz, line 16:
```suggestion
To change this behavior, you can change the amount of CPU and memory that an AU requests, allowing you to more efficiently provision resources based on the requirements for your Deployments. 
```

software/cluster-resource-provisioning.md
```diff
@@ -0,0 +1,33 @@
...
+    astronomer:
+      houston:
+        config:
+          deployments:
+            # This is a multiplying factor as a percentage of the limits. Defaults to 1
+            overProvisioningFactorMem: 1
+            overProvisioningFactorCPU: 1
+    ```
+
+    For example, if you set `overProvisioningFactorMem: 0.75` and `overProvisioningFactorCPU: 0.5`, an AU will be equivalent to 0.075 CPU and 192.5 MB of memory.
```
jwitz, line 30:
```suggestion
    For example, if you set `overProvisioningFactorMem: 0.75` and `overProvisioningFactorCPU: 0.5`, an AU will only request 0.075 CPU and 192.5 MB of memory on a node, allowing you to run more components on that node than before.
```

software/cluster-resource-provisioning.md
danielhoherd: I would reword this so the order of cpu and memory is the same in both parts:

```
For example, if you set `overProvisioningFactorCPU: 0.5` and `overProvisioningFactorMem: 0.75`, an AU will be equivalent to 0.075 CPU and 192.5 MB of memory.
```

software/cluster-resource-provisioning.md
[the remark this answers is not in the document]
danielhoherd: > factor by which you want to decrease the request for resources

I think this wording could be improved.

It's not decreasing by a factor, EG: if you set it to 0.9 it won't decrease by 0.9, it will decrease by 0.1. I'm not sure what the best way to describe that is though.

Also, I think rewording the last part to be more k8s standard would be good. Maybe "the factor that you want to set your resource requests to" ?

PR #2188, opened 2023-04-25
Update secrets-backend.md

fritz-astronomer:
Customer reports that the import path changed and the instructions were confusing

### DIFF
--- a/astro/secrets-backend.md
+++ b/astro/secrets-backend.md
@@ -455,7 +455,7 @@ Add the following environment variables to your `.env` file:
 AZURE_CLIENT_ID="<your-client-id>" # Found on App Registration page > 'Application (Client) ID'
 AZURE_TENANT_ID="<your-tenant-id>" # Found on App Registration page > 'Directory (tenant) ID'
 AZURE_CLIENT_SECRET="<your-client-secret>" # Found on App Registration Page > Certificates and Secrets > Client Secrets > 'Value'
-AIRFLOW__SECRETS__BACKEND=airflow.providers.microsoft.azure.secrets.azure_key_vault.AzureKeyVaultBackend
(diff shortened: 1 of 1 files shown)


astro/secrets-backend.md
```diff
@@ -465,21 +465,24 @@ By default, this setup requires that you prefix any secret names in Key Vault wi
```
jwitz, line 465: I think the note about Git is more appropriate here.

jwitz (changes requested):
@fritz-astronomer Is there a reason we're not telling users to set the environment variables as secret in Astro any more?

fritz-astronomer:
> @fritz-astronomer Is there a reason we're not telling users to set the environment variables as secret in Astro any more?

History, no good reason. Let's pick something and standardize 🤷

astro/secrets-backend.md
```diff
@@ -455,7 +455,7 @@ Add the following environment variables to your `.env` file:
 AZURE_CLIENT_ID="<your-client-id>" # Found on App Registration page > 'Application (Client) ID'
 AZURE_TENANT_ID="<your-tenant-id>" # Found on App Registration page > 'Directory (tenant) ID'
 AZURE_CLIENT_SECRET="<your-client-secret>" # Found on App Registration Page > Certificates and Secrets > Client Secrets > 'Value'
-AIRFLOW__SECRETS__BACKEND=airflow.providers.microsoft.azure.secrets.azure_key_vault.AzureKeyVaultBackend
+AIRFLOW__SECRETS__BACKEND=airflow.providers.microsoft.azure.secrets.key_vault.AzureKeyVaultBackend
```
fritz-astronomer, line 458: https://github.com/apache/airflow/blob/main/airflow/providers/microsoft/azure/secrets/key_vault.py#L37
Just proving to myself that this is the correct path, cuz I had originally just guessed 😅

jwitz (approved):

PR #2198, opened 2023-04-27
Fix elasticsearch config indentation

stuart23:
Indentation was wrong and would cause a problem if used in this format.

Indentation now matches https://astronomer.zendesk.com/knowledge/articles/4415725074835/en-us?brand_id=360001488713


stuart23:
Sorry, hold for 1 moment - its in the code twice

stuart23:
OK, now it's ready to go

kentdanas (approved):

PR #2195, opened 2023-04-27
[2.6] Rerun guide to include retries

TJaniF:
I tagged this PR with 2.6 because it does include 2.6 features as well, but the whole concept of retries has not been in this guide before :)

Added the new section on retries and updated the language and screenshots for the other sections.

PR #2200, opened 2023-04-28
Update dag-best-practices.md

RNHTTR:


learn/dag-best-practices.md
```diff
@@ -229,8 +229,9 @@ Additionally, if you change the `start_date` of your DAG you should also change
 
 Changing the name of a DAG also creates a new entry in the database that powers the dashboard. Follow a consistent naming convention since changing a DAG's name doesn't delete the entry in the database for the old name.
 
-### Set retries at the DAG level
+### Set retries
 
 In a distributed environment where task containers are executed on shared hosts, it's possible for tasks to be killed off unexpectedly. When this happens, you might see  a [zombie process](https://en.wikipedia.org/wiki/Zombie_process) in the Airflow logs.
 
-Issues like this can be resolved by using task retries. The best practice is to set retries as a `default_arg` so they are applied at the DAG level and get more granular for specific tasks only where necessary. A good range is ~2–4 retries.
+Issues like this can be resolved by using task retries. Retries can be set at the DAG level using `default_args`, at the task level using the parameter `retries`, or at the deployment level by setting the environment variable `AIRFLOW__CORE__DEFAULT_TASK_RETRIES`. Setting retries to 2 will protect a given task from problems common to distributed environments.
```
lzdanski, line 236:
```suggestion
You can resolve issues like zombie processes by using task retries. Retries can be set at different levels with the following:

- **Tasks:**  Use the parameter `retries`
- **DAGs:** Use `default_args`
- **Deployments:** Set the environment variable, `AIRFLOW__CORE__DEFAULT_TASK_RETRIES`

Set the value to `2` to protect a task from problems common to distributed environments.
```

It can be a little hard to skim a list like this, so I broke it out into bullet points. But I also have a couple of clarifying questions -  

1. In the last sentence, does **retries** only refer to the parameter `retries` or does it apply to all of these? 
2. I have these ordered from most specific to most general - is that ok?
3. What are accepted values? What's the default behavior?

lzdanski (changes requested):
This is a really cool feature - I have some suggestions about organizing the content and questions about the details.

TJaniF (commented):
Thank you @RNHTTR for the addition! One note: we have another PR open #2195 to reframe the rerun guide to include retires. I think having this chapter in best practices makes sense and your addition is great, just noting this to not forget to link to that guide from here depending on which PR gets merged first :)

learn/dag-best-practices.md
kentdanas: Bullets look good to me. Some general answers to your questions:

1. "retries" applies to all three parameters. They are all controlling the same behavior, just at different levels of specificity.
2. I don't think the order matters here too much, yours is fine with me.
3. I wouldn't say there's an "accepted" value for retries since it's so use case specific. The default for all of them is 0 retries.

kentdanas: I think we should add a note here about precedence since we're showing all three options. Precedence is the same as the current order of the bullets - if you set a deployment level variable it will apply to all tasks, unless it is overridden in a specific DAG or a specific task. Same with DAG-level args -> task-level args

kentdanas (approved):
LGTM, thanks for adding this @RNHTTR! Couple follow up comments/answers to Laura's suggestion, once that's resolved I think this is good to go.

lzdanski (approved):

learn/dag-best-practices.md
```diff
@@ -229,8 +229,13 @@ Additionally, if you change the `start_date` of your DAG you should also change
...
 
 In a distributed environment where task containers are executed on shared hosts, it's possible for tasks to be killed off unexpectedly. When this happens, you might see  a [zombie process](https://en.wikipedia.org/wiki/Zombie_process) in the Airflow logs.
 
-Issues like this can be resolved by using task retries. The best practice is to set retries as a `default_arg` so they are applied at the DAG level and get more granular for specific tasks only where necessary. A good range is ~2–4 retries.
+You can resolve issues like zombies by using task retries. Retries can be set at different levels with the following precedence:
+1. **Tasks:**  Pass the `retries` parameter to the task's Operator.
+2. **DAGs:** Include `retries` in a DAG's `default_args` object.
+3. **Deployments:** Set the environment variable `AIRFLOW__CORE__DEFAULT_TASK_RETRIES`.
+
+Setting retries to `2` will protect a task from most problems common to distributed environments.
```
kentdanas, line 241:
```suggestion
Setting retries to `2` will protect a task from most problems common to distributed environments. For more on using retries, see [Rerun DAGs and Tasks](rerunning-dags.md).
```

PR #2210, opened 2023-04-28
software: update correct config path for vector index pattern

pgvishnuram:
Docs change needed:   update correct config path for vector index pattern


danielhoherd (approved):
The new value is correct. The prior value was for the airflow-chart itself, but houston is what triggers the installation of the airflow chart, so users need to the value through houston since there is no direct way to do it.

lzdanski:
Thanks @danielhoherd! Merging.

PR #2218, opened 2023-05-02
Make alerting docs reflect GA status

jlaneve:


lzdanski:
@jlaneve - I think that this change is included in a larger update @jwitz is working on https://github.com/astronomer/docs/pull/2217

PR #2220, opened 2023-05-02
5.2 link checker - CI/CD Templates

lzdanski:
While investigating the broken link for a download from GitHub, I found that the page it originates from is an unlisted version of the outdated CI/CD Set Up page. https://github.com/astronomer/docs/issues/2211. 

This PR 

* Removes the [old CI/CD](https://docs.astronomer.io/astro/ci-cd)
* Updates the redirect file
* Updates relevant links to old CI/CD link path

Can you please confirm that the link replacements make sense? It was hard to judge when they needed to point to the [Template Overview](https://docs.astronomer.io/astro/ci-cd-templates/template-overview) and when they should go to the new [Choose a strategy](https://docs.astronomer.io/astro/set-up-ci-cd) page.


static/_redirects
```diff
@@ -96,7 +96,7 @@
 /cloud/api-keys /astro/api-keys 301
 /cloud/airflow-alerts /astro/airflow-alerts 301
 /cloud/airflow-api /astro/airflow-api 301
-/cloud/ci-cd /astro/ci-cd 301
+/cloud/ci-cd /astro/set-up-ci-cd 301
```
jwitz, line 99:
```suggestion
/cloud/ci-cd /astro/set-up-ci-cd 301
/astro/ci-cd /astro/set-up-ci-cd 301
```
We'll need redirects for both, since the Astro link might still be accessed directly.

PR #2247, opened 2023-05-08
update podman docs

sunkickr:
Let users know that rootful prereg is only necessary for  mac M1 machines

### DIFF
--- a/astro/cli/configure-cli.md
+++ b/astro/cli/configure-cli.md
@@ -120,7 +120,7 @@ Set up Podman on a Mac operating system so you can run Apache Airflow locally an
 
 - Podman 3 or later. See [Getting started with Podman](https://podman.io/getting-started/).
 - A running Podman machine with at least 4 GB of RAM. To confirm that Podman is running, run `podman ps`.
-- Turn on rootful mode for Podman by using `podman machine set --rootful`.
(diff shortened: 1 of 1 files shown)


astro/cli/configure-cli.md
```diff
@@ -120,7 +120,7 @@ Set up Podman on a Mac operating system so you can run Apache Airflow locally an
 
 - Podman 3 or later. See [Getting started with Podman](https://podman.io/getting-started/).
 - A running Podman machine with at least 4 GB of RAM. To confirm that Podman is running, run `podman ps`.
-- Turn on rootful mode for Podman by using `podman machine set --rootful`.
+- Mac M1 machines. Turn on rootful mode for Podman by using `podman machine set --rootful`.
```
lzdanski, line 123:
```suggestion
- (Mac M1 machines) Turn on rootful mode for Podman by using `podman machine set --rootful`.
```

We had a styling convention change! Decided to instead of having ` <Optional.>` we now use `(Optional)`. I haven't finished updating all the docs with that change so we're fixing it as we go. I think it applies to scenarios like this as well, where we're defining a context in this way.

lzdanski (changes requested):
Styling change

lzdanski (approved):
Even better. When does this need to be merged by?

PR #2251, opened 2023-05-09
Update airflow-dbt.md

sreekarj18:
Project yml file name should be "dbt_project" its "dbt-project" in the doc. Due to this error project yml file cannot be located in docker environment

### DIFF
--- a/learn/airflow-dbt.md
+++ b/learn/airflow-dbt.md
@@ -111,7 +111,7 @@ In this tutorial we will use an example dbt job that consists of two dbt models.
 
 2. In the `dbt` folder, create a folder called `my_energy_project`.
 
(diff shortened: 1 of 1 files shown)


sreekarj18 (commented):
Project yml file name should be "dbt_project" its "dbt-project" in the doc. Due to this error project yml file cannot be located in docker environment

TJaniF (approved):
Thanks for the great catch @sreekarj18 ! 🙂

@lzdanski @jwitz: I double checked and the file being named `dbt_project.yml` is a [dbt level config](https://docs.getdbt.com/reference/dbt_project.yml) and what I did in my sample repo. So basically this was my typo, sorry! 😅

PR #2252, opened 2023-05-09
Update cosmos_dag.py

sreekarj18:
Getting key error due to upper case keys, but in dataframe they are in lower case, to resolve it they need to be changed to lower case

### DIFF
--- a/code-samples/dags/airflow-dbt/cosmos_dag.py
+++ b/code-samples/dags/airflow-dbt/cosmos_dag.py
@@ -28,26 +28,26 @@ def log_data_analysis(df: pd.DataFrame):
     If the latest year in the data was also the year with the highest % of solar
     capacity and/or the year with the highest % of renewables capacity a
     celebratory message is logged as weel."""
-
-    latest_year = df.YEAR.max()
-    year_with_the_highest_solar_pct = df.loc[df["SOLAR_PCT"].idxmax(), "YEAR"]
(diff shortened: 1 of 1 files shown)


sreekarj18 (commented):
Getting key error due to upper case keys, but in dataframe they are in lower case, to resolve it they need to be changed to lower case

TJaniF:
Hello @sreekarj18 thank you for your PR! Good catch, this turned out to be database specific! 

I originally wrote the guide using a Postgres database, where the current upper-case `YEAR` is the correct key. But testing it with Snowflake just now, your version is the one that works. 
I think since the tutorial is written using Postgres, rather than changing the code it makes sense to add a comment at line 31 that says:

`# Note that if you are using another database than Postgres, you might need to adjust the pd.DataFrame column names in this code to be lowercase.`

@lzdanski @jwitz database specifics strike again. I think a code comment should be enough since the error is easy to interpret and fix for users with a database that creates lower case column names. But open to other opinions and to adding a note in the text as well :) 

@sreekarj18 if Laura and/or Jake agree with my comment idea feel free to change this PR to be just the insertion of that comment so you get the credit for finding this :)

jwitz:
I like the idea of an in-code note! What about

`# Written for Postgres. If you are using database other than Postgres, you might need to adjust the pd.DataFrame column names in this code to be lowercase.`

TJaniF:
Closing this PR in favour of #2301 :)

Issue #2234, opened 2023-05-05
The new cncf provider buit-in with Runtime 8 needs a default connection kubernetes_default


abhishekbhakat:
**Affected Document:**
https://docs.astronomer.io/astro/kubernetespodoperator

**Problem Type (Priority):**

- [ ] Error (Critical): This document has existing information that's incorrect and prevents readers from completing the documented setup (Add the `error/critical` label to this issue)
- [x] Misleading (Important): This document could benefit from additional context, guidance, or clarity (Add the `error/FYI` label to this issue)
- [ ] Typo (Low): Any issue related to spelling/ punctuation/ formatting (Add the `error/typo` label to this issue)

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**

**Additional Notes:**
Customer have started facing dag run failures after migrating to runtime 8 as it houses cncf providers 6.1. Whereas since cncf provider 6.0 a default connection `kubernetes_default` is mandatory. It would be nice to instructions to create a connection. A customer has pointed out these gaps in communication.

abhishekbhakat:
Fixed by https://github.com/apache/airflow/pull/31187

PR #2254, opened 2023-05-09
Add kubernetes cluster connection to KPO doc

abhishekbhakat:
An Airflow connection requirement as mentioned in the Issue #2234.


jwitz (changes requested):
Thanks for making this change @abhishekbhakat ! I tried simplifying the information to be a basic prerequisite. 

Is the `in_cluster` configuration information just an implementation detail for how Astro connects to the KPO cluster, or is that something the user can change or needs to take action on?

abhishekbhakat:
This was a separate issue but @dstandish  already fixed it.

Issue #2243, opened 2023-05-08
Update Runtimes considerations 


lzdanski:
**Affected Document:**
- https://docs.astronomer.io/astro/upgrade-runtime#version-upgrade-considerations
**Problem Type (Priority):**

- [x] Error (Critical): This document has existing information that's incorrect and prevents readers from completing the documented setup (Add the `error/critical` label to this issue)
- [ ] Misleading (Important): This document could benefit from additional context, guidance, or clarity (Add the `error/FYI` label to this issue)
- [ ] Typo (Low): Any issue related to spelling/ punctuation/ formatting (Add the `error/typo` label to this issue)

**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**

There is an incompatibility between:
apache-airflow-providers-cncf-kubernetes==6.1.0
apache-airflow-providers-google==10.0.0
These providers are built in for Runtime 8.0.0
A recent [PR](https://github.com/apache/airflow/pull/28125) was made 4 weeks ago that added a function get_xcom_sidecar_container_resources to be present in KubernetesHook but google uses GKEPodHook which doesn’t has this function just yet, so it breaks GKEStartPodOperator with new CNCF providers.

**Additional Notes:**

https://astronomer.slack.com/archives/C020PVBJPKK/p1683295430151119?thread_ts=1682867833.261559&cid=C020PVBJPKK

PR #2267, opened 2023-05-10
Add v8.1.0

lzdanski:
Follow up on https://github.com/astronomer/docs/issues/2243

- Adds note to runtime upgrade considerations about KPO
- adds note about runtine versions to KPO docs

see https://astronomer.slack.com/archives/C020PVBJPKK/p1683711074298299?thread_ts=1683686822.725169&cid=C020PVBJPKK

### DIFF
--- a/astro/upgrade-runtime.md
+++ b/astro/upgrade-runtime.md
@@ -31,6 +31,7 @@ This is where you'll find the upgrade considerations for specific Astro Runtime 
 #### Runtime 8 (Airflow 2.6)
 
 ##### Package dependency conflicts
+
 Astro Runtime 8.0.0 includes fewer default dependencies than previous versions. Specifically, the following provider packages are no longer installed by default:
 
 - `apache-airflow-providers-apache-hive`
@@ -43,6 +44,8 @@ Astro Runtime 8.0.0 includes fewer default dependencies than previous versions. 
 
 If your DAGs depend on any of these provider packages, add the provider packages to your Astro project `requirements.txt` file before upgrading. You can also [pin specific provider package versions](#optional-pin-provider-package-versions) to ensure that none of your provider packages change after upgrading.
 
+##### Provider incompatibilities
+
 There is also an incompatibility between Astro Runtime 8.0.0 and the following provider packages:
 
 - `apache-airflow-providers-cncf-kubernetes==6.1.0`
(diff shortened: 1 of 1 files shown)


lzdanski:
@jwitz I added the KPO things to the runtimes considerations and related doc.

astro/upgrade-runtime.md
```diff
@@ -52,12 +55,23 @@ That can be resolved by pinning `apache-airflow-providers-cncf-kubernetes==5.2.2
...
+##### `KubernetesPodOperators` and Runtime 8.1.0
+
+There is an incompatibility between `KubernetesPodOperators` and Astro Runtime 8.0.0. A fix for this issue has been included in Version 8.1.0. 
+
+To use this fix, update your Dockerfile to use the 8.1.0 image, like in the following example:
+
+```sh
+FROM [quay.io/astronomer/astro-runtime:8.1.0](http://quay.io/astronomer/astro-runtime:8.1.0)
+```
+
```
jwitz, line 67:
```suggestion
```

astro/kubernetespodoperator.md
```diff
@@ -28,6 +28,7 @@ On Astro, the Kubernetes infrastructure required to run the KubernetesPodOperato
 
 - Cross-account service accounts are not supported on Pods launched in an Astro cluster. To allow access to external data sources, you can provide credentials and secrets to tasks.
 - PersistentVolumes (PVs) are not supported on Pods launched in an Astro cluster.
+- There is a known incompatibility with [Astro Runtime Version 8.0.0](upgrade-runtime.md#kubernetespodoperators-and-runtime-810). Make sure you do not use that version by checking your Dockerfile configuration.
```
jwitz, line 31:
```suggestion
```

jwitz (approved):
LGTM @lzdanski except I don't think we need to document a bug in an early patch version as a limitation. All customers should have been notified of the need to upgrade to 8.1.0, so there's nothing we need to proactively warn about in docs. 

We already recommend always using the latest patch version across a variety of docs and also don't support non-latest patch versions, so to me this is just the equivalent of a bug fix. More context: https://astronomer.slack.com/archives/C03CT3DNZMH/p1683734792424019?thread_ts=1683725890.816619&cid=C03CT3DNZMH

jwitz (changes requested):
Actually @lzdanski let's hold off on this until we get clarity on https://github.com/astronomer/docs/pull/2254. If that PR is correct, we might need a separate consideration for that connection configuration (and it might conflict with our existing considerations here).

jwitz (approved):
@lzdanski I just resolved this with @abhishekbhakat and updated my suggestion to include a replacement consideration! Believe it or not, this is actually a separate issue to what we've been talking about, but it still applies until we issue a new Runtime patch. After that, we can delete it.

PR #2271, opened 2023-05-11
update workload identity docs

melugoyal:
updating per the changes in https://github.com/astronomer/astro/pull/13102. i suppose we shouldn't release this docs change till that PR makes it to prod

### DIFF
--- a/astro/connect-aws.md
+++ b/astro/connect-aws.md
@@ -137,7 +137,7 @@ Authorization is the process of verifying a user or service's permissions before
     ]}>
 <TabItem value="AWS IAM roles">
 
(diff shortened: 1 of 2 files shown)


melugoyal:
https://github.com/astronomer/docs/pull/2302 was merged instead

astro/connect-aws.md
```diff
@@ -154,21 +154,19 @@ To grant an Astro cluster access to a service that is running in an AWS account
                 "Effect": "Allow",
                 "Principal": {
                     "AWS": [
-                        "arn:aws:iam::<dataplane-AWS-account-ID>:role/AirflowS3Logs-<cluster-ID>"
+                        "<workload-identity-role>"
```
melugoyal, line 157: @jwitz looks like this change was missed in https://github.com/astronomer/docs/pull/2302/files which was merged instead. i think this is important to not have confusion about referring to the old hardcoded role. can we update the docs accordingly?

PR #2307, opened 2023-05-18
Hosted launch: Worker queues and executor updates

jwitz:


astro/configure-worker-queues.md
```diff
@@ -30,40 +30,39 @@ Worker queues can enhance performance, decrease cost, and increase the reliabili
...
 
 ## Worker queue settings
 
 You can configure each worker queue on Astro with the following settings:
 
 - **Name:** The name of your worker queue. Use this name to assign tasks to the worker queue in your DAG code. Worker queue names must consist only of lowercase letters and hyphens. For example, `machine-learning-tasks` or `short-running-tasks` or `high-cpu`.
-- **Worker Type:** The size and type of workers in the worker queue, defined as a node instance type that is supported by the cloud provider of your cluster. For example, a worker type might be `m5.2xlarge` or `c6i.4xlarge` for a Deployment running on an AWS cluster. A worker’s total available CPU, memory, storage, and GPU is defined by its worker type. Actual worker size is equivalent to the total capacity of the worker type minus Astro’s system overhead.
+- **Worker Type:** The size of workers in the worker queue, for example A5 or A20. A worker’s total available CPU, memory, storage, and GPU is defined by its worker size.
```
conorbev, line 44: We don't support GPUs yet so I would remove that part altogether. 

Also, for now, all 3 worker types have 5 GiB ephemeral storage. 

So maybe this should say:

`A worker’s total available CPU and memory is defined by its worker size. Each Worker Type comes with 5 GiB of ephemeral storage.`

astro/configure-worker-queues.md
```diff
@@ -30,40 +30,39 @@ Worker queues can enhance performance, decrease cost, and increase the reliabili
...
-- **Worker Type:** The size and type of workers in the worker queue, defined as a node instance type that is supported by the cloud provider of your cluster. For example, a worker type might be `m5.2xlarge` or `c6i.4xlarge` for a Deployment running on an AWS cluster. A worker’s total available CPU, memory, storage, and GPU is defined by its worker type. Actual worker size is equivalent to the total capacity of the worker type minus Astro’s system overhead.
+- **Worker Type:** The size of workers in the worker queue, for example A5 or A20. A worker’s total available CPU, memory, storage, and GPU is defined by its worker size. 
 - **Max Tasks per Worker:** The maximum number of tasks that a single worker can run at a time. If the number of queued and running tasks exceeds this number, a new worker is added to run the remaining tasks. This value is equivalent to [worker concurrency](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#worker-concurrency) in Apache Airflow. It is 16 by default.
```
conorbev, line 45: This part is different in Hosted. The setting is labeled as CONCURRENCY ( instead of Max Tasks per Worker ). In addition, the default is no longer 16, we are changing the default based on the size of the Worker Type ( try this out in dev and you'll see ).

If you choose A5 the default will change to 5
A10 --> 10
A20 --> 20

This is an advanced configuration and we don't recommend a user changing it unless they are sure.

astro/configure-worker-queues.md
```diff
@@ -30,40 +30,39 @@ Worker queues can enhance performance, decrease cost, and increase the reliabili
...
 - **Worker Count**: The minimum and maximum number of workers that can run at a time.  The number of running workers changes regularly based on Maximum Tasks per Worker and the current number of tasks in a `queued` or `running` state. By default, the minimum number of workers is 1 and the maximum is 10.
```
conorbev, line 46: This is labeled as MIN # WORKERS / MAX # WORKERS in Hosted.

In Hosted the min is also 0, so that's different too ( It also defaults to 0 ).

astro/executors.md
```diff
@@ -108,7 +108,7 @@ On Astro, you can configure Kubernetes executor in the following ways:
 
 ### Kubernetes worker autoscaling logic
 
-By default, each task on Astro runs in a dedicated Kubernetes Pod with 1 CPU and 256Mi of memory. These Pods run on a worker node in your Astro data plane. If a worker node can't run any more Pods, Astro automatically provisions a new worker node to begin running any queued tasks in new Pods.
+By default, each task on Astro runs in a dedicated Kubernetes Pod with 1 CPU and 256Mi of memory. These Pods run on a worker node in your Astro Deployment. If a worker node can't run any more Pods, Astro automatically provisions a new worker node to begin running any queued tasks in new Pods.
```
conorbev, line 111: hmm, this plays into the updated KPO/KE settings for which there is now a PR: https://github.com/astronomer/astro/pull/13352

This part: `These Pods run on a worker node in your Astro Deployment. If a worker node can't run any more Pods, Astro automatically provisions a new worker node to begin running any queued tasks in new Pods.` is not really relevant on Hosted and could be removed.

astro/executors.md
```diff
@@ -275,13 +275,13 @@ with DAG(
 
 ### Change the worker node type
 
-A Deployment using the Kubernetes executor runs worker Pods on a single `default` worker queue. You can change the type of worker node that this queue uses from the Cloud UI.
+A Deployment using the Kubernetes executor runs worker Pods on a single `default` worker queue. You can change the type of worker that this queue uses from the Cloud UI.
```
conorbev, line 278: This will be different in Hosted. You'll just set the quota for your KE pods, they won't be related to a worker type. 
image

conorbev (commented):
My general feeling here ( quick review, having the netlify preview would prob help ) is that although we made the names match and called them "Worker Queues"  the UI is actually _quite_ different for configuring these things between Hosted and Hybrid both in terms of some terminology and then very different when it comes to configuring KPO and KE quotas..

I didn't see it from a quick scan but if there is anywhere that mentions "1 worker == 1 node" that is no longer true on Hosted ( it's still true on Hybrid atm ).

astro/configure-worker-queues.md
jwitz: I'm making a note to link to the Hosted resource reference on worker info for this detail.

astro/configure-worker-queues.md
```diff
@@ -9,7 +9,7 @@ import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 import {siteVariables} from '@site/src/versions';
 
-By default, all tasks run in a default worker queue that does not require configuration or code. If you're using the Celery executor, you can create additional worker queues to enable multiple worker types or configurations for different groups of tasks, and assign tasks to queues in your DAG code. For more information about Airflow executors on Astro, see [Manage executors](executors.md).
+By default, all tasks using the Celery executor run in a `default` worker queue that does not require configuration or code. You can create additional worker queues to enable multiple worker types or configurations for different groups of tasks, and assign tasks to queues in your DAG code. For more information about Airflow executors on Astro, see [Manage executors](executors.md).
```
manmeetkaur, line 12: @jwitz not sure what we mean by this - `does not require configuration or code` . Default worker queue can also be configured based on the default workloads a user wants to run, like max tasks , max workers etc., which is same for any other worker Queue. And I think by `code`, we mean to say we do not require any changes to DAG code for using it. 

Given the above, can we rephrase something like "By default, all tasks using the Celery executor run in a `default` worker queue without any code change in your DAG. You can create additional worker queues to enable multiple worker types or configurations for different groups of tasks, and assign tasks to queues by changing your DAG code. See this [example](https://docs.astronomer.io/astro/configure-worker-queues?tab=classicoperator#assign-tasks-to-a-worker-queue) for details.

astro/configure-worker-queues.md
```diff
@@ -30,40 +30,39 @@ Worker queues can enhance performance, decrease cost, and increase the reliabili
...
 
-- You are running Task A and Task B in a Deployment on an AWS cluster.
+- You are running Task A and Task B in a Deployment.
 - Task A and Task B are dependent on each other, so they need to run in the same Deployment.
-- Task A is a long-running task that uses a lot of CPU and little memory, while Task B is a short-running task that uses minimal amounts of CPU and memory.
+- Task A is a long-running task that uses a lot of CPU and memory, while Task B is a short-running task that uses minimal amounts of CPU and memory.
 
-You can assign Task A to a worker queue that is configured to use the [`c6i.4xlarge`](https://aws.amazon.com/ec2/instance-types/c6i/) worker type that's optimized for compute. Then, you can assign Task B to a worker queue that is configured to use the [`m5.xlarge`](https://aws.amazon.com/ec2/instance-types/m5) worker type that is smaller and optimized for general usage.
+You can assign Task A to a worker queue that is configured to use the A20 worker type that's optimized for running larger tasks. Then, you can assign Task B to a worker queue that is configured to use the A5 worker type that's smaller and optimized for general usage.
```
manmeetkaur, line 37:
```suggestion
You can assign Task A to a worker queue that is configured to use the A20 worker type that's optimized for running compute-heavy tasks. Then, you can assign Task B to a worker queue that is configured to use the A5 worker type that's smaller and optimized for general usage.
```

astro/configure-worker-queues.md
```diff
@@ -30,40 +30,39 @@ Worker queues can enhance performance, decrease cost, and increase the reliabili
...
+- **Worker Type:** The size of workers in the worker queue, for example A5 or A20. A worker’s total available CPU, memory, and storage is defined by its worker size. For more information about worker types, see [LINK]
+- **Concurrency:** The maximum number of tasks that a single worker can run at a time. If the number of queued and running tasks exceeds this number, a new worker is added to run the remaining tasks. This value is equivalent to [worker concurrency](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#worker-concurrency) in Apache Airflow. The default for your worker type is suitable for most use cases.
+- **Min # Workers / Max # Workers**: The minimum and maximum number of workers that can run at a time.  The number of running workers changes regularly based on **Concurrency** and the current number of tasks in a `queued` or `running` state.
```
manmeetkaur, line 46: Feel like we should mention here about auto-scaling - "Your workers will auto-scale up to the max number of workers based on the concurrency and number of tasks in `queued` or `running` state."

astro/configure-worker-queues.md
```diff
@@ -30,40 +30,39 @@ Worker queues can enhance performance, decrease cost, and increase the reliabili
...
+- **Min # Workers / Max # Workers**: The minimum and maximum number of workers that can run at a time.  The number of running workers changes regularly based on **Concurrency** and the current number of tasks in a `queued` or `running` state. 
 
-### Default worker queue
+:::info Alternative Astro Hybrid setup 
 
-Each Deployment requires a worker queue named `default` to run tasks. Tasks that are not assigned to a worker queue in your DAG code are executed by workers in the default worker queue.
+On Astro Hybrid clusters, worker type is defined as a node instance type that is supported by the cloud provider of your cluster. For example, a worker type might be `m5.2xlarge` or `c6i.4xlarge` for a Deployment running on a Hybrid AWS cluster hosted on your cloud. Actual worker size is equivalent to the total capacity of the worker type minus Astro’s system overhead.
```
manmeetkaur, line 50: would be helpful to link here to the Cluster Setting Reference which gives the details about the amount of cpu and memory available minus the system overhead.

astro/configure-worker-queues.md
```diff
@@ -30,40 +30,39 @@ Worker queues can enhance performance, decrease cost, and increase the reliabili
...
 
-If you don’t change any settings in the default worker queue:
+Your Organization can enable up to 10 different worker types for each Hybrid cluster. After a worker type is enabled on an Astro Hybrid cluster, the worker type becomes available to any Deployment in that cluster and appears in the **Worker Type** menu of the Cloud UI.
```
manmeetkaur, line 52: this is not true for Azure, it only have 6 worker types - https://docs.astronomer.io/astro/resource-reference-azure#supported-worker-node-pool-instance-types

astro/configure-worker-queues.md
```diff
@@ -30,40 +30,39 @@ Worker queues can enhance performance, decrease cost, and increase the reliabili
...
 
-- A maximum of 16 tasks can run at one time per worker. If more than 16 tasks are queued or running, a new worker is added to run the remaining tasks.
-- A maximum of 10 workers can run at once, meaning that a maximum of 160 tasks can be in a `running` state at a time. Remaining tasks will stay in a `queued` or `scheduled` state until running tasks complete.
+1. Review the list of supported worker types for your cloud provider. See [AWS](resource-reference-aws.md#worker-node-types), [Azure](resource-reference-azure.md#worker-node-size-resource-reference), or [GCP](resource-reference-gcp.md#worker-node-size-resource-reference).
+2. Contact [Astronomer support](https://cloud.astronomer.io/support) with the name of the worker type(s) you want to enable for your cluster. For example, `m6i.2xlarge`.
```
manmeetkaur, line 55:
```suggestion
2. Contact [Astronomer support](https://cloud.astronomer.io/support) with the name of the worker type(s) you want to enable for your cluster along with the cluster name. For example, `m6i.2xlarge` for my-AWS-dev-Cluster.
```

astro/configure-worker-queues.md
```diff
@@ -30,40 +30,39 @@ Worker queues can enhance performance, decrease cost, and increase the reliabili
...
 
-## Request a worker type
+:::
 
-Your organization can enable up to 10 different worker types for each cluster. After a worker type is enabled on an Astro cluster, the worker type becomes available to any Deployment in that cluster and appears in the **Worker Type** menu of the Cloud UI.
+## Default worker queue
 
-1. Review the list of supported worker types for your cloud provider. See [AWS](resource-reference-aws.md#worker-node-types), [Azure](resource-reference-azure.md#worker-node-size-resource-reference), or [GCP](resource-reference-gcp.md#worker-node-size-resource-reference).
-2. Contact [Astronomer support](https://cloud.astronomer.io/support) with the name of the worker type(s) you want to enable for your cluster. For example, `m6i.2xlarge`.
+Each Deployment requires a worker queue named `default` to run tasks. Tasks that are not assigned to a worker queue in your DAG code are executed by workers in the default worker queue.
```
manmeetkaur, line 63:
```suggestion
Each Deployment requires a worker queue named `default` to run tasks. Tasks that are not specifically assigned to a worker queue in your DAG code are executed by workers in the default worker queue.
```

astro/executors.md
```diff
@@ -26,7 +26,7 @@ The Celery executor is the default for all new Deployments. It uses a group of w
 The Celery executor is a good option for most use cases. Specifically, the Celery executor is a good fit for your Deployment if:
 
 - You're just getting started with Airflow.
-- You want to use multiple worker queues. This allows you to use multiple worker node types for different types of tasks and optimize for task performance. See [Configure worker queues](configure-worker-queues.md).
+- You want to use multiple worker queues. This allows you to use multiple worker types for different types of tasks and optimize for task performance. See [Configure worker queues](configure-worker-queues.md).
```
manmeetkaur, line 29: may be start with the problem statement first, because a user coming from OSS airflow is not familiar with Worker Queues - "You want to use different worker types based on the type of workload to optimize your task performance. It allows you to create worker queues with different worker types and assign specific tasks in your DAG to use them."

astro/executors.md
```diff
@@ -60,10 +60,17 @@ On Astro, you can configure Celery executor in the following ways:
 - The type and size of your workers.
 - The minimum and maximum number of workers that your Deployment can run at a time.
 - The number of tasks that each worker can run at a time.
+- The resources used by the [KubernetesPodOperator](kubernetespodoperator.md).
```
manmeetkaur, line 63: I am confused why this is here, this is not an Executor configuration but an Operator configuration.

astro/executors.md
```diff
@@ -103,12 +110,29 @@ For each worker queue on your Deployment, you have to specify certain settings t
 
 On Astro, you can configure Kubernetes executor in the following ways:
 
-- Customize individual tasks, including CPU and memory requests, using a `pod_override` configuration. 
-- Change the worker node type on which your Pods run.
+- Change the resource usage for the default Pods on which your tasks run.
```
manmeetkaur, line 113: since only this is available via Cloud UI, should we explicitly say "using Cloud UI"?

astro/executors.md
```diff
@@ -103,12 +110,29 @@ For each worker queue on your Deployment, you have to specify certain settings t
...
+- Customize individual Pods for tasks, including CPU and memory requests, using a `pod_override` configuration. 
+
+### Set defaults and limits for Kubernetes executor Pods in the Cloud UI
+
+You can configure the minimum and maximum resources that any Kubernetes executor Pod uses from the Cloud UI. Tasks that make requests above your resource quotas or **Max Pod Size** will fail.
```
manmeetkaur, line 118: as per the UI, we can only configure maximum
```suggestion
You can configure the maximum resources that any Kubernetes executor Pod uses from the Cloud UI. Tasks that make requests above your resource quotas or **Max Pod Size** will fail.
```

astro/executors.md
```diff
@@ -103,12 +110,29 @@ For each worker queue on your Deployment, you have to specify certain settings t
...
+
+1. In the Cloud UI, select a Deployment.
+2. Click **Resource quotas**.
+3. Configure the following values:
 
-### Kubernetes worker autoscaling logic
+    - **CPU quota**: The maximum amount of CPU for all currently running Pods on your Deployment. 
+    - **Memory Quota**: The maximum amount of memory for all currently running Pods on your Deployment. 
+    - **Default CPU**: The amount of CPU that a task Pod uses if it has no configured requests.
+    - **Default Memory**: The amount of memory that a task Pod uses if it has no configured requests.
```
manmeetkaur, line 127:
```suggestion
    - **Default CPU**: The amount of CPU that a task Pod uses if it has no configured requests.
    - **Default Memory**: The amount of memory that a task Pod uses if it has no configured requests.
```
This is just for viewing, no changes are allowed. they can only changed via `pod_override` in the DAG code

manmeetkaur: Screenshot 2023-05-22 at 10 15 16 PM

astro/executors.md
```diff
@@ -103,12 +110,29 @@ For each worker queue on your Deployment, you have to specify certain settings t
...
 
-By default, each task on Astro runs in a dedicated Kubernetes Pod with 1 CPU and 256Mi of memory. These Pods run on a worker node in your Astro data plane. If a worker node can't run any more Pods, Astro automatically provisions a new worker node to begin running any queued tasks in new Pods.
+Your CPU quota and memory quota determine your **Max Pod Size**, which is the maximum amount of resources that a task can request for its Pod. If your CPU and memory quota exceed a certain limit, your **Max Pod Size** is determined by the size of the Astro-hosted infrastructure running your tasks.
```
manmeetkaur, line 129: this is not clear. What do we mean by certain limit? for example, if I set the cPU and memory max quota from cloud ui, and if that limit is reached , then the max limit is decided by my infrastructure? If yes, then what is the use of Max Pod Size ? Or we are trying to say if the Max Pod Size requested is more than the infrastructure, then the actual limits are dictated by infrastructure available ?

astro/executors.md
```diff
@@ -103,12 +110,29 @@ For each worker queue on your Deployment, you have to specify certain settings t
...
+
+:::info Alternative setup on Astro Hybrid
+
+On Astro Hybrid, Pods run on a worker node in your Astro cluster. If a worker node can't run any more Pods, Astro automatically provisions a new worker node to begin running any queued tasks in new Pods. To give your tasks more or less resources, change the worker type in the task's worker queue and then change your resource requests using a `pod_override` configuration. See [(Hybrid clusters only) Change the Kubernetes executor's worker node type](#hybrid-clusters-only-change-the-kubernetes-executors-worker-node-type).
```
manmeetkaur, line 133: so this means that Kubernetes Executor on Hosted does not auto-scale ?

astro/executors.md
```diff
@@ -273,15 +297,21 @@ with DAG(
     )
 ```
 
-### Change the worker node type
+### (Hybrid clusters only) Change the Kubernetes executor's worker node type
+
+:::info
+
+This topic applies only to Astro Hybrid users. To see whether you're an Astro Hybrid user, open your Organization in the Cloud UI and go to **Settings** > **General**. Your version of Astro is listed under **Product Type**.
```
manmeetkaur, line 304: may be "version of Astro" is not the right term here. "Your Astro product type is listed under **Product Type**"

astro/executors.md
```diff
@@ -273,15 +297,21 @@ with DAG(
...
+
+:::
 
-A Deployment using the Kubernetes executor runs worker Pods on a single `default` worker queue. You can change the type of worker node that this queue uses from the Cloud UI.
+A Hybrid Deployment using the Kubernetes executor runs worker Pods on a single `default` worker queue. You can change the type of worker that this queue uses from the Cloud UI.
```
manmeetkaur, line 308: I don't think we should confuse the user by saying "Hybrid Deployment" - Deployment is a Deployment and it can run on Astro Hybrid or Astro Hosted. So may be rephrase like , "A Deployment running on Astro Hybrid with Kubernetes Executor ..."

astro/kubernetespodoperator.md
```diff
@@ -35,6 +35,21 @@ On Astro, the Kubernetes infrastructure required to run the KubernetesPodOperato
...
+2. Click **Resource quotas**.
+3. Configure the following values:
+
+    - **CPU quota**: The maximum amount of CPU for all currently running Pods on your Deployment. 
+    - **Memory Quota**: The maximum amount of memory for all currently running Pods on your Deployment. 
+    - **Default CPU**: The amount of CPU that a task Pod uses if it has no configured requests.
+    - **Default Memory**: The amount of memory that a task Pod uses if it has no configured requests.
+
+Your CPU quota and memory quota determine your **Max Pod Size**, which is the maximum amount of resources that a task can request for its Pod. If your CPU and memory quota exceed a certain limit, your **Max Pod Size** is determined by the size of the Astro-hosted infrastructure running your tasks.
+
```
manmeetkaur, line 52: aren't these steps repetition of what exists in executors.md? Should we just refer it to that section?

astro/configure-worker-queues.md
jwitz: Yes Azure only has 6 worker types, but if we had more you could technically enable up to 10. This is more of a message on cluster limits than it is about how many worker types we support in general.

astro/configure-worker-queues.md
jwitz: This word seems unnecessary. Is there any way to assign a task to a worker queue thats unspecifically?

astro/executors.md
jwitz: I think "On Astro" before this covers that. Also, you should be able to do this with the Astro CLI soon

astro/executors.md
jwitz: >  Or we are trying to say if the Max Pod Size requested is more than the infrastructure, then the actual limits are dictated by infrastructure available ?

This! Tried to make this clear in my edit. Good call out

astro/executors.md
jwitz: I think so, as there are no workers to autoscale on Hosted. It just launches new Pods in your cluster without worker limitations.

astro/kubernetespodoperator.md
jwitz: They are, but you can configure KPO separately from the Kubernetes executor (if you're running Celery). Because so many users look for KPO information, I think it's okay to have some redundant steps to improve discoverability of the setup.

astro/executors.md
```diff
@@ -63,7 +63,13 @@ On Astro, you can configure Celery executor in the following ways:
 
 You can set these configurations per worker queue, which is a set of configurations that apply to a group of workers in your Deployment. With the celery executor, you can configure multiple worker queues for different types of tasks and assign tasks to those queues in your DAG code.
 
-This topic discusses basic Celery executor configurations for a single worker queue. For instructions on how to configure multiple worker queues, see [Configure worker queues](configure-worker-queues.md). To add more worker types to your Deployment, see [Manage worker types](modify-cluster.md#manage-worker-types).
+This topic discusses basic Celery executor configurations for a single worker queue. For instructions on how to configure multiple worker queues, see [Configure worker queues](configure-worker-queues.md).
```
lzdanski, line 66:
```suggestion
The following section discusses basic Celery executor configurations for a single worker queue. For instructions on how to configure multiple worker queues, see [Configure worker queues](configure-worker-queues.md).
```

astro/executors.md
```diff
@@ -103,12 +109,27 @@ For each worker queue on your Deployment, you have to specify certain settings t
...
+3. Configure the following values:
 
-### Kubernetes worker autoscaling logic
+    - **CPU quota**: The maximum amount of CPU for all currently running Pods on your Deployment. 
+    - **Memory Quota**: The maximum amount of memory for all currently running Pods on your Deployment. 
 
-By default, each task on Astro runs in a dedicated Kubernetes Pod with 1 CPU and 256Mi of memory. These Pods run on a worker node in your Astro data plane. If a worker node can't run any more Pods, Astro automatically provisions a new worker node to begin running any queued tasks in new Pods.
+Your CPU quota and memory quota determine your **Max Pod Size**, which is the maximum amount of resources that a task can request for its Pod. If the CPU and memory quotas you specify exceed exceed the limits of Astro's infrastructure, your **Max Pod Size** is instead determined by the size of the Astro-hosted infrastructure running your tasks.
+
+:::info Alternative setup on Astro Hybrid
```
lzdanski, line 128:
```suggestion
:::info 

**Alternative setup on Astro Hybrid**
```

astro/executors.md
```diff
@@ -133,7 +154,7 @@ See [Manage task CPU and memory](#example-set-CPU-or-memory-limits-and-requests)
 
 #### Example: Set CPU or memory limits and requests
 
-One of the most common use cases for customizing a Kubernetes worker Pod is to request a specific amount of resources for a task. When requesting resources, make sure that your requests don't exceed the available resources in your current [Pod worker node type](#change-the-pod-worker-node-type).
+One of the most common use cases for customizing a Kubernetes worker Pod is to request a specific amount of resources for a task. When requesting resources, make sure that your requests don't exceed the resource limits in your [default Pod](#configure-the-default-kubernetes-pod).
```
lzdanski, line 157: **One of the most common use cases...** - I can see how this is direct, but it feels a little awkward a wording.

astro/executors.md
```diff
@@ -185,7 +206,7 @@ with DAG(
     resource_requirements_override_example()
 ```
 
-When this DAG runs, it launches a Kubernetes Pod with exactly 0.5m of CPU and 1024Mi of memory as long as that infrastructure is available in your cluster. Once the task finishes, the Pod terminates gracefully.
+When this DAG runs, it launches a Kubernetes Pod with exactly 0.5m of CPU and 1024Mi of memory as long as that infrastructure is available in your Deployment. Once the task finishes, the Pod terminates gracefully.
```
lzdanski, line 209:
```suggestion
When this DAG runs, it launches a Kubernetes Pod with exactly 0.5m of CPU and 1024Mi of memory, as long as that infrastructure is available in your Deployment. After the task finishes, the Pod terminates gracefully.
```

astro/kubernetespodoperator.md
```diff
@@ -35,6 +35,19 @@ On Astro, the Kubernetes infrastructure required to run the KubernetesPodOperato
 - An [Astro project](develop-project.md#create-an-astro-project).
 - An Astro [Deployment](create-deployment.md).
 
+## Configure the KubernetesPodOperator in the Cloud UI
+
+While you still need to configure the KubernetesPodOperator in your DAG code to define your task environment, you can set some safeguards on Astro so that tasks in your Deployment don't request more CPU or memory than expected. Do this by configuring default Pod limits and requests from the Cloud UI. If a task requests more CPU or memory than is currently allowed in your configuration, the task fails.
```
lzdanski, line 40:
```suggestion
While you still need to configure the KubernetesPodOperator in your DAG code to define your task environment, you can set some safeguards on Astro so that tasks in your Deployment don't request more CPU or memory than expected. Set safeguards by configuring default Pod limits and requests from the Cloud UI. If a task requests more CPU or memory than is currently allowed in your configuration, the task fails.
```

astro/kubernetespodoperator.md
```diff
@@ -67,7 +80,7 @@ This is the minimum configuration required to run tasks with the KubernetesPodOp
 
 ## Configure task-level Pod resources
 
-Astro automatically allocates resources to Pods created by the KubernetesPodOperator. Resources used by the KubernetesPodOperator are not technically limited, meaning that the operator could theoretically use any CPU and memory that's available in your cluster to complete a task. Because of this, Astronomer recommends specifying [compute resource requests and limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) for each task.
+Astro automatically allocates resources to Pods created by the KubernetesPodOperator. Resources used by the KubernetesPodOperator are not technically limited, meaning that the operator could theoretically use any CPU and memory that's available in your Deployment to complete a task. Because of this, Astronomer recommends specifying [compute resource requests and limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) for each task.
```
lzdanski, line 83:
```suggestion
Astro automatically allocates resources to Pods created by the KubernetesPodOperator. Resources used by the KubernetesPodOperator are not technically limited, which means that the operator could theoretically use any CPU and memory that's available in your Deployment to complete a task. Because of this, Astronomer recommends specifying [compute resource requests and limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) for each task.
```

astro/kubernetespodoperator.md
```diff
@@ -96,9 +109,9 @@ KubernetesPodOperator(
 )
 ```
 
-Applying the code above ensures that when this DAG runs, it will launch a Kubernetes Pod with exactly 800m of CPU and 3Gi of memory as long as that infrastructure is available in your cluster. Once the task finishes, the Pod will terminate gracefully.
+Applying the code above ensures that when this DAG runs, it will launch a Kubernetes Pod with exactly 800m of CPU and 3Gi of memory as long as that infrastructure is available in your Deployment. Once the task finishes, the Pod will terminate gracefully.
```
lzdanski, line 112:
```suggestion
Applying the previous code example ensures that when this DAG runs, it launches a Kubernetes Pod with exactly 800m of CPU and 3Gi of memory as long as that infrastructure is available in your Deployment. After the task finishes, the Pod will terminate gracefully.
```

lzdanski (commented):
This looks good overall! Just some minor things and a couple of link suggestions.

astro/configure-worker-queues.md
manmeetkaur: @jwitz make sense. So then let's rephrase to "up to 10 additional different worker types". Because total 11 are supported in GCS and AWS.

astro/kubernetespodoperator.md
manmeetkaur: okay, but then we have to change here as well that the Default CPU and Default Memory are not configurable . the problem is we have to keep updating in two places.

astro/configure-worker-queues.md
manmeetkaur: it is not, but we have to specify a "queue" to assign it to a particular worker Queue, else it just goes to default. So wanted to emphasize , if you change your DAG code and add a `queue` parameter, then only tasks are assigned to a different queue. Most of the OSS users are unfamiliar with Queue or Worker Queue concept, so I think it is important to make it clear.

astro/executors.md
jwitz, line 128:
```suggestion
:::info Alternative Astro Hybrid setup
```

astro/executors.md
jwitz, line 157:
```suggestion
You can request a specific amount of resources Kubernetes worker Pod so that a task always has enough resources to run successfully. When requesting resources, make sure that your requests don't exceed the resource limits in your [default Pod](#configure-the-default-kubernetes-pod).
```

astro/kubernetespodoperator.md
jwitz: Just updated, lmk what you think!

PR #1569, opened 2022-12-01
Issue form overhaul

jwitz:
To do: Add Notion home for unannounced new feature requests

PR #2325, opened 2023-05-23
Templates for Learn issues

TJaniF:
Heavily stolen from #1569 drafts for a form for Learn bugs/typos as well as a separate form to request new learn content. :) 
Very open to feedback, I tried to put in information I would find helpful to know when issues are submitted but very possible I forgot something. 

@kentdanas if you click on view file you can see the template mostly rendered :)


.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: target_user_description
+    attributes:
+      label: Describe who the content is intended for, e.g. Airflow beginners, expert users...?
+  - type: input
+    id: due_date
+    attributes:
+      label: Due date
```
kentdanas, line 46:
```suggestion
      label: Desired publish date
```

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+      description: For example, 'Q2 2023', '2/12/2023', 'Within the next month'
```
kentdanas, line 47:
```suggestion
      description: For example, is there a particular release, relevant event, etc. that this should be coordinated with?
```

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
+name: Request new Learn content
```
kentdanas, line 1: Wondering if we should title this to include 'updated content', etc. It seems like the majority of the requests we get are basically for extra sections within an existing doc, so I want to make it clear to users that this is the form they should use for that, instead of the bug report.

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
+name: Request new Learn content 
+description: New content on an existing Airflow feature or behavior
```
kentdanas, line 2:
```suggestion
description: New or updated content on an existing Airflow feature or use case
```

kentdanas (approved):
@TJaniF I think this looks great! Some very minor nit-picky comments, but honestly I'd say let's just try these and we can always make modifications later once we see how it's working.

TJaniF (commented):
Thank you @kentdanas !

@jwitz ready for you, I stole most of the fields from your templates :)

.github/ISSUE_TEMPLATE/learn-bug-report.yaml
```diff
@@ -0,0 +1,59 @@
+name: Request a fix in Learn content
+description: Any request related to existing documentation that contains a bug or is misleading
```
jwitz, line 2: Hm.. I don't know if we should have a separate bug fix template for Learn, especially if it's not all clear what Learn really encapsulates to an outside audience. Our current bug fix template doesn't discriminate by product, so I think that can cover Learn as well. The less choices someone has to make before they provide us with the fix info, the better.

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+      options:
+        - Additional information on a feature or behavior already described
+        - Add a previously not described Airflow feature or behavior to an existing page
+        - Add a new page
+    validations:
+      required: true
+  - type: textarea
+    id: request_description
+    attributes:
+      label: Describe your Learn request. If you want to add information to an existing page, please provide the link to it!
```
jwitz, line 36:
```suggestion
      label: Describe your Learn request.
      description: If you want Astronomer to add information to an existing page, please provide the link to it here.
```

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
+name: Request updates or new content in Learn
+description: New or updated content on an existing Airflow feature or use case
```
jwitz, line 2:
```suggestion
description: Request new or updated [Learn](https://docs.astronomer.io/learn) content on an Apache Airflow feature or use case.
```

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
+name: Request updates or new content in Learn
```
jwitz, line 1:
```suggestion
name: Request new Apache Airflow content in Learn
```
"Updates" could sound like we're asking a user for bug fix info

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+    attributes:
+      label: Content type
+      multiple: true
+      options:
+        - Get started
+        - Airflow concepts
+        - Airflow tutorials
+        - Integrations
+        - References
+        - I am not sure where this best fits
```
jwitz, line 19: Good call, we probably need a fallback option like this in the docs templates

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - type: textarea
+    id: request_description
+    attributes:
+      label: Describe your Learn request. If you want to add information to an existing page, please provide the link to it!
+    validations:
+      required: true
+  - type: textarea
+    id: target_user_description
+    attributes:
+      label: Describe who the content is intended for, e.g. Airflow beginners, expert users...?
```
jwitz, line 42: What decisions do you expect to make based on this answer? If you want to keep this, I think it's better to make it a list of options to select from.

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - type: input
+    id: due_date
+    attributes:
+      label: Desired publish date
+      description: For example, is there a particular release, relevant event, etc. that this should be coordinated with?
```
jwitz, line 47:
```suggestion
      description: Is there a particular release, relevant event, etc. that this should be coordinated with?
```

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - type: input
+    id: airflow_version
+    attributes:
+      label: Airflow Version
+      description: Airflow version in which the feature/behavior was added. Please provide for features which are upcoming.
```
jwitz, line 52:
```suggestion
      description: Airflow version in which the feature/behavior was added. Please provide for features which are not yet released.
```

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+      placeholder: ex. 2.6.1
```
jwitz, line 53:
```suggestion
      placeholder: For example: Airflow 2.6.1
```

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - type: input
+    id: sme
+    attributes:
+      label: SME / technical reviewer
```
jwitz, line 57:
```suggestion
      label: Reviewer
```

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+    validations:
+      required: true
+  - type: dropdown
+    id: request_type
+    attributes:
+      label: Request type
+      multiple: true
+      options:
+        - Additional information on a feature or behavior already described
+        - Add a previously not described Airflow feature or behavior to an existing page
```
jwitz, line 29:
```suggestion
        - New content for an undocumented or under-documented Airflow feature
```

jwitz (changes requested):
Adding a few thoughts here @TJaniF , lmk what you think! I'll take one more look before we merge

jwitz (changes requested):
Adding a few thoughts here @TJaniF , lmk what you think! I'll take one more look before we merge

.github/ISSUE_TEMPLATE/learn-bug-report.yaml
TJaniF: Good point yes, I'll remove it :)

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
TJaniF: Hmm yeah there is that danger when using the term "update". The reason for the 'updates' there is because we get a lot of issues that are "hey xyz is outdated/ there is a better way of doing this now/ xyz should be added to this list" etc. 

So neither a bug in the sense that something does not work but also not really new content, at least in my mind "new" means a new tutorial or at least a new section in a guide 🤔 

How about "Request new Apache Airflow content or content additions in Learn" ? 

cc: @kentdanas

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
TJaniF: I was thinking for example if someone creates an issue "How to use Airflow with FancyTool" if they select beginner it would be more suited for a shorter integration tutorial, if they select expert for a use-case based example with a repository. 🤔 

But I am not fully convinced the question is necessary, I think most of the time we decide how to integrate the content anyways so... I am ok with removing the question for brevity, not sure I would have thought of it if it had not been there in the old template.

TJaniF (commented):
Thank you for the review @jwitz ! :) 

I accepted most suggestions and left two comments. :)

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
kentdanas: Yeah....I agree most of the issues we get are "hey this part is outdated" or "hey can we add context on this additional thing here". However, the vast majority of these get raised to us in Slack and it's us that ends up creating the issue, so I wouldn't be too worried about people being confused here. Maybe we just keep it simple with Jake's suggestion, and can adjust later if needed.

.github/ISSUE_TEMPLATE/learn-new-content-template.yaml
```diff
@@ -0,0 +1,64 @@
...
+    id: due_date
+    attributes:
+      label: Desired publish date
+      description: Is there a particular release, relevant event, etc. that this should be coordinated with?
+  - type: input
+    id: airflow_version
+    attributes:
+      label: Airflow Version
+      description: Airflow version in which the feature/behavior was added. Please provide for features which are not yet released.
+      placeholder: For example, Airflow 2.6.1
```
jwitz, line 54:
```suggestion
```
Not sure we need a placeholder example at all tbh, knowing that this part of the form will only be used by experts that are in the know on future Airflow releases

jwitz (approved):
Just pushed a change to get the build to work, and now this LGTM!

TJaniF (commented):
Thank you @jwitz ! Excited to see this in action :)

PR #2326, opened 2023-05-23
Add docs for jenkins custom cicd template

pohek321:
Did this for a customer awhile back, forgot to add it here in the official docs.

### DIFF
--- a/astro/ci-cd-templates/jenkins.md
+++ b/astro/ci-cd-templates/jenkins.md
@@ -168,8 +168,7 @@ If your Astro project requires additional build-time arguments to build an image
 
 2. At the root of your Astro Git repository, add a [Jenkinsfile](https://www.jenkins.io/doc/book/pipeline/jenkinsfile/) that includes the following script:
 
-```text
-{`pipeline {
+<pre><code parentName="pre">{`pipeline {
         agent any
         stages {
             stage('Deploy to Astronomer') {
@@ -181,11 +180,11 @@ If your Astro project requires additional build-time arguments to build an image
                 steps {
                     checkout scm
                     sh '''
-                    export astro_id=$(date +%Y%m%d%H%M%S)
-                    docker build -f Dockerfile --progress=plain --build-arg <your-build-arguments> -t $astro_id .
+                    export astro_id=${siteVariables.jenkinsenv3}
+                    docker build -f Dockerfile --progress=plain --build-arg <your-build-arguments> -t ${siteVariables.jenkinsenv4} .
                     curl -LJO https://github.com/astronomer/astro-cli/releases/download/v${siteVariables.cliVersion}/astro_${siteVariables.cliVersion}_linux_amd64.tar.gz
                     tar -zxvf astro_${siteVariables.cliVersion}_linux_amd64.tar.gz astro && rm astro_${siteVariables.cliVersion}_linux_amd64.tar.gz
-                    ./astro deploy --image-name $astro_id
+                    ./astro deploy --image-name ${siteVariables.jenkinsenv4}
                     '''
                 }
             }
@@ -195,8 +194,7 @@ If your Astro project requires additional build-time arguments to build an image
                 cleanWs()
             }
         }
-    }`}
-    ```
+    }`}</code></pre>
 
     This `Jenkinsfile` triggers a code push to Astro every time a commit or pull request is merged to the `main` branch of your repository.
 
--- a/astro/ci-cd-templates/jenkins.md
+++ b/astro/ci-cd-templates/jenkins.md
@@ -145,6 +145,57 @@ To automate code deploys across multiple Deployments using [Jenkins](https://www
     This `Jenkinsfile` triggers a code push to an Astro Deployment every time a commit or pull request is merged to the `dev` or `main` branch of your repository.
 
 </TabItem>
+
+<TabItem value="custom">
+
+If your Astro project requires additional build-time arguments to build an image, you need to define these build arguments using Docker's [`build-push-action`](https://github.com/docker/build-push-action).
+
+#### Configuration requirements
+
+- An Astro project that requires additional build-time arguments to build the Runtime image.
+
+1. In your Jenkins pipeline configuration, add the following parameters:
+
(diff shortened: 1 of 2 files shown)


astro/ci-cd-templates/jenkins.md
```diff
@@ -145,6 +145,57 @@ To automate code deploys across multiple Deployments using [Jenkins](https://www
...
+
+    - `ASTRONOMER_KEY_ID`: Your Deployment API key ID
+    - `ASTRONOMER_KEY_SECRET`: Your Deployment API key secret
+    - `ASTRONOMER_DEPLOYMENT_ID`: The Deployment ID of your production deployment
+
+    Be sure to set the values for your API credentials as secret.
+
+2. At the root of your Astro Git repository, add a [Jenkinsfile](https://www.jenkins.io/doc/book/pipeline/jenkinsfile/) that includes the following script:
+
+    <pre><code parentName="pre">{`pipeline {
```
lzdanski, line 167:
```suggestion
```text
{`pipeline {
```

astro/ci-cd-templates/jenkins.md
```diff
@@ -145,6 +145,57 @@ To automate code deploys across multiple Deployments using [Jenkins](https://www
...
+                    '''
+                }
+            }
+        }
+        post {
+            always {
+                cleanWs()
+            }
+        }
+    }`}</code></pre>
```
lzdanski, line 193:
```suggestion
    }`}
    ```
```

lzdanski (changes requested):
These changes look good! I saw one place where we might want to add a sentence about where or how users can find their Deployment information and there's a markdown formatting syntax change for the codeblock that's needed.

astro/ci-cd-templates/jenkins.md
```diff
@@ -145,6 +145,57 @@ To automate code deploys across multiple Deployments using [Jenkins](https://www
...
+
+- An Astro project that requires additional build-time arguments to build the Runtime image.
+
+1. In your Jenkins pipeline configuration, add the following parameters:
+
+    - `ASTRONOMER_KEY_ID`: Your Deployment API key ID
+    - `ASTRONOMER_KEY_SECRET`: Your Deployment API key secret
+    - `ASTRONOMER_DEPLOYMENT_ID`: The Deployment ID of your production deployment
+
+    Be sure to set the values for your API credentials as secret.
```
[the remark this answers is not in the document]
lzdanski: Would the single branch and custom use the same method to find the parameters? Is it just the cli - [`astro deployment inspect`](https://docs.astronomer.io/astro/cli/astro-deployment-inspect) or go to the deployment in the Cloud UI?

astro/ci-cd-templates/jenkins.md
pohek321: These changes broke the CI/CD checks -- not sure what that `<pre>` tag does @lzdanski, maybe you could elaborate?

lzdanski: Absolutely! So with markdown syntax, it accepts both markdown and HTML markup. The declaration `<pre><code>` is a way to declare to the framework that we're using HTML for that section, not markdown. However, the build errors are being reported from line 241 in that file, which isn't from the changes you made. It looks like it's happy with our syntax, but not happy going back to using HTML for codeblocks. 

Can I tinker with it and push a change back to your branch?
Screenshot 2023-05-23 at 4 20 43 PM

astro/ci-cd-templates/jenkins.md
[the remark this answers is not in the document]
pohek321: @lzdanski how to find these values is already defined in the same page under the [prerequisite section](https://docs.astronomer.io/astro/ci-cd-templates/jenkins?tab=multibranch#prerequisites)

jwitz:
@lzdanski Sorry this is a niche part of the site that I'd hoped you never had to encounter. Jenkins is one of the few pipeline tools left that requires users to `curl` specific versions of the CLI in the pipeline, rather than always using latest. We've had to use sitewide environment variables to ensure that these pipelines are always updated when we release a new version of the Astro CLI.

A negative side effect of this implementation is that any raw code starting with `$` needs to be formatted so that Docusaurus doesn't try to interpret it as a non-existent environment variable. You'll see I made a few additional variables to reformat what was before a raw `$`-led value. @pohek321 Can you check the preview site to make sure that all of the indentation appears as you'd hoped?

astro/ci-cd-templates/jenkins.md
```diff
@@ -145,6 +147,59 @@ To automate code deploys across multiple Deployments using [Jenkins](https://www
...
+            }
+        }
+        post {
+            always {
+                cleanWs()
+            }
+        }
+    }`}</code></pre>
+
+    This `Jenkinsfile` triggers a code push to Astro every time a commit or pull request is merged to the `main` branch of your repository.
```
lzdanski, line 199:
```suggestion
This `Jenkinsfile` triggers a code push to Astro every time a commit or pull request is merged to the `main` branch of your repository.
```

This is going to sound strange, but did you indent with tabs or spaces? Sometimes Markdown syntax has tab = plain text and 4 spaces = code formatting. I haven't seen it pop up in a while, but it looks like that's happening here. We can either remove the indent or you can try using a tab (which for some reason Github suggestions won't allow me to do).

pohek321: Not sure, I'll commit up a change with this commit suggestion though

pohek321:
> @lzdanski Sorry this is a niche part of the site that I'd hoped you never had to encounter. Jenkins is one of the few pipeline tools left that requires users to `curl` specific versions of the CLI in the pipeline, rather than always using latest. We've had to use sitewide environment variables to ensure that these pipelines are always updated when we release a new version of the Astro CLI.
> 
> A negative side effect of this implementation is that any raw code starting with `$` needs to be formatted so that Docusaurus doesn't try to interpret it as a non-existent environment variable. You'll see I made a few additional variables to reformat what was before a raw `$`-led value. @pohek321 Can you check the preview site to make sure that all of the indentation appears as you'd hoped?

@jwitz yeah, it looks good!

jwitz (approved):

PR #2364, opened 2023-06-01
Remove replication example

jedcunningham:
We don't want to suggest folks use in-cluster dbs, much less with replication.

### DIFF
--- a/software/install-azure-standard.md
+++ b/software/install-azure-standard.md
@@ -249,7 +249,7 @@ kubectl create secret generic astronomer-bootstrap \
 > **Note:** You must URL encode any special characters in your Postgres password.
 
 A few additional configuration notes:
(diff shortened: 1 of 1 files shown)


kaxil (approved):

software/install-azure-standard.md
```diff
@@ -249,7 +249,7 @@ kubectl create secret generic astronomer-bootstrap \
 > **Note:** You must URL encode any special characters in your Postgres password.
 
 A few additional configuration notes:
-- If you provision an external database, `postgresqlEnabled` should be set to `false` in Step 8.
+- If you provision an external database, `postgresqlEnabled` should be set to `false` in Step 8. The in-cluster database should only be used for development or proof of concept installations. All meaningful installations should use an external database.
```
jwitz, line 252:
```suggestion
- If you provision an external database, `postgresqlEnabled` should be set to `false` in Step 8. The in-cluster database should be used only for development or proof-of-concept installations. All production installations should use an external database.
```

jwitz (approved):

PR #2374, opened 2023-06-02
New Examples section with local dev examples

kentdanas:
Creates a new Examples section in Learn with our first subsection- local dev containing 2 ADE modules. Fast-follow will be other sections and docs.


sidebarsLearn.js
```diff
@@ -167,5 +167,29 @@ module.exports = {
...
+      },
+      items: [
+        {
+          type: 'category',
+          label: 'Local Development',
+          link: {
+            type: 'generated-index',
+            title: 'Local Development',
+            description: 'See examples of setting up local development environments for Airflow.',
+          },
```
jwitz, line 186:
```suggestion
```

learn/pycharm-local-dev.md
```diff
@@ -0,0 +1,67 @@
...
+description: "Integrate the Astro CLI with PyCharm for local development."
+id: pycharm-local-dev
+sidebar_label: "PyCharm"
+---
+
+This example shows how to set up [PyCharm]((https://www.jetbrains.com/pycharm/)) for local development with Airflow and the [Astro CLI](https://docs.astronomer.io/astro/cli/overview).
+
+## Before you start
+
+Before using this example, make sure you have:
```
jwitz, line 12:
```suggestion
Before trying this example, make sure you have:
```

learn/pycharm-local-dev.md
```diff
@@ -0,0 +1,67 @@
...
+
+Here is an example of deprecated imports:
+
+![Deprecated Imports](/img/examples/pycharm_local_dev_deprecated_import.png)
+
+Here is an example of an unused import:
+
+![Unused Imports](/img/examples/pycharm_local_dev_unused_import.png)
+
+Here is an example error syntax highlighting:
```
jwitz, line 43:
```suggestion
Like with other Python projects, PyCharm will highlight syntax errors in your Airflow code:
```

learn/pycharm-local-dev.md
```diff
@@ -0,0 +1,67 @@
...
+
+![Syntax Highlighting](/img/examples/pycharm_local_dev_syntax_highlighting.png)
+
+Here is an example of code autocompletion and built in definitions:
```
jwitz, line 47:
```suggestion
Lastly, here is an example of PyCharm autocompleting code and showing built-in definitions:
```

learn/pycharm-local-dev.md
```diff
@@ -0,0 +1,67 @@
...
+
+From the Services pane, start Docker by clicking the green play button and you’ll see the same containers appear as when you run `docker ps`:
+
+![Containers](/img/examples/pycharm_local_dev_containers.png)
+
+You can view logs for the containers by clicking on `/scheduler`, `/triggerer`, `/webserver`, or `/airflow-dev_2cd823-postgres` (this may be a different string based on what the name of the parent directory for your project is):
+
+![Logs](/img/examples/pycharm_local_dev_logs.png)
+
+You can run Airflow CLI commands by right clicking on `/scheduler` and selecting `Create Terminal` to bash into the container:
```
jwitz, line 65:
```suggestion
Run Airflow CLI commands by right clicking on `/scheduler` and selecting **Create Terminal** to bash into the container:
```

sidebarsLearn.js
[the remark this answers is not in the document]
jwitz: I don't think it's worth having too many menus for people to click through on the sidebar. I'm hoping we'll have a better search system for these soon!

static/img/examples/pycharm_local_dev_interpreter.png
jwitz, line 1: Could use a crop on the bottom here

jwitz (changes requested):
This is a great start @kentdanas ! I think the style and structure changes are spot on. Only requesting changes based on the content in the PyCharm doc, but also leaving a few general notes:

- I think it's still okay to use numbered steps, just as long as we don't have scores of them across multiple topics
- I like the screenshots, but we should conservative with the size and frequency of images to avoid making articles too long. I marked some places where we might want to crop or resize images, but I think there are more optimizations to be made. We should also consider rallying around a specific tool/ style for screenshot taking - on the core Astro docs we've used orange for our borders/ numbers.
- Loving the minimalism but still think we could add another sentence or two at the start of topics to communicate the "why" of these setups.
- Until we have a better search system, we should be generous in our "See also" sections to help link related content, especially for people who have come from Google.

learn/pycharm-local-dev.md
```diff
@@ -0,0 +1,67 @@
+---
+title: "Develop locally with PyCharm"
+description: "Integrate the Astro CLI with PyCharm for local development."
+id: pycharm-local-dev
+sidebar_label: "PyCharm"
+---
+
+This example shows how to set up [PyCharm]((https://www.jetbrains.com/pycharm/)) for local development with Airflow and the [Astro CLI](https://docs.astronomer.io/astro/cli/overview).
+
```
TJaniF, line 9: This might be out of scope but do we want to have 1-2 sentences why doing this is beneficial? Or maybe it is self explanatory but something like:

"By setting up your local dev environment you will be able to iterate more quickly when developing Airflow DAGs for example by [debugging interactively with dag.test()](https://docs.astronomer.io/learn/testing-airflow#debug-interactively-with-dagtest)."

learn/pycharm-local-dev.md
```diff
@@ -0,0 +1,67 @@
+---
+title: "Develop locally with PyCharm"
+description: "Integrate the Astro CLI with PyCharm for local development."
+id: pycharm-local-dev
+sidebar_label: "PyCharm"
+---
+
+This example shows how to set up [PyCharm]((https://www.jetbrains.com/pycharm/)) for local development with Airflow and the [Astro CLI](https://docs.astronomer.io/astro/cli/overview).
```
TJaniF, line 8:
```suggestion
This example shows how to set up [PyCharm](https://www.jetbrains.com/pycharm/) for local development with Airflow and the [Astro CLI](https://docs.astronomer.io/astro/cli/overview).
```

learn/pycharm-local-dev.md
```diff
@@ -0,0 +1,67 @@
...
+## Before you start
+
+Before using this example, make sure you have:
+
+- [PyCharm](https://www.jetbrains.com/pycharm/)
+- The [Astro CLI](https://docs.astronomer.io/astro/cli/install-cli) installed and a project running locally
+
+## Configure the Python interpreter
+
+In your PyCharm preferences go to **Build, Execution, Deployment** >> **Docker** and set up the following:
```
TJaniF: This probably is because I have never used pycharm but I am a little confused what I am supposed to set. 
Is it
"select the correct Docker daemon connection for your set up and add a path mapping for `/Users` to be able to mount files from this folder"

not sure if that setting is correct, I am assuming by mapping Users to Users you are able to create cp statements for files in Users, you are not mounting the whole User directory....? 👀

learn/pycharm-local-dev.md
```diff
@@ -0,0 +1,67 @@
...
+
+![Docker settings](/img/examples/pycharm_local_dev_docker_settings.png)
+
+Go to **Project: your-project-name** >> **Python Interpreter** and click the settings gear icon next to Python Interpreter, then click **Add**:
+
+![Interpreter settings](/img/examples/pycharm_local_dev_interpreter.png)
+
+In the window that appears, click on **Docker** in the left pane, and in the **Image name** field select the Docker image that your Airflow environment is running. Then click **OK**.
```
TJaniF: ! does this allow you to use the python env from the docker image in the IDE when using dag.test? So you would not need to locally install the provider packages but can still use IDE debugging? 🤔

learn/pycharm-local-dev.md
```diff
@@ -0,0 +1,67 @@
...
+
+![Image settings](/img/examples/pycharm_local_dev_docker_image.png)
+
+## Write Airflow code with PyCharm
```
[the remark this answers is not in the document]
kentdanas: This section is second because you need to configure the Python interpreter before you can do this. Will update the descriptions to make that more clear.

TJaniF (commented):
This is really cool. I think these will be very popular! 😄 

I added some comments. One overall thought I had is that we could mention IDE debugging, either with another screenshot of by linking out to that part of the debugging guide? :)

sidebarsLearn.js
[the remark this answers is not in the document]
kentdanas: Yeah good call. I definitely want a `Use cases` section within Examples since those will be a full section of their own and we want to have a fair amount of them. But for other things like this I would be okay keeping them at the top level instead of having other subsections, at least for now when there won't be too many. 

Do you think we should remove the `Local development` section for now and see how it goes without it? I don't expect we'll have more than these two docs in this section for now...

learn/pycharm-local-dev.md
```diff
@@ -0,0 +1,67 @@
...
+
+Here is an example error syntax highlighting:
+
+![Syntax Highlighting](/img/examples/pycharm_local_dev_syntax_highlighting.png)
+
+Here is an example of code autocompletion and built in definitions:
+
+![Code Autocompletion](/img/examples/pycharm_local_dev_autocomplete.png)
+
+## Manage Docker containers with PyCharm
```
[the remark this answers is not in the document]
kentdanas: Hmm....I'm on the fence about this. Generally the flow is you have to set up an interpreter to do anything Python related in PyCharm, then you can write code/do other things. For this example, we use a docker-specific interpreter so that you have the option of interacting with docker containers, but even if you were to remove this, an interpreter is still necessary for the writing code part.

So the flow we have write now is configure an interpreter (and choose one that allows you to do "other things") -> write code -> do "other things" (manage your docker containers). I would lean towards keeping these sections and I'll add more explanation around it, but definitely push back if you think it still doesn't work!

learn/pycharm-local-dev.md
kentdanas: Hmm...I believe it would? It basically allows you to exec into the containers just like you would from your terminal, but directly from the IDE. So I guess you could probably do that and then run your Python file from there? Not sure whether you can run the file from PyCharm and have it run within Docker?

learn/pycharm-local-dev.md
kentdanas: Your text is correct, although the /Users part is prepopulated, but can be updated. I updated this section to link to the PyCharm docs instead of the screenshot, since they go in more depth and it might vary for different setups.

learn/vscode-local-dev.md
```diff
@@ -0,0 +1,44 @@
...
+
+1. Open the folder containing your Astro Project in VS Code. In the bottom left corner of your VS Code window, click the green containers icon and select `Open Folder in Container...`
+    
+    ![Open Folder](/img/examples/vscode_local_dev_open_folder.png)
+    
+2. The Finder/File Explorer will open prompting you to select the project folder. Ensure that your Astro project is selected and click **Open**, then **From 'Dockerfile'**.
+    
+3. A new VS Code window appears. Note in the lower left corner of the window that your IDE is now pointing to that running Docker container.
+    
+    ![Container](/img/examples/vscode_local_dev_container.png)
```
kentdanas, line 31: I feel like we could probably take this screenshot out as well and just keep the text....thoughts?

jwitz (approved):
This is looking great now @kentdanas ! I tried playing around with the headers but tbh long as the content stays compact like this, I think the more simple, straightforward structure makes sense (e.g. not having nested topics, rigid structures, etc)

learn/pycharm-local-dev.md
TJaniF: That looks really useful, I had no idea, thank you! :)

PR #2387, opened 2023-06-07
Add missing supported GCP nodes

jwitz:
Resolves https://github.com/astronomer/docs/issues/2077

I updated this based on a figure provided by @frank-ye : 
```
 daemonset:
      cpu: 2 # 1.766 vCPU plus a buffer, calculated by summing cluster daemonset container CPU limits - do not change units
      ram: 1500 # 1204 MiB plus a buffer, calculated by summing cluster daemonset container memory limits - do not change units
```

I think we need to update all other instance types as well, but confirming with Frank before I do that


jwitz:
@frank-ye confirmed via Slack that the existing node types needed to have their figures updated as well.

PR #2391, opened 2023-06-07
Update kubernetespodoperator.md

RNHTTR:
`resources` is actually `container_resources`

https://github.com/apache/airflow/blob/providers-cncf-kubernetes/7.0.0/airflow/providers/cncf/kubernetes/operators/pod.py#L182


jwitz (approved):
Thanks @RNHTTR !

Issue #2401, opened 2023-06-08
Remove `--dag-deploy enable` from "Run your First DAG"


paolaperaza:
[Not released yet] 

### Context

We know that for trial users, installing Docker and needing Docker to deploy to Astro for the first time is a point of friction. We can make this better by automatically enabling DAG-only Deploys for trial accounts.

PR here from @melugoyal: https://github.com/astronomer/astro/pull/14045

### Ask

Here: https://docs.astronomer.io/astro/create-first-DAG#step-3-deploy-example-dags-to-your-astro-deployment

Remove Step 3 -- 

<img width="748" alt="image" src="https://github.com/astronomer/docs/assets/31111752/e01a488f-7345-4c7e-bfb3-d234280b4612">

PR #2426, opened 2023-06-13
Astro 6/13 release notes

jwitz:
Tagging PMs from there respective areas to confirm that no changes are missing here.

Resolves https://github.com/astronomer/docs/issues/2401

### DIFF
--- a/astro/release-notes.md
+++ b/astro/release-notes.md
@@ -31,6 +31,7 @@ Use the new **Billing** page in the Cloud UI to see both high-level and detailed
 
 ### Additional improvements
 
+- Trial Deployments now have [DAG-only deploys](deploy-code.md#deploy-dags-only) enabled by default.
 - The Cloud UI now shows your **Organization Short Name** and **Astro SAML Connection Name** in the Cloud UI.
 - You can now view mapped tasks from the **DAGs** page in the Cloud UI.
 
--- a/astro/release-notes.md
+++ b/astro/release-notes.md
@@ -25,7 +25,9 @@ Astronomer is committed to continuous delivery of both features and bug fixes to
 
 ## June 13, 2023
 
-### Manage 
+### Manage billing and track usage for Astro Hosted
+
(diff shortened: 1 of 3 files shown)


chomipi88:
One addition that we are rolling out this week:

1. Astronomer will be enforcing IMDSv2 only to strengthen the security posture of new AWS Data Planes as of today. Existing AWS clusters will have the enforcement applied within the next 2 weeks.

chomipi88 (commented):
see comment

jwitz:
@chomipi88 My instinct is to treat that as a standard upgrade which falls under our security policy, so I wouldn't add it to release notes unless it changes behavior on the platform for users/ admins

cmarteepants:
> @chomipi88 My instinct is to treat that as a standard upgrade which falls under our security policy, so I wouldn't add it to release notes unless it changes behavior on the platform for users/ admins

It has the potential to break DAGs. At minimum, this needs to be be added to release notes.

astro/release-notes.md
```diff
@@ -23,6 +23,40 @@ Astronomer is committed to continuous delivery of both features and bug fixes to
...
+
+:::warning Breaking change
+
+If your DAGs assume IAM roles to directly access metadata on your cluster using IMDSv1, this change can result in DAG run failures. Upgrade your DAGs to use IMDSv2 for all cluster metadata requests.
+
+See [Use IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) for more information.
+
+:::
+
+Astronomer now enforces IMDSv2 on all AWS clusters. Any requests for resources on your clusters are now session-based and muse include a token in the request body.
```
jwitz, line 48:
```suggestion
Astronomer now enforces IMDSv2 on all AWS clusters. Any requests for resources on your clusters are now session-based and muse include a token in the request body.

```

chomipi88:
@jwitz lgtm

PR #2437, opened 2023-06-15
[Tiny PR] Use case dbt title change

TJaniF:
Removing the "create" from the current use case title because there are no steps anymore in the use case.

Happy to take other suggestions for the title of course! :)

### DIFF
--- a/learn/use-case-airflow-dbt.md
+++ b/learn/use-case-airflow-dbt.md
@@ -1,5 +1,5 @@
 ---
-title: "Create an ELT pipeline with Airflow, dbt Core, Cosmos, and the Astro Python SDK"
+title: "ELT pipeline with Airflow, dbt Core, Cosmos, and the Astro Python SDK"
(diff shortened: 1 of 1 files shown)


learn/use-case-airflow-dbt.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Create an ELT pipeline with Airflow, dbt Core, Cosmos, and the Astro Python SDK"
+title: "ELT pipeline with Airflow, dbt Core, Cosmos, and the Astro Python SDK"
```
jwitz, line 2:
```suggestion
title: "ELT pipelines with Airflow, dbt Core, Cosmos, and the Astro Python SDK"
```
I know it's only one that we're running in the example, but this looks a little nicer as a title

jwitz (approved):

learn/use-case-airflow-dbt.md
kentdanas: Yeah I like the plural!

kentdanas (approved):

PR #2482, opened 2023-06-26
Add event promo banner component to Docs

### LINKED ISSUE #1486: Create toggleable new feature banner for home pages
We should have a CSS element that's similar to `AstroCard` but can be hidden by a user when they click an `x` button. Instead of having additional large buttons, it would only have plain text and hyperlinked text. 

Screen Shot 2022-11-15 at 3 04 55 PM

This would be used by the docs team to highlight a significant new change and would not always be active on the site.

katiebrady-astro:


paolaperaza:
Thank you for this, @katiebrady-astro! I definitely think we need this. Before we merge, 3 quick things:

- The colors. As part of this PR, let's actually go ahead and update that purple color in the nav bar to the same blue as our homepage graphic. And then try two banners:
     - A purple banner with white CTA potentially
     - The blue gradient banner with orange CTA (like the homepage)
- Goes without saying but let's make sure that this is dismissable
- I'm ok with this banner even if it's not our "new feature" banner. @jwitz I think you had some thoughts here but but let's just move forward with this for now

(And Katie let me know if you have issues with the colors. For reference, here's the treatment we were gonna do on Academy:

image

docusaurus.config.js
```diff
@@ -108,6 +108,11 @@ module.exports = {
       successMessage: "Success! ✓",
       errorMessage: "Sorry, there was issue sending your email. Please try again.",
     },
+    promoBanner: {
+      text: "Don't miss the BIGGEST Airflow event of the year: Airflow Summit 2023 - Sept 19-23",
```
jwitz, line 112:
```suggestion
      text: "Don't miss the biggest Airflow event of the year: Airflow Summit 2023, Sept 19-23",
```
Can I executive override some of this copy haha

jwitz (approved):
Yeah with Paola's comments this LGTM! As long as it doesn't close the current issue though, because we still need a feature banner card

docusaurus.config.js
katiebrady-astro: Totally! haha

katiebrady-astro:
Thanks for the feedback @jwitz and @paolaperaza! I'll get the colors + design updated before merging this and will reopen the original issue (since I think it's going to auto-close with this branch name 😅) to build a feature card banner too!

katiebrady-astro:
Hey @paolaperaza and @jwitz here's a preview of the requested color schemes! Let me know which we like and I'll get that commited to the branch. Thanks!
Screenshot 2023-06-28 at 4 25 48 PM
Screenshot 2023-06-28 at 4 26 10 PM

paolaperaza (approved):

Issue #2416, opened 2023-06-12
docker-compose is being deprecated in favor of 'docker compose'


danielhoherd:
### Issue Type

Missing or misleading content

### Links to Affected Docs

Everywhere we suggest people run the `docker-compose` command: https://github.com/search?q=repo%3Aastronomer%2Fdocs%20docker-compose&type=code

### Describe the Issue and Proposed Fix

We should replace all `docker-compose` commands with `docker compose`

### Additional Notes

Docker is deprecating the standalone `docker-compose` python package in favor of using a `docker` plugin called `compose`. `docker-compose` will not be supported upstream beyond June 2023.

- https://docs.docker.com/compose/migrate deprecation announcement
- https://github.com/astronomer/issues/issues/5656 related ticket where we are making code changes in gen1 repos
- https://pypi.org/project/docker-compose/#history docker-compose hasn't been updated in over 2 years

PR #2487, opened 2023-06-27
Replace all instances of docker-compose

jwitz:
Resolves https://github.com/astronomer/docs/issues/2416

@danielhoherd I didn't replace instances of `docker-compose-override.yml` because that didn't seem to be covered in the deprecaction, but lmk if those need to change as well.

### DIFF
--- a/archived_software_docs/version-0.16/kubepodoperator-local.md
+++ b/archived_software_docs/version-0.16/kubepodoperator-local.md
@@ -9,7 +9,7 @@ description: Test the KubernetesPodOperator on your local machine.
 
 ### Windows and Mac
 
-The latest version of Docker for Windows and Mac comes with the ability to run a single node Kubernetes cluster on your local machine. If you are on Windows, follow [this guide](https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly) for setting up Docker for Windows 10 and WSL (you don’t need to install docker-compose if you don’t want to).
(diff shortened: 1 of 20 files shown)


archived_software_docs/version-0.16/kubepodoperator-local.md
```diff
@@ -9,7 +9,7 @@ description: Test the KubernetesPodOperator on your local machine.
 
 ### Windows and Mac
 
-The latest version of Docker for Windows and Mac comes with the ability to run a single node Kubernetes cluster on your local machine. If you are on Windows, follow [this guide](https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly) for setting up Docker for Windows 10 and WSL (you don’t need to install docker-compose if you don’t want to).
+The latest version of Docker for Windows and Mac comes with the ability to run a single node Kubernetes cluster on your local machine. If you are on Windows, follow [this guide](https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly) for setting up Docker for Windows 10 and WSL (you don’t need to install docker compose if you don’t want to).
```
danielhoherd, line 12: I'm not a windows user, but it looks like the linked post is superseded by the WSL 2 setup guide. See the page for details: https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly

> Update in 2020: Now that Microsoft has released the Spring 2020 Windows update we have access to WSL 2 on all editions of Windows 10

AFAIK `docker compose` functionality is included in Docker for windows and Mac, so we should avoid mentioning `docker-compose` installation at all.

danielhoherd:
@jwitz Good call on not replacing the filenames. Those are staying the same. Only the command `docker-compose` is changing.

archived_software_docs/version-0.16/kubepodoperator-local.md
jwitz: I did a catch-all find and replace that affected some archived versions, I don't think this is referenced in our published docs anymore

PR #2491, opened 2023-06-28
map correct key in docs for registry backend 

pgvishnuram:
In our docs we have mapped the key to a non valid value it should replicate the one created from secret.


jwitz (approved):
Thank you @pgvishnuram !

PR #2493, opened 2023-06-28
Small PR: Clarify wording around consumer DAGs in Datasets guide

TJaniF:
Noticed this when reading the guide yesterday and thought it was a little confusing since tasks can't directly consume datasets (yet). Open to rephrasing of course :)

### DIFF
--- a/learn/airflow-datasets.md
+++ b/learn/airflow-datasets.md
@@ -40,13 +40,13 @@ When you define a task's `outlets` parameter, Airflow labels the task as a produ
 
 <CodeBlock language="python">{dataset_producer}</CodeBlock>
 
-After a dataset is defined in one or more producer tasks, consumer DAGs in your Airflow environment listen to the producer tasks and run whenever the task completes, rather than running on a time-based schedule. For example, if you have a DAG that should run when the `INSTRUCTIONS` and `INFO` datasets are updated, you define the DAG's schedule using the names of the datasets.
+After a dataset is defined in one or more producer tasks, consumer DAGs in your Airflow environment listen to the producer tasks and run whenever a dataset is produced to, rather than running on a time-based schedule. For example, if you have a DAG that should run when the `INSTRUCTIONS` and `INFO` datasets are updated, you define the DAG's schedule using the names of those two datasets.
 
-Any task that is scheduled with a dataset is considered a consumer task even if that task doesn't consume the referenced dataset. In other words, it is up to you as the DAG author to correctly reference and use the dataset that the consumer DAG is scheduled on.
+Any DAG that is scheduled with a dataset is considered a consumer DAG even if that DAG doesn't actually perform actions on the referenced dataset. In other words, it is up to you as the DAG author to correctly reference and use datasets.
 
 <CodeBlock language="python">{dataset_consumer}</CodeBlock>
 
-Any number of datasets can be provided to the `schedule` parameter as a list. The DAG is triggered after all of the datasets have received at least one update due to a producing task completing successfully. 
+Any number of datasets can be provided to the `schedule` parameter as a list. The DAG is triggered after all of the datasets have received at least one update due to a producing task completing successfully.
 
 When you work with datasets, keep the following considerations in mind:
 


learn/airflow-datasets.md
```diff
@@ -40,13 +40,13 @@ When you define a task's `outlets` parameter, Airflow labels the task as a produ
...
+After a dataset is defined in one or more producer tasks, consumer DAGs in your Airflow environment listen to the producer tasks and run whenever a dataset is produced to, rather than running on a time-based schedule. For example, if you have a DAG that should run when the `INSTRUCTIONS` and `INFO` datasets are updated, you define the DAG's schedule using the names of those two datasets.
 
-Any task that is scheduled with a dataset is considered a consumer task even if that task doesn't consume the referenced dataset. In other words, it is up to you as the DAG author to correctly reference and use the dataset that the consumer DAG is scheduled on.
+Any DAG that is scheduled with a dataset is considered a consumer DAG even if that DAG doesn't actually perform actions on the referenced dataset. In other words, it is up to you as the DAG author to correctly reference and use datasets.
```
jwitz, line 45:
```suggestion
Any DAG that is scheduled with a dataset is considered a consumer DAG even if that DAG doesn't actually access the referenced dataset. In other words, it's up to you as the DAG author to correctly reference and use datasets.
```

learn/airflow-datasets.md
```diff
@@ -40,13 +40,13 @@ When you define a task's `outlets` parameter, Airflow labels the task as a produ
 
 <CodeBlock language="python">{dataset_producer}</CodeBlock>
 
-After a dataset is defined in one or more producer tasks, consumer DAGs in your Airflow environment listen to the producer tasks and run whenever the task completes, rather than running on a time-based schedule. For example, if you have a DAG that should run when the `INSTRUCTIONS` and `INFO` datasets are updated, you define the DAG's schedule using the names of the datasets.
+After a dataset is defined in one or more producer tasks, consumer DAGs in your Airflow environment listen to the producer tasks and run whenever a dataset is produced to, rather than running on a time-based schedule. For example, if you have a DAG that should run when the `INSTRUCTIONS` and `INFO` datasets are updated, you define the DAG's schedule using the names of those two datasets.
```
jwitz, line 43:
```suggestion
A consumer DAG listens to a producer task and runs whenever the producer task completes, rather than running on a time-based schedule. For example, if you have a DAG that should run when the `INSTRUCTIONS` and `INFO` datasets are updated, you define the DAG's schedule using the names of those two datasets.
```
- "is produced to" didn't sound quite right
- tried removing even more language to make this more clear

jwitz (approved):

learn/airflow-datasets.md
TJaniF: I like having less words here but I wonder if it might lead to a misunderstanding, since the consumer DAG is not listening to a specific task but rather to a dataset and will run whenever that dataset is updated due to a producer task completing successfully. These updates can happen from different tasks, so the mapping is not direct. Maybe:

> A consumer DAG listens to a dataset and runs whenever the dataset is updated by a producer task completing successfully, rather than... 

? :) 

And yes I agree, produce to sounds weird haha, I'll try to use updated for datasets when possible. :)

kentdanas (approved):

learn/airflow-datasets.md
TJaniF: I like that :) 
@jwitz would that work for you as well?

jwitz: Don't we say earlier tho that the task doesn't have to explicitly update the dataset? This implies that the task takes action on the dataset

TJaniF: I think it depends on what you refer to as the dataset (naming again) 😅 

`Dataset("some_uri")` is the Airflow class that is being produced to / updated by the Airflow task. This action has to happen for the downstream DAG to run. This is separate from any actual dataset as in file or table that may or may not exist and be updated which is where the URI in the Dataset class points to. 

I am not sure how to best take it apart because it is the same word for two separate things 🤔

PR #2495, opened 2023-06-28
Update astro-deployment-connection-create.md

fritz-astronomer:
Was incorrectly referring to `list`


jwitz (approved):
😰 Thanks @fritz-astronomer

PR #2498, opened 2023-06-28
Add pytest process for outputting coverage or other artifacts

stuart23:
There are no clear instructions on how to do this. See https://github.com/astronomer/astro-cli/issues/1142

### DIFF
--- a/astro/cli/astro-dev-pytest.md
+++ b/astro/cli/astro-dev-pytest.md
@@ -46,7 +46,6 @@ astro dev pytest --args "–-cov-config path"
 astro dev pytest --args "--cov --cov-report xml:include/coverage.xml"
 ```
 
-
 ## Related Commands
 
 - [`astro dev init`](cli/astro-dev-init.md)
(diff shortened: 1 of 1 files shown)


frankcash:
https://docs.pytest.org/en/7.1.x/how-to/output.html#creating-junitxml-format-files

frankcash (approved):

astro/cli/astro-dev-pytest.md
```diff
@@ -8,6 +8,8 @@ hide_table_of_contents: true
 
 Run unit tests for your data pipelines on Astro with `pytest`, a testing framework for Python. When you run this command, the Astro CLI creates a local Python environment that includes your DAG code, dependencies, and Astro Runtime Docker image. The CLI then runs any pytests in the `tests` directory of your Astro project and shows you the results of those tests in your terminal.
 
+This command runs `pytest` in a container. If your test is generating artifacts that you need to use (e.g. code coverage reports), these can be output to the `include` directory in the container so they can be accessed after the test has finished.
```
jwitz, line 11:
```suggestion
The command runs `pytest` in a container. If your test generates artifacts, such as code coverage reports, you can output the artifacts to the `include` folder of your Astro project so they can be accessed after the test has finished.
```
Is this right? If not, how would you access the `include` directory if it was in a container?

jwitz (changes requested):
I made a suggestion that might include a change in technical meaning, so requesting changes to make sure that's all good before merging!

astro/cli/astro-dev-pytest.md
stuart23: Exactly! The include dir is shared inside and outside the container and is therefore the only place we can pass stuff out of the test environment. Your changes look good

jwitz (approved):

PR #2499, opened 2023-06-28
Update astro-dev-pytest.md

frankcash:
## Description

Includes example command to export pytest xml report file to local host


frankcash:
https://github.com/astronomer/docs/pull/2498 is more complete than my PR

PR #2505, opened 2023-06-30
Remove note about software worker termination grace period

kentdanas:
This guide needs a broader update....but removing this now to avoid confusion since this isn't a feature for Astro.


TJaniF (approved):
LGTM!

jwitz (approved):

PR #2506, opened 2023-06-30
Improve Software Airgap Instructions and Examples

rob-1126:
The previous instructions did not specify how to source the nginx image and left it to customers to source their own copy. This pdated yaml document contains configuration changes to make it work with one of the images we ship by default and is a drop-in replacement functionality wise for customers. It also adds a NetworkPolicy that allows access to that container since we are denying access by default to containers that don't have a network policy defined.

### DIFF
--- a/software/install-airgapped.md
+++ b/software/install-airgapped.md
@@ -201,7 +201,7 @@ To complete this setup:
         spec:
           containers:
           - name: astronomer-releases
-            image: <use same image as es-ingress-controller>
+            image: ap-nginx-es
             resources:
               requests:
                 memory: "32Mi"
--- a/software/install-airgapped.md
+++ b/software/install-airgapped.md
@@ -201,7 +201,7 @@ To complete this setup:
         spec:
           containers:
           - name: astronomer-releases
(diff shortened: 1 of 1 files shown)


software/install-airgapped.md
```diff
@@ -201,7 +201,7 @@ To complete this setup:
         spec:
           containers:
           - name: astronomer-releases
-            image: 012345678910.dkr.ecr.us-east-1.amazonaws.com/nginx:stable # Replace with own image
+            image: <use same image as es-ingress-controller>
```
jwitz, line 204:
```suggestion
            image: <your-es-ingress-controller-image>
```

jwitz (approved):
LGTM! One minor edit to the placeholder text

rob-1126:
Hrmm, it actually usually cannot match their ingress controller image and
that's a different image than we use as our ingress controller we ship TWO
variants if nginx ingress cus reasons.

--Rob

jwitz:
@rob-1126 Sorry not sure I understand. I thought my edit wan't changing the technical meaning of the original change. Should we be saying `astronomer-default-es-ingress-controller-image` or something similar instead?

jwitz:
@rob-1126 Updated my suggestion, lmk if that works!

PR #2511, opened 2023-07-03
Tiny update vscode-local-dev.md

jwitz:
I updated step 4 to only have the confirmation step, then moved some of the content to a new paragraph that describes what the integration does. This might help users who only want to see the "end state" for the setup and don't want to go through the steps for that.


TJaniF (approved):
Makes sense and LGTM :)

Issue #2521, opened 2023-07-05
Deleted branches are referenced in "Software" docs


danielhoherd:
There are some references in the "Software" documentation to branches that have been deleted. These should be replaced with the last tag of the referenced release.

PR #2522, opened 2023-07-05
Replace deleted branch names with working tag names

danielhoherd:
Closes https://github.com/astronomer/docs/issues/2521 by replacing deleted branch names in broken URLs with working tag names.

These versions are deprecated, but some folks including CRE engineers still use these docs.


jwitz (approved):
Thanks for writing the issue AND making the fix @danielhoherd!!

PR #2527, opened 2023-07-07
Title shortening suggestions

TJaniF:
As per the Slack discussion :)

These titles are just what made sense to me, happy to hear other suggestions!

### DIFF
--- a/learn/cloud-ide-tutorial.md
+++ b/learn/cloud-ide-tutorial.md
@@ -1,6 +1,6 @@
 ---
-title: 'Write and schedule a simple ML pipeline using the Astro Cloud IDE'
-sidebar_label: 'Astro Cloud IDE - Write an ML pipeline'
+title: 'ML with the Astro Cloud IDE'
+sidebar_label: 'ML with the Astro Cloud IDE'
 id: cloud-ide-tutorial
 description: 'Use tutorials and guides to make the most out of Airflow and Astronomer.'
 ---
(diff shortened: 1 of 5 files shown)


learn/use-case-airflow-dbt.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: "ELT pipelines with Airflow, dbt Core, Cosmos, and the Astro Python SDK"
+title: "ELT pipelines with Airflow and dbt Core"
```
kentdanas, line 2: I think we should be consistent between this and the databricks one with either "ELT pipelines" or just "ELT". I would probably lean towards just "ELT", but not a strong preference.

learn/cloud-ide-tutorial.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: 'Write and schedule a simple ML pipeline using the Astro Cloud IDE'
-sidebar_label: 'Astro Cloud IDE - Write an ML pipeline'
+title: 'ML with the Astro Cloud IDE'
+sidebar_label: 'ML with the Astro Cloud IDE'
```
kentdanas, line 3: Do you think ML Ops would make sense here? Or maybe "ML pipelines"? Just "ML" seems too broad to me, at least for the title. For the sidebar label I think it's fine.

kentdanas (approved):
Looks good! A few minor comments, but I'm okay with whatever you decide for these.

learn/cloud-ide-tutorial.md
TJaniF: I think ML Ops might give people false expectations, like interaction with ML Ops tools. I do like ML pipelines, will change it to that :)

TJaniF (commented):
@kentdanas thank you for the review, used both suggestions :)

PR #2529, opened 2023-07-07
Fix astro registry CLI command docs

TJaniF:
Hey :) 
Found out that it is `astro registry dag add` instead of `astro registry add dag` and the same for adding a provider. :)
Also the `upload_files_to_s3` DAG does not have a 1.2.0 version yet


jwitz (approved):
Oof thank you @TJaniF ! cc @sunkickr Not sure how this got through review 🤔

PR #2541, opened 2023-07-10
Runtime 8.7 release notes

jwitz:


astro/runtime-release-notes.md
```diff
@@ -29,7 +29,7 @@ To upgrade Astro Runtime, see [Upgrade Astro Runtime](upgrade-runtime.md). For g
 - Release date: July 10, 2023
 - Airflow version: 2.6.3
 
-## Airflow 2.6.3
+### Airflow 2.6.3
```
ephraimbuddy, line 32: Let's highlight this fix https://github.com/apache/airflow/pull/32293. It wasn't made a significant item in the release note due to the security nature of the fix. We can let users know that if triggering dags manually and they update the run_id in the configuration page, then the run_id must conform with the regex setting at `[scheduler]allowed_run_id_pattern`.

Also, are we not going to include something like line 34 below for this note?

ephraimbuddy (approved):

astro/runtime-release-notes.md
```diff
@@ -24,6 +24,24 @@ Astro Runtime is a Docker image built and published by Astronomer that extends t
 
 To upgrade Astro Runtime, see [Upgrade Astro Runtime](upgrade-runtime.md). For general product release notes, see [Astro Release Notes](release-notes.md). If you have any questions or a bug to report, contact [Astronomer support](https://cloud.astronomer.io/support).
 
+## Astro Runtime 8.7.0
+
+- Release date: July 10, 2023
```
jwitz, line 29:
```suggestion
- Release date: July 11, 2023
```

Issue #2468, opened 2023-06-22
Algolia search returns tab titles in results


jwitz:
### Issue Type

Issue with website functionality

### Links to Affected Docs

All

### The Issue or Context

When you search something in the Astro search bar, tab names come up as possible headers to click on. This isn't ideal, as some tab names seem to be relevant to the search, but end up on a totally unrelated page.

### Fix or Proposed Change

Add a rule to our Algolia crawling so that it disregards tabs.

PR #2561, opened 2023-07-13
Create exclusion pattern for tab URLs in Algolia

jwitz:
Resolves https://github.com/astronomer/docs/issues/2468 (bluntly)

This PR should remove all duplicate indexes of docs by excluding any duplicate URLs that include tabIDs. Syntax is based on regex. Docs for config are here: https://docsearch.algolia.com/docs/legacy/config-file/#stop_urls-optional

The downside of this approach is that search results will no longer open the correct tab where content might be, so the content a user searches for and selects might be hidden behind a tab until they manually select it. The ideal solution would be:

- If tabs exist for a given topic, exclude the version of the URL without tabs
- It tabs don't exist for a topic, do not exclude the URL

But I don't really have the bandwidth to mess with regex like that and this will at least make our search not be broken in most uses.

### DIFF
--- a/algolia-config.json
+++ b/algolia-config.json
@@ -4,7 +4,9 @@
     "https://docs.astronomer.io"
   ],
   "stop_urls": [
-    "(?=tab=)"
+    {
+    "url": "(?=tab=)"
+    }
   ],
   "selectors": {
     "lvl0": {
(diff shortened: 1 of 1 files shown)


stuart23:
Yeah, that ideal behaviour would be great, but I think this is still pretty intuitive and a step forwards rather than backwards.

katiebrady-astro (approved):
Also looks good to me! ✅

PR #2568, opened 2023-07-14
Fix reindex script, `stop_urls` for Algolia search

katiebrady-astro:
This patch fixes the reindex action. Looks like the issue was with the format of `stop_urls` array. Sorry for all the PRs! 🙏


jwitz:
I'll merge this, but I think we might get an error on the Algolia side based on this doc: https://docsearch.algolia.com/docs/legacy/config-file/#using-regular-expressions

PR #2572, opened 2023-07-17
Tiny PR: highlight what not to miss in ENV for cloud auth and remove a typo

TJaniF:
Based on a recent Slack thread, I think it is easy to miss what the individual clouds need as an extra key in `AIRFLOW__SECRETS__BACKEND_KWARGS`. Made a note for each to highlight that and removed a dangling `)`.


jwitz (approved):
I wish there was an easier way to show the value in the code block, but you can't add newlines to an environment variable right? This works for now!

TJaniF:
@jwitz if there is a way I have not found it haha, definitely tried before because I also really don't like having to side-scroll. 😅

PR #2588, opened 2023-07-19
Move 'Intro to Airflow'

paolaperaza:
Found myself looking for this doc elsewhere in the menu. Since it's an intro, should it be in our getting started menu?

Also, kind of confusing to have `Tutorials` in getting started but then a separate `Airflow tutorials` section elsewhere in the menu. I wonder if we could just remove the `tutorials` header there, or maybe just keep part 1 and throw part 2 into the main tutorials menu?

### DIFF
--- a/sidebarsLearn.js
+++ b/sidebarsLearn.js
@@ -7,9 +7,10 @@ module.exports = {
       link: {
         type: 'generated-index',
         title: 'Get started',
-        description: 'Get started with Airflow.',
+        description: 'Get started with Apache Airflow.',
       },
       items: [
+        'intro-to-airflow',
         'airflow-quickstart',
         {
           type: 'category',
@@ -47,7 +48,6 @@ module.exports = {
             'connections',
             'dags',
             'what-is-a-hook',
-            'intro-to-airflow',
             'managing-airflow-code',
             'airflow-openlineage',
             'what-is-an-operator',


TJaniF (approved):
I think this makes a lot of sense! 

With the doc being more prominent I wonder if there should be an `Airflow features` section that links out to more info on key features like Datasets, dynamic task mapping etc, that are relevant to people who might be evaluating several orchestrators to decide which one to use 🤔 

I'll put looking at that on the backlog :) 

About the tutorials, I agree the header names are not ideal. I wonder if putting part 2 into the main tutorials section would be confusing either for people who find part 2 there and for people who look for part 2 after completing part 1... 
Maybe the section could be titled `First steps` instead? or something like that 😅

jwitz:
Merging this as is cus it looks great! @TJaniF what if we made the category name "Quickstart" and used the Docusaurus [doc item sidebar](https://docusaurus.io/docs/sidebar/items#category-doc-link) to make our existing "[quickstart](https://docs.astronomer.io/learn/airflow-quickstart)" guide appear when you click the collapsible menu button?

TJaniF:
Hmmmm @jwitz I think this might be confusing because the quickstart guide has nothing to do with the tutorial that has a part 1 and 2 😅 . Maybe that is my brain but I would assume if Part 1 is listed under a header that the header is the overview over the parts. IA is hard, hmm... I'll put this on my showerthought list, will see if 🧠 can come up with something.

jwitz:
D'oh... morning 🧠 at work

PR #2602, opened 2023-07-24
Postgres, MSSQL Server and Azure blob storage

manmeetkaur:
This is the 3rd and final PR in the how-to-connection series. This adds the following connection types:

- Postgres
- MSSQL Server
- Azure Blob storage


learn/connections/ms-sqlserver.md
```diff
@@ -0,0 +1,91 @@
...
+
+- Host (also known as the endpoint URL, server name, or instance ID depending on your cloud provider)
+- Port (default is 1433)
+- Username
+- Password
+- Schema (default is `dbo`)
+
+The method to retrieve these values will vary based which cloud provider you use to host Microsoft SQL Server. Refer to the following documents to for more information about retrieveing these values.
+
+- AWS: Connect to Microsoft SQL Server running [on RDS](https://aws.amazon.com/getting-started/hands-on/create-microsoft-sql-db/)
```
jwitz, line 32: Either here or in a follow-up PR, we should make these tabs and include the steps for each provider. We don't need to use tabs for the "Create the connection" section because there's only one value that changes between all the setups.

static/img/examples/connection-azure-blob-storage-conn-string.png
jwitz, line 1: Now that you have Snagit, I'd like to see all of these images with a border 😉 . We should also crop the image so that it only covers the values that a user would enter. For example, here we could crop everything from the bottom up to **Blob storage shared access key** Because the user doesn't have to enter those values.

jwitz (changes requested):
Will take one more look after the screenshots are addressed, but overall LGTM!

static/img/examples/connection-azure-blob-storage-conn-string.png
manmeetkaur: @jwitz I don't agree with cropping the image, because the user will get confused why they see certain fields and we don't have those fields. I kept all the fields in the view on purpose.

learn/connections/ms-sqlserver.md
manmeetkaur: Honestly, I don't think that is worth the time and effort, because a user familiar with these tools would know how to get the host, port etc. as they are very common. the goal here is to let the users teach "how to create a connection in Airflow" and we are explaining that pretty clearly in Create your connection.

learn/connections/postgres.md
```diff
@@ -0,0 +1,78 @@
...
+
+A connection from Airflow to Postgres requires the following information:
+
+- Host (also known as the endpoint URL, server name, or instance ID based on your cloud provider)
+- Port (default is 5432)
+- Username 
+- Password
+- Schema (default is `public`)
+
+The method to retrieve these values will vary based which cloud provider you use to host Microsoft SQL Server. Refer to the following documents to for more information about retrieveing these values:
```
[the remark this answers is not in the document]
manmeetkaur: same comment as above

manmeetkaur:
@jwitz added borders. ready for review.

learn/connections/azure-data-factory.md
```diff
@@ -33,7 +33,7 @@ Complete the following steps to retrieve all of these values:
 1. In your Azure portal, open your [data factory](https://portal.azure.com/#view/HubsExtension/BrowseResource/resourceType/Microsoft.DataFactory%2FdataFactories) service and select the subscription that contains your data factory.
 2. Copy the **Name** of your data factory and the **Resource group**.
 3. Click on the subscription for your data factory, then copy the **Subscription ID** from the subscription window.
-4. Go to your [Azure AD application](https://portal.azure.com/#view/Microsoft_AAD_RegisteredApps/ApplicationsListBlade). Then, from the **Overview** tab, copy the **Application (client) ID** and **Directory (tenant) ID**.
+4. Navigate to your Azure AD application. Then, from the **Overview** tab, copy the **Application (client) ID** and **Directory (tenant) ID**.
```
jwitz, line 36:
```suggestion
4. Open your Azure AD application. Then, from the **Overview** tab, copy the **Application (client) ID** and **Directory (tenant) ID**.
```
"Open" or "Go" is always preferred to "Navigate". Navigate has other connotations and is longer and more complicated than the former two options

static/img/examples/connection-azure-blob-storage-key.png
jwitz, line 1: This should be cropped up to  **Blob Storage Key**

static/img/examples/connection-ms-sqlserver.png
jwitz, line 1: We don't need to show Extra here, we can crop it

static/img/examples/connection-postgres.png
jwitz, line 1: See previous comment

jwitz (approved):
LGTM

PR #2604, opened 2023-07-24
Update install-cli.md

fritz-astronomer:
Update the docs to a more-recent version of the `astro` CLI for the correct naming convention


jwitz (approved):
Thanks @fritz-astronomer !

PR #2623, opened 2023-07-31
Add missing system vars

manmeetkaur:
Add some missing system vars that are generic Airflow environment variables.


astro/platform-variables.md
```diff
@@ -45,6 +47,7 @@ If you need to set one of these variables for a particular use case, contact [As
 | `AIRFLOW__METRICS__STATSD_ON`                | Determines whether metrics are sent to Statsd                                                                                   | `True`                                                     |
 | `AIRFLOW__METRICS__STATSD_HOST`              | The hostname for sending metrics to Statsd                                                                                      | `statsd.Hostname`                                          |
 | `AIRFLOW__METRICS__STATSD_PORT`              | The port for sending metrics to Statsd                                                                                          | `<statsd-metrics-port>`                                    |
+| `AIRFLOW__METRICS__STATSD_PREFIX`            | The prefix for sending the metrics to Statsd                                                                                    | `airflow`                                                  |
```
melugoyal, line 50: can't leave a comment on the lines below, but it's no longer dangerous to set `AWS_ACCESS_KEY_ID` or `AWS_SECRET_ACCESS_KEY`.

`GOOGLE_APPLICATION_CREDENTIALS` is dangerous to set, specifically on GCP. doing so will prevent task logs from being written to the GCS bucket.

astro/platform-variables.md
```diff
@@ -27,6 +27,8 @@ If you need to set one of these variables for a particular use case, contact [As
...
 | `AIRFLOW__LOGGING__REMOTE_BASE_LOG_FOLDER`   | Location of remote logging storage                                                                                              | `baseLogFolder`                                            |
+| `AIRFLOW__LOGGING__REMOTE_LOG_CONN_ID`       | Connection ID of the connection that provides access to your remote logging location                                            | `<connection-name>`                                        |
+| `AIRFLOW__LOGGING__LOGGING_CONFIG_CLASS`     | The class name that specifies the logging configuration                                                                         |`astronomer.runtime.logging.logging_config`                |
 | `AIRFLOW_CONN_ASTRO_S3_LOGGING`              | Connection URI for writing task logs to Astro's managed S3 bucket                                                               | `<Connection-URI>`                                         |
```
melugoyal, line 32: in addition to this, `AIRFLOW_CONN_ASTRO_GCS_LOGGING` can't be overwritten on GCP, and `AIRFLOW_CONN_ASTRO_AZURE_LOGS` on Azure.

but technically, we don't allow these env vars to be overwritten at all, ie we set them no matter what. the text above says "The Cloud UI does not currently prevent you from setting these environment variables, but attempting to set them can result in unexpected behavior that can include access problems, missing task logs, and failed tasks.". should we change that text, or trim this list to only the variables that are actually dangerous/that we don't replace?

astro/platform-variables.md
manmeetkaur: Thanks @melugoyal I will update the PR with these details.

astro/platform-variables.md
manmeetkaur: That is a good point. So basically there are three things:
- Env Vars that can't be overwritten even though it seems like they are overwritten - actually have no impact 
- Env Vars that can be overwritten and can have an adverse impact
- Env Vars which are not [Airflow Env vars](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html) but are Astro env vars like the ones you mentioned above 

I feel like we should definitely mention about the first two. About the third one, on one hand I feel we can omit Astro Env vars, but on the other hand, I feel like we should publish them as RESERVED KEYWORDs or Env Vars, just in case if someone accidentally uses that name and it affects our settings.

jwitz: Yeah I'm not sure the best way to document that third type. If the Cloud UI doesn't provide a detailed error message when you try to set those variables, we should document them here in a new section called `Reserved variable names`. Ideally though, there would just be good enough error checking in the UI and we wouldn't have to publish this list. I think it's okay to not include the values if they're irrelevant to users

astro/platform-variables.md
```diff
@@ -27,6 +27,8 @@ If you need to set one of these variables for a particular use case, contact [As
...
 | `AIRFLOW__LOGGING__REMOTE_LOGGING`           | Enables remote logging                                                                                                          | `True`                                                     |
 | `AIRFLOW__LOGGING__REMOTE_BASE_LOG_FOLDER`   | Location of remote logging storage                                                                                              | `baseLogFolder`                                            |
+| `AIRFLOW__LOGGING__REMOTE_LOG_CONN_ID`       | Connection ID of the connection that provides access to your remote logging location                                            | `<connection-name>`                                        |
+| `AIRFLOW__LOGGING__LOGGING_CONFIG_CLASS`     | The class name that specifies the logging configuration                                                                         |`astronomer.runtime.logging.logging_config`                |
```
jwitz, line 31:
```suggestion
| `AIRFLOW__LOGGING__LOGGING_CONFIG_CLASS`     | Class name that specifies the logging configuration                                                                         |`astronomer.runtime.logging.logging_config`                |
```

astro/platform-variables.md
```diff
@@ -27,6 +27,8 @@ If you need to set one of these variables for a particular use case, contact [As
 | `AIRFLOW__LOGGING__DAG_PROCESSOR_LOG_TARGET` | Routes scheduler logs to stdout                                                                                                 | `stdout`                                                   |
 | `AIRFLOW__LOGGING__REMOTE_LOGGING`           | Enables remote logging                                                                                                          | `True`                                                     |
 | `AIRFLOW__LOGGING__REMOTE_BASE_LOG_FOLDER`   | Location of remote logging storage                                                                                              | `baseLogFolder`                                            |
+| `AIRFLOW__LOGGING__REMOTE_LOG_CONN_ID`       | Connection ID of the connection that provides access to your remote logging location                                            | `<connection-name>`                                        |
```
jwitz, line 30:
```suggestion
| `AIRFLOW__LOGGING__REMOTE_LOG_CONN_ID`       | ID of the connection that provides access to your remote logging location                                            | `<connection-name>`                                        |
```

jwitz (approved):
Approving the additions, let's cover the reogranization of the variables in a separate PR

Issue #2606, opened 2023-07-25
Elaborate all the possible options in Create Deployment


manmeetkaur:
### Issue Type

Missing content for an existing feature

### Links to Affected Docs

https://docs.astronomer.io/astro/create-deployment

### The Issue or Context

We are not clearly explaining all the methods to create a Deployment in one place. Can't expect the user to jump to diff pages and go oh! this is another way.

Other issues:
- The CLI box doesn’t feel like enough given CLI command reference is sparse
- There’s a double title

Related slack thread: https://astronomer.slack.com/archives/C03CT3DNZMH/p1690260650927989

### Fix or Proposed Change

Add/eXpand the following:
- Overview
- via UI
- via CLI
- via deployment-as-code

Issue #2536, opened 2023-07-10
Review & add screenshots on `Create a Deployment`


paolaperaza:
### Context

Creating a Deployment is the first thing any new user on Astro does in the Cloud UI. This doc is ok, but I think it warrants a being a bit more engaging and visually accessible: https://docs.astronomer.io/astro/create-deployment

### Ask

- [ ] Consider changes in this doc based on the new Deployment creation screen
- [ ] Fix release notes? Not quite right - https://docs.astronomer.io/astro/release-notes#additional-improvements
- [ ] Add screenshot of Deployment create screen (we'll sadly be having to update this relatively often)

PR #2650, opened 2023-08-04
Update Deployment creation options

jwitz:
Resolves https://github.com/astronomer/docs/issues/2606

I don't think we need to "elaborate" on the programmatic options in this doc. Those are complex tools for complex use cases, and this doc is more meant to help explain each item in the Cloud UI configuration screen that a user might want to know more about. Toploading it with too much expository info/ strategy info might frustrate users who just want to get started with a Deployment

### DIFF
--- a/astro/create-deployment.md
+++ b/astro/create-deployment.md
@@ -11,11 +11,16 @@ id: create-deployment
 
 An Astro Deployment is an Astro Runtime environment that is powered by the core components of Apache Airflow, including the Airflow webserver, scheduler, and one or more workers.
 
-You can create a Deployment from a Workspace on Astro. After you create a Deployment, you can deploy DAGs to it from the Astro CLI or from a continuous delivery (CI/CD) process. All DAGs and tasks on Astro are executed within a Deployment.
+The most basic way to create a Deployment is through the Cloud UI. After you create a Deployment, you can deploy DAGs to it from the Astro CLI or from a continuous delivery (CI/CD) process. All DAGs and tasks on Astro are executed within a Deployment.
 
 ![Deployment creation screen example](/img/docs/create-deployment.png)
 
-Every Deployment is hosted on a single Astro cluster with its own dedicated resources, which you can customize to meet the unique requirements of your Organization. Every Astro cluster operates with a primary database that hosts the individual databases for each Deployment.  To restrict communication between Deployments, resources for each Deployment are isolated within a corresponding Kubernetes namespace. See [Deployment network isolation](data-protection.md#deployment-network-isolation).
+Every Deployment is hosted on a single Astro cluster with its own dedicated resources, which you can customize to meet the unique requirements of your Organization. Every Astro cluster operates with a primary database that hosts the individual databases for each Deployment. To restrict communication between Deployments, resources for each Deployment are isolated within a corresponding Kubernetes namespace. See [Deployment network isolation](data-protection.md#deployment-network-isolation).
+
+This document explains how to create Deployments through the Cloud UI. To create Deployments programmatically using the Astro CLI, see:
+
+- [Astro CLI reference guide](cli/astro-deployment-create.md)
+- [Manage Deployments as code](manage-deployments-as-code.md)
 
 ## Prerequisites
 
@@ -29,13 +34,14 @@ If you prefer, you can also run `astro deployment create` to create a Deployment
 
 :::
 
-1. In the Cloud UI, select a Workspace.
+1. In the Cloud UI, go to **Deployments**.
 
-2. On the **Deployments** page, click **+ Deployment**.
+2. Click **+ Deployment**.
 
     ![Deployment creation button](/img/docs/create-button.png)
 
 3. Complete the following fields:
+
     - **Basic**  
 
         - **Name**: Enter a name for your Deployment.
--- a/astro/create-deployment.md
+++ b/astro/create-deployment.md
@@ -13,9 +13,9 @@ An Astro Deployment is an Airflow environment that is powered by [Astro Runtime]
 
 There are three ways to create a Deployment: 
 
-- Using the Cloud UI. This is the most basic way to create a Deployment and is the focus of this document.
-- Using [`astro deployment create`](cli/astro-deployment-create.md).
-- Using a Deployment template file. See [Manage Deployments as code](manage-deployments-as-code.md).
+- Manually, using the Cloud UI. This is the most basic way to create a Deployment and is the focus of this document.
+- Programmatically, using a Deployment template file. See [Manage Deployments as code](manage-deployments-as-code.md).
+- Programmatically, using [`astro deployment create`](cli/astro-deployment-create.md).
 
 After you create a Deployment, you can deploy DAGs to it using the Astro CLI on your local machine or a continuous integration/continuous delivery (CI/CD) tool. All DAGs and tasks on Astro are executed within a Deployment.
 
--- a/astro/create-deployment.md
+++ b/astro/create-deployment.md
@@ -38,7 +38,9 @@ This document explains how to create Deployments through the Cloud UI. To create
 
 5. Click **Create Deployment**.
 
-     A confirmation message appears indicating that the Deployment is in progress. Select the **Deployments** link to go to the **Deployments** page. The Deployment status is **Creating** until all underlying components in your Astro cluster are healthy, including the Airflow webserver and scheduler. During this time, the Airflow UI is unavailable and you can't deploy code or modify Deployment settings. When the Deployment is ready, the status changes to **Healthy**.
+     A confirmation message appears indicating that the Deployment is in progress. The Deployment status is **Creating** until all underlying components in your Astro cluster are healthy, including the Airflow webserver and scheduler. During this time, the Airflow UI is unavailable and you can't deploy code or modify Deployment settings. When the Deployment is ready, the status changes to **Healthy**.
+     
+You can continue to access your other Deployments by selecting the **Deployments** link.
     
     For more information about possible Deployment health statuses, see [Deployment health](deployment-metrics.md#deployment-health).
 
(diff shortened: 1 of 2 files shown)


manmeetkaur (changes requested):
@jwitz I think "create a deployment" should just tell what steps you should focus on "how to  create " very similar to what we are doing with other pages. Currently the screenshot is at the top, which i think obstructs and also too much detail. I have created an issue to discuss specifically this where we create a page/section for Astro components, where the detail could reside. Although there are errors on this page that should be corrected, we should also discuss that issue on Monday.

jwitz:
@manmeetkaur The screenshot was requested by Paola, so you'll have to discuss that with her. I'm fine with/without it https://github.com/astronomer/docs/issues/2536

As for the content, it all depends whether we want to tell users how to create a Deployment without _any_ information on the configuration screen they see. Is it enough for us to say "Configure your resources. See this other doc"? I think it's helpful to mention some smart defaults and the bare minimum fields to configure in this doc, even if it makes it a little bit longer.

jwitz:
@manmeetkaur Made the changes per discussion in standup, ready for you to look again!

astro/create-deployment.md
```diff
@@ -9,60 +9,34 @@ id: create-deployment
...
-    
-    These are the essential fields for configuring a Deployment. You can optionally configure additional details for your Deployment based on resource usage, user permissions, and resiliency. For all available Deployment options, see [Configure Deployment resources](configure-deployment-resources.md).
+4. (Optional) Configure other details for your Deployment, including cluster, executor, and worker resources. If you have no specific resource requirements for running Airflow, the default configurations should work in most cases. For all available Deployment options, see [Configure Deployment resources](configure-deployment-resources.md).
 
-4. Click **Create Deployment**.
+5. Click **Create Deployment**.
 
      A confirmation message appears indicating that the Deployment is in progress. Select the **Deployments** link to go to the **Deployments** page. The Deployment status is **Creating** until all underlying components in your Astro cluster are healthy, including the Airflow webserver and scheduler. During this time, the Airflow UI is unavailable and you can't deploy code or modify Deployment settings. When the Deployment is ready, the status changes to **Healthy**.
```
manmeetkaur, line 41: The confirmation message appears on the same page and users necessarily don't need to go the Deployments page again. 
```suggestion
     A confirmation message appears indicating that the Deployment is in progress. The Deployment status is **Creating** until all underlying components in your Astro cluster are healthy, including the Airflow webserver and scheduler. During this time, the Airflow UI is unavailable and you can't deploy code or modify Deployment settings. When the Deployment is ready, the status changes to **Healthy**.
     
You can continue to access your other Deployments by selecting the **Deployments** link.
```

manmeetkaur (commented):
Hey @jwitz I think what we discussed was three (or two ? by combining both astro cli options in 1) tabs 
- Cloud UI
- Astro CLI
  - `astro login` -> `astro deployment create ...` -> `astro deployment list`
- Deployment as code
  - `astro deployment inspect ...` -> edit the template file -> `astro deployment create --deployment-file ...` 

this way all options are available to the user in 1 place and other pages are just for reference. Do you want me to make these changes? Happy to take it up!

astro/create-deployment.md
```diff
@@ -9,62 +9,38 @@ id: create-deployment
...
 
-![Deployment creation screen example](/img/docs/create-deployment.png)
+Every Deployment is hosted on an Astro cluster with its own dedicated resources, which you can customize to fine-tune your resource usage. To restrict communication between Deployments, resources for each Deployment are isolated within a corresponding Kubernetes namespace. See [Deployment network isolation](data-protection.md#deployment-network-isolation).
 
-Every Deployment is hosted on a single Astro cluster with its own dedicated resources, which you can customize to meet the unique requirements of your Organization. Every Astro cluster operates with a primary database that hosts the individual databases for each Deployment.  To restrict communication between Deployments, resources for each Deployment are isolated within a corresponding Kubernetes namespace. See [Deployment network isolation](data-protection.md#deployment-network-isolation).
+This document explains how to create Deployments through the Cloud UI. To create Deployments programmatically using the Astro CLI, see:
+
+- [Astro CLI reference guide](cli/astro-deployment-create.md)
```
jwitz, line 20:
```suggestion
- [`astro deployment create`](cli/astro-deployment-create.md)
```

jwitz:
@manmeetkaur I don't agree with including the CLI and as-code steps as tabs in this doc. Those are workflows for different contexts and should be mentioned in different docs. 

Would a user realistically choose between creating a Deployment in the UI and creating one programmatically? If they just want to get a Deployment going ASAP, they will almost always want to use the UI. The only circumstance where that wouldn't be the case is if they don't have access to a web browser. Our docs should make that _the_ manual option if that's what we recommend - making them choose between the CLI and UI could cause more confusion than convenience.

This is the same pattern I thought we were following for all automation docs: include the manual UI steps in one doc and the automation scripts/ examples in another. I don't remember whether we agreed on tabs, I only remember our decision to mention the automation options _somewhere_ in the doc.

manmeetkaur:
@jwitz I actually think quite the opposite. It will depend on the user, what pattern to use based on the use-case. Our goal is to clearly make all the info available on a single page with steps without any overview or concept explanation. Users will not have to think, which doc to open if they want to use as code or via plain CLI command, they will know "Create a Deployment" explains all options, let's go there. Why and what of the process can stay where it is currently. 

It also explains to what @paolaperaza mentioned that our CLI is under-documented. I feel it is not under-documented, it not documented in the correct context.

Think about a user A, who wants to use the CLI to create a deployment. Natural progression is to come to create a Deployment, and be given the option of all CLI commands that are required not to push him/her to a reference doc, where they might need to jump couple of more pages to find how to login, how to select workspace etc. goal is new users come in, and see 

```
astro login
astro workspace switch
astro deployment create ...
astro deployment list
```
One-stop shop to help you login. We don't currently dictate that our docs are just cloud ui or don't necessarily have to I think. 

This is worth discussing with @paolaperaza again I believe. Because this was my understanding, so may be we clear this out.

jwitz:
I don't think this logic extends nicely to the other parts of our documentation. For "Configure Deployment resources", do we want to have tabs for each configuration explaining how to update it via deployments-as-code? What about for user management? It also breaks the pattern that we're currently working towards of keeping complex automation docs in a separate location from the least-friction UI workflows.

lzdanski:
FWIW - with the API as another path in the future, we're going to have another programmatic way of creating and managing deployments. To me, it makes more sense to keep the UI path separate from the more complex and possibly nuanced options where this doc can works as an introduction/orientation to the UI and briefly describe that there are other options for managing deployments with code. (I don't have a strong opinion about the CLI/automation docs, but thought what we decide for the API might be a pattern we can use here too)

jwitz:
@manmeetkaur Want to give this a final pass when you have the chance?

astro/create-deployment.md
```diff
@@ -9,62 +9,39 @@ id: create-deployment
...
-An Astro Deployment is an Astro Runtime environment that is powered by the core components of Apache Airflow, including the Airflow webserver, scheduler, and one or more workers.
+An Astro Deployment is an Airflow environment that is powered by [Astro Runtime](runtime-image-architecture.md). It runs all core Airflow components, including the Airflow webserver, scheduler, and workers, plus additional tooling for reliability and observability.
 
-You can create a Deployment from a Workspace on Astro. After you create a Deployment, you can deploy DAGs to it from the Astro CLI or from a continuous delivery (CI/CD) process. All DAGs and tasks on Astro are executed within a Deployment.
+There are three ways to create a Deployment: 
 
-![Deployment creation screen example](/img/docs/create-deployment.png)
+- Manually, using the Cloud UI. This is the most basic way to create a Deployment and is the focus of this document.
+- Programmatically, using a Deployment template file. See [Manage Deployments as code](manage-deployments-as-code.md).
+- Programmatically, using [`astro deployment create`](cli/astro-deployment-create.md).
```
manmeetkaur, line 18:
```suggestion
- Manually, using the Cloud UI. This is the most basic way to create a Deployment and is the focus of this document.
- Programmatically, using [`astro deployment create`](cli/astro-deployment-create.md).
- Programmatically, using a Deployment template file. See [Manage Deployments as code](manage-deployments-as-code.md#create-a-deployment-from-a-template-file).
```

astro/create-deployment.md
```diff
@@ -9,62 +9,39 @@ id: create-deployment
...
 
-4. Click **Create Deployment**.
+5. Click **Create Deployment**.
 
-     A confirmation message appears indicating that the Deployment is in progress. Select the **Deployments** link to go to the **Deployments** page. The Deployment status is **Creating** until all underlying components in your Astro cluster are healthy, including the Airflow webserver and scheduler. During this time, the Airflow UI is unavailable and you can't deploy code or modify Deployment settings. When the Deployment is ready, the status changes to **Healthy**.
+     A confirmation message appears indicating that the Deployment is in progress. The Deployment status is **Creating** until all underlying components in your Astro cluster are healthy, including the Airflow webserver and scheduler. During this time, the Airflow UI is unavailable and you can't deploy code or modify Deployment settings. When the Deployment is ready, the status changes to **Healthy**.
+     
+You can continue to access your other Deployments by selecting the **Deployments** link.
     
     For more information about possible Deployment health statuses, see [Deployment health](deployment-metrics.md#deployment-health).
```
manmeetkaur, line 46:
```suggestion
For more information about possible Deployment health statuses, see [Deployment health](deployment-metrics.md#deployment-health).
```

astro/create-deployment.md
manmeetkaur: I feel we should put the simple command before the template file and also point to the exact section in the manage a deployment page. Just a thought!

manmeetkaur (approved):
Just minor change for formatting and order change. Do you think we should create a issue to create a page - idk where - which talks about steps of login to Astro CLI -> astro create deployment to avoid linking to the command reference ?

PR #2673, opened 2023-08-09
Fix mapping steps for Azure AD SCIM

jwitz:
Resolves https://github.com/astronomer/docs/pull/2673

### DIFF
--- a/astro/set-up-scim-provisioning.md
+++ b/astro/set-up-scim-provisioning.md
@@ -131,14 +131,27 @@ Complete the manual setup if you configured your existing Astro app without usin
     - **Provisioning mode**: Set to **Automatic**.
     - **Admin Credentials** > **Tenant URL**: Enter the **SCIM integration URL** including the Azure AD feature flag parameter.
     - **Secret Token**: Enter your Organization API token. 
-    - **Mappings**: Configure the following mappings:
+
+8. In **Mappings**, open the **Groups** mapping configuration.
+9. In **Target Object Actions**, tick the checkboxes for **Create**, **Update**, and **Delete**.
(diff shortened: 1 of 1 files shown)


astro/set-up-scim-provisioning.md
```diff
@@ -131,14 +131,27 @@ Complete the manual setup if you configured your existing Astro app without usin
...
     | -------------------------------- | --------------- |
-    | userPrincipalName                | userName        |
     | displayName                      | displayName     |
-    | givenName                        | name.givenName  |
-    | surname                          | name.familyName |
+    | members                          | members         |
+
+11. Go back to the **Mappings** menu and open the **User** mapping configuration.
+12. In **Target Object Actions**, tick the checkboxes for **Create**, **Update**, and **Delete**.
+13. In the **Attribute Mappings** table, add the following mappings:
```
aliotta, line 146: We need to tell them they have to explicitly delete any mapping not in this chart. TBH I would include in our docs a snapshot of what the azure screen looks like afterwards.

astro/set-up-scim-provisioning.md
```diff
@@ -131,14 +131,27 @@ Complete the manual setup if you configured your existing Astro app without usin
...
+
+    | Azure Active Directory Attribute                            | Astro Attribute |
+    | ----------------------------------------------------------- | --------------- |
+    | userPrincipalName                                           | userName        |
+    | Switch([IsSoftDeleted], , "False", "True", "True", "False") | active          |
+    | displayName                                                 | displayName     |
+    | givenName                                                   | name.givenName  |
```
aliotta, line 153: givenName and surname are incorrect. They should not be in this table.

PR #2680, opened 2023-08-10
Replace flowchart with screenshot due to darkmode issue

TJaniF:
Hey, Rocco pointed this out and I decided to offer a quick fix, though maybe there is a reason why the flowchart was defined in markdown here (I had no idea this was a thing this is really cool)?


TJaniF:
Fixed by @jwitz via 🧜‍♀️ hotfix :)

PR #2683, opened 2023-08-11
Create data dependency doc from existing content

jwitz:
For an upcoming IDE release that relies on this content in help text. Also added a section on how to source external data in a task, would be great to flesh that out more


astro/cloud-ide/pass-data-to-cells.md
```diff
@@ -0,0 +1,85 @@
...
+
+The table created by the `SELECT` statement is automatically converted to pandas DataFrame and passed to the Python cell.
+
+The following Python cell is dependent on a SQL cell named `my_sql_cell`.
+
+```python
+df = my_sql_cell # my_sql_cell is a SQL cell which gets converted to a pandas DataFrame by default
+df['col_a'] = df['col_a'] + 1
+return df
+```
```
jlaneve, line 49: now that we're going to have explicit data types shown in the UI ("this is a table", "this is a string", etc), what do you think about collapsing each of these cell type sections to a single example? I believe what we want the user to understand after reading the doc is "this is the syntax that I can use in my XX cell to reference my other cell's output"

astro/cloud-ide/pass-data-to-cells.md
```diff
@@ -0,0 +1,85 @@
+---
+sidebar_label: Pass data to cells
+title: Pass data to Astro Cloud IDE cells
+id: pass-data-to-cells
+description: Learn how to run Python code by creating and configuring Python cells in the Astro Cloud IDE.
+---
+
+You can use the output of an Astro Cloud IDE cell as the input for another cell in your pipeline. When you pass data from one cell to another, you create a data dependency between those tasks. A cell that is downstream of a data dependency will not run until the upstream cell finishes and produces data. The Astro Cloud IDE automatically renders these dependencies in your project code and in the **Pipeline** view of your project.
```
lzdanski, line 8:
```suggestion
You can use the output of an Astro Cloud IDE cell as the input for another cell in your pipeline. When you pass data from one cell to another, you create a data dependency between those tasks and defines the order in which the cells must be executed. A cell that is downstream of a data dependency will not run until the upstream cell finishes and produces data. The Astro Cloud IDE automatically renders these dependencies in your project code and orders them in the **Pipeline** view of your project.
```
Digging through the IDE docs, noticed that we have information in the [quickstart](https://docs.astronomer.io/astro/cloud-ide/quickstart#step-7-create-dependencies-between-cells) about why dependencies are important for the IDE, but I didn't see the relationship called out between the dependency and DAG structure show up in this new doc.

lzdanski (commented):
Needed a smidge more detail added from the quickstart to this doc about the importance of cell dependencies for the IDE.

astro/cloud-ide/pass-data-to-cells.md
jwitz: Is the format the same for all data types in a given cell type? If so that's a great idea!

astro/cloud-ide/pass-data-to-cells.md
```diff
@@ -0,0 +1,70 @@
+---
+sidebar_label: Pass data to cells
+title: Pass data to Astro Cloud IDE cells
+id: pass-data-to-cells
```
jwitz, line 4:
```suggestion
id: pass-data-between-cells
```

PR #2689, opened 2023-08-15
Add page rank to crawl results

jwitz:
This is an attempt to add page ranking to our indices so that Software docs rarely if ever appear before Astro docs in the homepage search bar on https://docs.astronomer.io/

See https://docsearch.algolia.com/docs/legacy/config-file/#using-page-rank

I am worried that I'm misusing "selector key", but I also don't know how to test this without merging to main!

### DIFF
--- a/algolia-config.json
+++ b/algolia-config.json
@@ -3,10 +3,12 @@
   "start_urls": [
     {
       "url": "https://docs.astronomer.io/astro",
+      "selectors_key": "astro",
       "page_rank": 5
     },
     {
       "url": "https://docs.astronomer.io/software",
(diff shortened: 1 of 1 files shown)


algolia-config.json
```diff
@@ -1,7 +1,18 @@
...
   "start_urls": [
-    "https://docs.astronomer.io"
+    {
+      "url": "https://docs.astronomer.io/astro",
+      "page_rank": 5
+    },
+    {
+      "url": "https://docs.astronomer.io/learn",
+      "page_rank": 5
+    }
```
jwitz, line 11:
```suggestion
    },
```

PR #2695, opened 2023-08-16
Update configure-deployment-resources.md

RNHTTR:
Warn users about requesting resources for a task that are equal to the deployment's quota

### DIFF
--- a/astro/configure-deployment-resources.md
+++ b/astro/configure-deployment-resources.md
@@ -157,6 +157,7 @@ Set safeguards by configuring default Pod limits and requests from the Cloud UI.
 Your CPU and memory quotas determine how many tasks can run at once on your Deployment. For example, if your Deployment has a CPU quota of 3vCPU and a memory quota of 6GiB, and a task requests this amount, then your Deployment can run only that task until it completes.
 
 The CPU and memory quotas also determine the **Max Pod Size**, which is the maximum amount of resources that a task can request.
+
 :::caution
 
 The largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**. Because tasks can only run in a single Pod, your tasks can't request resources that exceed the **Max Pod Size**.
--- a/astro/configure-deployment-resources.md
+++ b/astro/configure-deployment-resources.md
@@ -161,7 +161,7 @@ The CPU and memory quotas also determine the **Max Pod Size**, which is the maxi
 
 :::caution
 
-The largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**. Because tasks can only run in a single Pod, your tasks can't request resources that exceed the **Max Pod Size**.
+The largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**. Because tasks run in a single Pod, your tasks can't request resources that exceed the **Max Pod Size**, even if your quota is larger.
 
 For example, if your Deployment has a CPU quota of 150vCPU and a memory quota of 300GiB, your **Max Pod Size** might only be 12 vCPU and 24GiB RAM. If you try to run a task that requests 20vCPU, the task won't run even though it's within your quotas.
 
--- a/astro/configure-deployment-resources.md
+++ b/astro/configure-deployment-resources.md
@@ -159,7 +159,7 @@ If the max pod size is equal to the deployment's quota, and a task requests the 
 
 :::
 
-The maximum quotas for CPU and memory are currently 400 vCPU and 800 GiB RAM, respectively. The CPU and memory quotas determine the **Max Pod Size**, which is the maximum amount of resources that a task can request for its pod. The maximum max pod size is currently allows for 27 vCPU and 54 GiB RAM for Standard Clusters and 12 vCPU and 24 GiB RAM for Dedicated Clusters.
+The CPU and memory quotas determine the **Max Pod Size**, which is the maximum amount of resources that a task can request for its pod.
 
 :::info Alternative Astro Hybrid setup
 
--- a/astro/configure-deployment-resources.md
+++ b/astro/configure-deployment-resources.md
@@ -153,13 +153,17 @@ Set safeguards by configuring default Pod limits and requests from the Cloud UI.
     - **Default Pod Size (CPU)**: The amount of CPUs that your tasks run with if no CPU usage is specified in their Pod configuration.
     - **Default Pod Size (Memory)**: The amount of memory that your tasks run with if no memory usage is specified in their Pod configuration.
 
-:::warning
 
-If the max pod size is equal to the deployment's quota, and a task requests the max pod size, only one task will run at a time, because that task is consuming the deployment's entire quota.
+Your CPU and memory quotas determine how many tasks can run at once on your Deployment. For example, if your Deployment has a CPU quota of 3vCPU and a memory quota of 6GiB, and a task requests this amount, then your Deployment can run only that task until it completes.
+
+The CPU and memory quotas also determine the **Max Pod Size**, which is the maximum amount of resources that a task can request.
+:::caution
+
+The largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**. Because tasks can only run in a single Pod, your tasks can't request resources that exceed the **Max Pod Size**.
+
+For example, if your Deployment has a CPU quota of 150vCPU and a memory quota of 300GiB, your **Max Pod Size** might only be 12 vCPU and 24GiB RAM. If you try to run a task that requests 20vCPU, the task won't run even though it's within your quotas.
 
 :::
-
-The CPU and memory quotas determine the **Max Pod Size**, which is the maximum amount of resources that a task can request for its pod.
 
 :::info Alternative Astro Hybrid setup
 
--- a/astro/configure-deployment-resources.md
+++ b/astro/configure-deployment-resources.md
@@ -161,7 +161,7 @@ The CPU and memory quotas also determine the **Max Pod Size**, which is the maxi
 
 :::caution
 
-The largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**. Because tasks run in a single Pod, your tasks can't request resources that exceed the **Max Pod Size**, even if your quota is larger.
+For Deployments running on dedicated clusters, the largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**. Because tasks run in a single Pod, your tasks can't request resources that exceed the **Max Pod Size**, even if your quota is larger.
 
 For example, if your Deployment has a CPU quota of 150vCPU and a memory quota of 300GiB, your **Max Pod Size** might only be 12 vCPU and 24GiB RAM. If you try to run a task that requests 20vCPU, the task won't run even though it's within your quotas.
 
--- a/astro/configure-deployment-resources.md
+++ b/astro/configure-deployment-resources.md
@@ -148,10 +148,11 @@ Set safeguards by configuring default Pod limits and requests from the Cloud UI.
 2. Click **Resource quotas**.
 3. Configure the following values:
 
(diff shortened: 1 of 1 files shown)


astro/configure-deployment-resources.md
```diff
@@ -148,12 +148,18 @@ Set safeguards by configuring default Pod limits and requests from the Cloud UI.
...
+    - **Default Pod Size (Memory)**: The amount of memory that your tasks run with if no memory usage is specified in their Pod configuration.
 
-Your CPU quota and memory quota determine your **Max Pod Size**, which is the maximum amount of resources that a task can request for its Pod. If the CPU and memory quotas you specify exceed the limits of Astro's infrastructure, your **Max Pod Size** is determined by the size of the Astro-hosted infrastructure running your tasks.
+:::warning
+
+If the max pod size is equal to the deployment's quota, and a task requests the max pod size, only one task will run at a time, because that task is consuming the deployment's entire quota.
+
+:::
+
+The maximum quotas for CPU and memory are currently 400 vCPU and 800 GiB RAM, respectively. The CPU and memory quotas determine the **Max Pod Size**, which is the maximum amount of resources that a task can request for its pod. The maximum max pod size is currently allows for 27 vCPU and 54 GiB RAM for Standard Clusters and 12 vCPU and 24 GiB RAM for Dedicated Clusters.
```
jwitz, line 163:
```suggestion
:::caution

The largest possible **Max Pod Size** has fewer resources than the largest possible CPU and memory quotas. If your CPU and memory quotas exceed your **Max Pod Size**, the **Max Pod Size** becomes the maximum amount of resources you can request in a single task. 

For example, if you have a CPU quota of 150 and a memory quota of 200, your **Max Pod Size** will be 12 vCPU and 24GiB RAM. If you try to run a task that requests 20vCPU, the task won't run even though it's within your quotas.

:::

```

jwitz (changes requested):
@RNHTTR I think I'm having trouble understanding the current warning. I tried adding a different warning based on what my understanding of the issue is, can you confirm if that's correct? 

In its current state, I don't know why we'd need to communicate the warning about one task running at a time. That seems like expected behavior if you have a proper understanding of CPU and memory quotas.

astro/configure-deployment-resources.md
```diff
@@ -148,12 +148,18 @@ Set safeguards by configuring default Pod limits and requests from the Cloud UI.
...
+    - **Quota (Memory)**: The maximum amount of memory for all currently running Pods on your Deployment.
+    - **Default Pod Size (CPU)**: The amount of CPUs that your tasks run with if no CPU usage is specified in their Pod configuration.
+    - **Default Pod Size (Memory)**: The amount of memory that your tasks run with if no memory usage is specified in their Pod configuration.
 
-Your CPU quota and memory quota determine your **Max Pod Size**, which is the maximum amount of resources that a task can request for its Pod. If the CPU and memory quotas you specify exceed the limits of Astro's infrastructure, your **Max Pod Size** is determined by the size of the Astro-hosted infrastructure running your tasks.
+:::warning
+
+If the max pod size is equal to the deployment's quota, and a task requests the max pod size, only one task will run at a time, because that task is consuming the deployment's entire quota.
+
+:::
```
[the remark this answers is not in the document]
RNHTTR: > we can inclide it out of a note.

Do you mean like an info block? I think that'd be fine. I just think it ought to be highlighted.

astro/configure-deployment-resources.md
RNHTTR: cpu to memory is always 1:2, so this example would have to be "150 vCPU and a memory quota of 300GiB". My concern here is the difference in max pod size between Standard/Dedicated (and your point about "what if these maximums change?")

astro/configure-deployment-resources.md
```diff
@@ -148,12 +148,18 @@ Set safeguards by configuring default Pod limits and requests from the Cloud UI.
...
 3. Configure the following values:
 
-    - **CPU quota**: The maximum amount of CPU for all currently running Pods on your Deployment. 
-    - **Memory Quota**: The maximum amount of memory for all currently running Pods on your Deployment.
-    - **Default Pod Size** > **CPU**: The amount of CPUs that your tasks run with if no CPU usage is specified in their Pod configuration.
-    - **Default Pod Size** > **Memory**: The amount of memory that your tasks run with if no memory usage is specified in their Pod configuration.
+    - **Quota (CPU)**: The maximum amount of CPU for all currently running Pods on your Deployment. 
+    - **Quota (Memory)**: The maximum amount of memory for all currently running Pods on your Deployment.
+    - **Default Pod Size (CPU)**: The amount of CPUs that your tasks run with if no CPU usage is specified in their Pod configuration.
+    - **Default Pod Size (Memory)**: The amount of memory that your tasks run with if no memory usage is specified in their Pod configuration.
```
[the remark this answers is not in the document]
jwitz: The field says "CPU Quota" instead of "Quota (CPU)". Bold styling in our docs mean "You could ctrl+f

Would an ellipses work better? Updated my suggestion, let me know what you think

astro/configure-deployment-resources.md
jwitz: Updated my suggestion to afford the possibility that the values could change in the future. I think it's okay to use these values as examples even though the exact ratios might change over time. The main point is "Quotas can be greater than max pod size"

astro/configure-deployment-resources.md
```diff
@@ -148,12 +148,18 @@ Set safeguards by configuring default Pod limits and requests from the Cloud UI.
...
 
-Your CPU quota and memory quota determine your **Max Pod Size**, which is the maximum amount of resources that a task can request for its Pod. If the CPU and memory quotas you specify exceed the limits of Astro's infrastructure, your **Max Pod Size** is determined by the size of the Astro-hosted infrastructure running your tasks.
+:::warning
+
+If the max pod size is equal to the deployment's quota, and a task requests the max pod size, only one task will run at a time, because that task is consuming the deployment's entire quota.
+
+:::
+
+The CPU and memory quotas determine the **Max Pod Size**, which is the maximum amount of resources that a task can request for its pod.
```
jwitz, line 162:
```suggestion
Your CPU and memory quotas determine how many tasks you can run at once on your Deployment. For example, if a task requests the full amount of your CPU and memory quotas, then your Deployment can run only that task until it completes.

The CPU and memory quotas also determine the **Max Pod Size**, which is the maximum amount of resources that a task can request for its pod.
```

astro/configure-deployment-resources.md
```diff
@@ -148,12 +148,22 @@ Set safeguards by configuring default Pod limits and requests from the Cloud UI.
...
+    - **Quota (Memory)**: The maximum amount of memory for all currently running Pods on your Deployment.
+    - **Default Pod Size (CPU)**: The amount of CPUs that your tasks run with if no CPU usage is specified in their Pod configuration.
+    - **Default Pod Size (Memory)**: The amount of memory that your tasks run with if no memory usage is specified in their Pod configuration.
 
-Your CPU quota and memory quota determine your **Max Pod Size**, which is the maximum amount of resources that a task can request for its Pod. If the CPU and memory quotas you specify exceed the limits of Astro's infrastructure, your **Max Pod Size** is determined by the size of the Astro-hosted infrastructure running your tasks.
+
+Your CPU and memory quotas determine how many tasks can run at once on your Deployment. For example, if your Deployment has a CPU quota of 3vCPU and a memory quota of 6GiB, and a task requests this amount, then your Deployment can run only that task until it completes.
+
+The CPU and memory quotas also determine the **Max Pod Size**, which is the maximum amount of resources that a task can request.
```
jwitz, line 159:
```suggestion
The CPU and memory quotas also determine the **Max Pod Size**, which is the maximum amount of resources that a task can request.

```

jwitz (commented):
I tried addressing the issue your customer faced in the description after the bulleted list @RNHTTR, as well as including a caution about exceeding max pod size when your quotas are larger. LMK what you think

lzdanski (commented):
That value configuration step is really tricky. I don't see a very clear "right" answer, but we might have some other tradeoff options.

astro/configure-deployment-resources.md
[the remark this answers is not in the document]
RNHTTR: Nested bullets would be fine. I think a colon would be better than an ellipses or `>` character

astro/configure-deployment-resources.md
```diff
@@ -148,12 +148,24 @@ Set safeguards by configuring default Pod limits and requests from the Cloud UI.
...
+
+:::caution
+
+The largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**. Because tasks can only run in a single Pod, your tasks can't request resources that exceed the **Max Pod Size**.
```
jwitz, line 164:
```suggestion
The largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**. Because tasks run in a single Pod, your tasks can't request resources that exceed the **Max Pod Size**, even if your quota is larger.
```

jwitz:
@RNHTTR good to merge?

RNHTTR:
> @RNHTTR good to merge?

> For example, if your Deployment has a CPU quota of 150vCPU and a memory quota of 300GiB, your **Max Pod Size** might only be 12 vCPU and 24GiB RAM. If you try to run a task that requests 20vCPU, the task won't run even though it's within your quotas.

This is true only in Dedicated, because the max pod size is larger (I think 27vCPU and 54 Mem).

> The largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**

I don't think this needs to be a warning. The warning I want to convey is "if your max pod size is equal to your quota, and you request a task with resources equal to max pod size/quota, then that is the only task that will run until it completes".

It's not a bug, but it is a non-obvious footgun.

astro/configure-deployment-resources.md
```diff
@@ -148,12 +148,24 @@ Set safeguards by configuring default Pod limits and requests from the Cloud UI.
...
 
-Your CPU quota and memory quota determine your **Max Pod Size**, which is the maximum amount of resources that a task can request for its Pod. If the CPU and memory quotas you specify exceed the limits of Astro's infrastructure, your **Max Pod Size** is determined by the size of the Astro-hosted infrastructure running your tasks.
+
+Your CPU and memory quotas determine how many tasks can run at once on your Deployment. For example, if your Deployment has a CPU quota of 3vCPU and a memory quota of 6GiB, and a task requests this amount, then your Deployment can run only that task until it completes.
+
+The CPU and memory quotas also determine the **Max Pod Size**, which is the maximum amount of resources that a task can request.
+
+:::caution
+
+The largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**. Because tasks run in a single Pod, your tasks can't request resources that exceed the **Max Pod Size**, even if your quota is larger.
```
jwitz, line 164:
```suggestion
For Deployments running on dedicated clusters, the largest possible CPU and memory quotas can exceed the largest possible **Max Pod Size**. Because tasks run in a single Pod, your tasks can't request resources that exceed the **Max Pod Size**, even if your quota is larger.
```

jwitz:
@RNHTTR Got it, think I covered the 1 task warning with:

> Your CPU and memory quotas determine how many tasks can run at once on your Deployment. For example, if your Deployment has a CPU quota of 3vCPU and a memory quota of 6GiB, and a task requests this amount, then your Deployment can run only that task until it completes.

Also added a caveat that the warning is only for dedicated clusters.

lzdanski (approved):
@jwitz - edits look good in preview but @RNHTTR might need to give final approval on your content changes. Looks like y'all had a discussion but I couldn't tell if you had come to a consensus about the wording.

jwitz (approved):
Going to approve and merge because it looks like all remaining comments have been addressed! Thanks for bearing with us @RNHTTR

RNHTTR:
sorry for the slow communication. looks good!

Issue #2708, opened 2023-08-17
Missed details about adding private certificate


afastronomer:
### Issue Type

Inaccurate, misleading, or out-of-date information, Missing content for an existing feature

### Links to Affected Docs

https://docs.astronomer.io/astro/develop-project#add-a-ca-certificate-to-an-astro-runtime-image

### The Issue or Context

Important details are missed in instruction.

manmeetkaur:
latest update on this: https://github.com/astronomer/docs/pull/2709#issuecomment-1717871436

lzdanski:
Discussion reopened: https://astronomer.slack.com/archives/CL44PA4DB/p1697558617305879

PR #2709, opened 2023-08-17
Update develop-project.md

afastronomer:
correction for CA cert adding part

### DIFF
--- a/astro/develop-project.md
+++ b/astro/develop-project.md
@@ -579,10 +579,14 @@ Your Astro project can now utilize Python packages from your private PyPi index.
 
 ### Add a CA certificate to an Astro Runtime image
 
-If you need your Astro Deployment to communicate securely with a remote service using a certificate signed by an untrusted or internal certificate authority (CA), you need to add the CA certificate to the trust store inside your Astro project's Docker image. It happens when Security/Network team implements Man-in-the-Middle (MitM) approach for traffic monitoring.
+If you need your Astro Deployment to communicate securely with a remote service using a certificate signed by an untrusted or internal certificate authority (CA), you need to add the CA certificate to the trust store inside your Astro project's Docker image. You might need to use this approach if your Security or Network team implements a Man-in-the-Middle (MitM) approach for traffic monitoring.
 
-1. Ask Security/Network team for internal root certificate.
-2. Create `pip.conf` file in your project folder with following content:
+#### Prerequisites
+- The internal root certificate. You might need to request this from your Security or Network team
+- Your Astro project `dockerfile`. See [Install packages from Private repository](/astro/develop-project#install-python-packages-from-private-sources) for an example.
+
+#### Step 1: Add trusted hosts
+1. Create `pip.conf` file in your project folder with following content:
 
     ```shell
     [global]
@@ -592,11 +596,16 @@ If you need your Astro Deployment to communicate securely with a remote service 
                    pip.astronomer.io
                    # more hosts can be added if needed
     ```
-   
-3. Use Dockerfile similar to [Install packages from Private repository](/astro/develop-project#install-python-packages-from-private-sources).
-4. Add the following entry below `FROM stage1 AS stage2` line in our example or any other location before installing Python packages:
+
+#### Step 2: Configure your Dockerfile
+
+1. Specify the Runtime version in your Dockerfile, and make sure you use Runtime **base** image. The base image tag is formatted as `quay.io/astronomer/astro-runtime:8.8.0-base`, and not the regular image at `quay.io/astronomer/astro-runtime:8.8.0`.
+
+2. Include the following code in your Astro project `dockerfile`. This entry adds the CA to the trust store and needs to be declared before installing Python packages. 
 
     ```docker
+    FROM stage1 AS stage2
+    # Add CA to trust store
     USER root
     COPY <internal-ca.crt> /usr/local/share/ca-certificates/<internal-ca.crt>
     RUN update-ca-certificates
@@ -604,15 +613,25 @@ If you need your Astro Deployment to communicate securely with a remote service 
     USER astro
     ```
 
-    :::info
+    The following example uses code from the [Install packages from Private repository](/astro/develop-project#install-python-packages-from-private-sources) example, and shows where to add the CA to the trust store following, between the `FROM stage1 AS stage2` line and before the `# Install Python Packages`:
 
-   Make sure you use Runtime **base** image i.e. quay.io/astronomer/astro-runtime:8.8.0-base but not regular one quay.io/astronomer/astro-runtime:8.8.0 .
+    ```docker
+    FROM stage1 AS stage2
+    # Add CA to trust store
+    USER root
+    COPY <internal-ca.crt> /usr/local/share/ca-certificates/<internal-ca.crt>
+    RUN update-ca-certificates
+    COPY pip.conf pip.conf
+    USER astro
+    # Install Python Packages
+    ARG PIP_EXTRA_INDEX_URL
+    ENV PIP_EXTRA_INDEX_URL=${PIP_EXTRA_INDEX_URL}
+    COPY requirements.txt .
+    RUN pip install --no-cache-dir -q -r requirements.txt
+    ```
+3. (Optional) Add additional `COPY` statements before the `RUN update-ca-certificates` stanza for each CA certificate your organization uses for external access.
 
-   :::
-    
-5. Optional. Add additional `COPY` statements before the `RUN update-ca-certificates` stanza for each CA certificate your organization is using for external access.
-
-6. [Restart your local environment](/astro/develop-project.md#restart-your-local-environment) or deploy to Astro. See [Deploy code](/astro/deploy-code.md).
+4. [Restart your local environment](/astro/develop-project.md#restart-your-local-environment) or deploy to Astro. See [Deploy code](/astro/deploy-code.md).
 
 ## Unsupported project configurations 
 
--- a/astro/develop-project.md
+++ b/astro/develop-project.md
@@ -350,23 +350,6 @@ RUN ls
 
 This is supported both on Astro and in the context of local development.
 
-### Add a CA certificate to an Astro Runtime image
-
-If you need your Astro Deployment to communicate securely with a remote service using a certificate signed by an untrusted or internal certificate authority (CA), you need to add the CA certificate to the trust store inside your Astro project's Docker image.
-
-1. In your Astro project `Dockerfile`, add the following entry below the existing `FROM` statement which specifies your Astro Runtime image version:
-
-    ```docker
-    USER root
-    COPY <internal-ca.crt> /usr/local/share/ca-certificates/<your-company-name>/
-    RUN update-ca-certificates
-    USER astro
-    ```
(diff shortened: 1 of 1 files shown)


afastronomer:
#2708

astro/develop-project.md
```diff
@@ -593,6 +576,44 @@ Your Astro project can now utilize Python packages from your private PyPi index.
...
+    [global]
+    trusted-host = pypi.python.org
+                   pypi.org
+                   files.pythonhosted.org
+                   pip.astronomer.io
+                   # more hosts can be added if needed
+    ```
+   
+3. Use Dockerfile similar to [Install packages from Private repository](/astro/develop-project#install-python-packages-from-private-sources).
+4. Add the following entry below `FROM stage1 AS stage2` line in our example or any other location before installing Python packages:
```
lzdanski, line 597:
```suggestion
4. Add the following entry to the file before installing Python packages, such as after the `FROM stage1 AS stage2` line:
```

astro/develop-project.md
```diff
@@ -593,6 +576,44 @@ Your Astro project can now utilize Python packages from your private PyPi index.
...
+    USER root
+    COPY <internal-ca.crt> /usr/local/share/ca-certificates/<internal-ca.crt>
+    RUN update-ca-certificates
+    COPY pip.conf pip.conf
+    USER astro
+    ```
+
+    :::info
+
+   Make sure you use Runtime **base** image i.e. quay.io/astronomer/astro-runtime:8.8.0-base but not regular one quay.io/astronomer/astro-runtime:8.8.0 .
```
lzdanski, line 609:
```suggestion
   Make sure you use Runtime **base** image in your Dockerfile.  The base image tag is formatted as `quay.io/astronomer/astro-runtime:8.8.0-base`, and not the regular image at `quay.io/astronomer/astro-runtime:8.8.0`.
```
Will this need to be updated to runtime 9.0 when it gets released?

astro/develop-project.md
```diff
@@ -593,6 +576,44 @@ Your Astro project can now utilize Python packages from your private PyPi index.
...
+    ```shell
+    [global]
+    trusted-host = pypi.python.org
+                   pypi.org
+                   files.pythonhosted.org
+                   pip.astronomer.io
+                   # more hosts can be added if needed
+    ```
+   
+3. Use Dockerfile similar to [Install packages from Private repository](/astro/develop-project#install-python-packages-from-private-sources).
```
lzdanski, line 596: I think this step might need to be either removed if we make the Dockerfile a pre-req _or_ we can add a code example to this step. Again, just want to prevent users from having to hop between two different procedures once they get into the middle of a process.

lzdanski (commented):
@afastronomer - This is really helpful detail! There's one structural thing that I think might need a change. 

Right now, there's a couple steps in this flow that require users to bounce between processes or reach out to another team, which can sometimes be discouraging. What do you think about adding those things as pre-reqs or including a short Dockerfile code example? (The comments do a better job of showing what I've got in mind.)

astro/develop-project.md
```diff
@@ -593,6 +576,44 @@ Your Astro project can now utilize Python packages from your private PyPi index.
...
+### Add a CA certificate to an Astro Runtime image
+
+If you need your Astro Deployment to communicate securely with a remote service using a certificate signed by an untrusted or internal certificate authority (CA), you need to add the CA certificate to the trust store inside your Astro project's Docker image. It happens when Security/Network team implements Man-in-the-Middle (MitM) approach for traffic monitoring.
+
+1. Ask Security/Network team for internal root certificate.
+2. Create `pip.conf` file in your project folder with following content:
+
+    ```shell
+    [global]
+    trusted-host = pypi.python.org
```
jedcunningham, line 589: The point of adding the customers CA is so you we don't have to turn off validation. We should remove this part.

afastronomer: for my recent case only adding certs and "trusted hosts" correction allowed to pull python libraries. any of these solutions didn't help by itself.

astro/develop-project.md
```diff
@@ -593,6 +576,63 @@ Your Astro project can now utilize Python packages from your private PyPi index.
...
+    FROM stage1 AS stage2
+    # Add CA to trust store
+    USER root
+    COPY <internal-ca.crt> /usr/local/share/ca-certificates/<internal-ca.crt>
+    RUN update-ca-certificates
+    COPY pip.conf pip.conf
+    USER astro
+    ```
+
+    The following example uses code from the [Install packages from Private repository](/astro/develop-project#install-python-packages-from-private-sources) example, and shows where to add the CA to the trust store following, between the `FROM stage1 AS stage2` line and before the `# Install Python Packages`:
```
lzdanski, line 616: This second stanza of example code might be redundant.

lzdanski:
Definitely out of scope to make any big changes to this page structure, but wow there's a lot of content on this page related to some other conversations about the docs going on with the team. I kicked off a slack conversation about [how this overall page is organized](https://astronomer.slack.com/archives/C015V2JFKT5/p1692638612313439), so we can maybe keep the scope of this PR focused on just clarifying the CA cert steps. I can see how my changes/suggestions might send us down major scope creep, and I don't want that to happen.  

@afastronomer - how do these changes look?

astro/develop-project.md
rob-1126: Is it because the trusted ca is added in the user layers and there are ONBUILD directives that run before that that attempt to install from pip and fail? I'd still expect the same with modfiying the pip config as well though.

astro/develop-project.md
```diff
@@ -593,6 +576,63 @@ Your Astro project can now utilize Python packages from your private PyPi index.
...
 </Tabs>
 
+
+### Add a CA certificate to an Astro Runtime image
+
+If you need your Astro Deployment to communicate securely with a remote service using a certificate signed by an untrusted or internal certificate authority (CA), you need to add the CA certificate to the trust store inside your Astro project's Docker image. You might need to use this approach if your Security or Network team implements a Man-in-the-Middle (MitM) approach for traffic monitoring.
+
+#### Prerequisites
+- The internal root certificate. You might need to request this from your Security or Network team
+- Your Astro project `dockerfile`. See [Install packages from Private repository](/astro/develop-project#install-python-packages-from-private-sources) for an example.
```
jwitz, line 586:
```suggestion
#### Prerequisites

- The internal root certificate. You might need to request this from your security or network team
- An Astro project.
```

astro/develop-project.md
```diff
@@ -593,6 +576,63 @@ Your Astro project can now utilize Python packages from your private PyPi index.
...
+    trusted-host = pypi.python.org
+                   pypi.org
+                   files.pythonhosted.org
+                   pip.astronomer.io
+                   # more hosts can be added if needed
+    ```
+
+#### Step 2: Configure your Dockerfile
+
+1. Specify the Runtime version in your Dockerfile, and make sure you use Runtime **base** image. The base image tag is formatted as `quay.io/astronomer/astro-runtime:8.8.0-base`, and not the regular image at `quay.io/astronomer/astro-runtime:8.8.0`.
```
jwitz, line 602:
```suggestion
1. If you aren't already using it, specify the `base` distribution of your Astro Runtime version in your Dockerfile. The base image tag is formatted as `quay.io/astronomer/astro-runtime:<version>-base`.
```

astro/develop-project.md
```diff
@@ -593,6 +576,63 @@ Your Astro project can now utilize Python packages from your private PyPi index.
...
+    RUN update-ca-certificates
+    COPY pip.conf pip.conf
+    USER astro
+    # Install Python Packages
+    ARG PIP_EXTRA_INDEX_URL
+    ENV PIP_EXTRA_INDEX_URL=${PIP_EXTRA_INDEX_URL}
+    COPY requirements.txt .
+    RUN pip install --no-cache-dir -q -r requirements.txt
+    ```
+3. (Optional) Add additional `COPY` statements before the `RUN update-ca-certificates` stanza for each CA certificate your organization uses for external access.
```
jwitz, line 632:
```suggestion
3. (Optional) Add additional `COPY` statements before the `RUN update-ca-certificates` command for each CA certificate your organization uses.
```

jwitz (changes requested):
Overall LGTM, but holding off on merging until we resolve Jed's discussion on Step 1

astro/develop-project.md
jedcunningham: Okay, I think we are getting somewhere now.

https://pip.pypa.io/en/latest/topics/https-certificates/

"This does not use the system certificate store but, instead, uses a bundled CA certificate store from certifi."

Basically, just adding the CA to the system isn't enough, and turning off validation with trusted-host is why it actually "worked".

3.10 can potentially use truststore and a feature flag, but that won't work as ~half of our supported runtime images aren't on 3.10 anyways. We could document the truststore path for those on 3.10, but we will need an alternative for the others.

Overwriting `pip.conf` also isn't the safest thing to do, as we do have some stuff in our image already that's important.

Probably the best option is setting the `PIP_CERT` env var.

astro/develop-project.md
```diff
@@ -593,6 +576,49 @@ Your Astro project can now utilize Python packages from your private PyPi index.
...
+                   files.pythonhosted.org
+                   pip.astronomer.io
+                   # more hosts can be added if needed
+    ```
+
+#### Step 2: Configure your Dockerfile
+
+1. If you aren't already using it, specify the `base` distribution of your Astro Runtime version in your Dockerfile. The base image tag is formatted as `quay.io/astronomer/astro-runtime:<version>-base`.
+
+2. Include the following code in your Astro project `dockerfile`. This entry adds the CA to the trust store and needs to be executed before installing Python packages.
```
jwitz, line 606:
```suggestion
2. Include the following code in your Astro project `dockerfile`. This entry adds the CA to the trust store and needs to be executed before any other build-time arguments
```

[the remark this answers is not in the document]
jwitz: Added one suggestion - Python package installation is only one possible Docker image use case, I think we need to clarify that this needs to go before any and all Dockerfile commands

astro/develop-project.md
lzdanski: @jedcunningham - it sounds like this changes the procedure here quite a bit. I dug around some examples of using `PIP_CERT` with other tools like [Cloud Foundry](https://docs.cloudfoundry.org/buildpacks/python/index.html#private-repos-pip). So I've got some questions about how it affects the content of this PR. 

It sounds like now the process is:

Step 1: Add `PIP_CERT` to your environment variables and have it point to the internal CA certificate

Step 2: Configure your Dockerfile to add the CA to the trust store
- Does `PIP_CERT` need to be added to the[ Dockerfile specifically?] (https://docs.astronomer.io/astro/environment-variables#set-environment-variables-in-your-dockerfile) or can it be added through the Cloud UI? 
- Is adding a certificate in plaintext in the dockerfile a recommended action? Are there any security concerns?
- If it should be added to the Dockerfile, does it need to be added so that it's executed before or after the CA is added to the trust store in Step 2?

lzdanski:
Closing PR due to inactivity. We can reopen and address with increase in customer feedback or when there's time for technical discussion on what specific docs change is needed here.

Issue #2721, opened 2023-08-22
Add allowlisting info to Hosted docs


jwitz:
### Issue Type

Missing content for an existing feature

### Links to Affected Docs

https://docs.astronomer.io/astro/create-first-DAG

### The Issue or Context

It would be worth adding a troubleshooting note for users who might need to get Astro domains allowlisted on their organization's private network.

PR #2722, opened 2023-08-22
Add note about allowlisting for Hosted

jwitz:
Resolves https://github.com/astronomer/docs/issues/2721


lzdanski:
I know we discussed this yesterday and there's a bunch of tradeoffs for where this information goes, and this change looks good to me. In future, I would much rather it live in a troubleshooting page, reference, or some other place because the note format takes up a lot of real estate between landing on the page and taking the first action in step one. But I don't know we have content right now where it would be more ideally placed than this!

lzdanski (approved):
Might still need other reviewers to take a look, but the formatting and preview lgtm

PR #2723, opened 2023-08-22
Update Runtime release table

jwitz:
Remove redundant releases/ separate channel tables

### DIFF
--- a/astro/runtime-version-lifecycle-policy.md
+++ b/astro/runtime-version-lifecycle-policy.md
@@ -77,7 +77,7 @@ Occasionally, Astronomer might deviate from the defined response policy and back
 
 <!--- Version-specific -->
 
-The following table contains the exact lifecycle for each published version of Astro Runtime. These timelines are based on the LTS and Stable release channel maintenance policies.
+The following table contains the exact lifecycle for each published version of Astro Runtime. These timelines are based on the LTS and Stable [release channel maintenance policies](#release-channels).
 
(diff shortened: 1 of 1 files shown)


astro/runtime-version-lifecycle-policy.md
```diff
@@ -79,25 +79,13 @@ Occasionally, Astronomer might deviate from the defined response policy and back
 
 The following table contains the exact lifecycle for each published version of Astro Runtime. These timelines are based on the LTS and Stable release channel maintenance policies.
```
lzdanski, line 80: The only consideration I might recommend is linking back up to the definitions of the release channels or moving those definitions here. I'm most familiar with stable and end of life/ deprecated as terms about maintenance, so I always have to scroll around to remind myself what our terms mean. Doing a quick search, we've got about 60 total links to this page, 16 of them are to the main page, and 11 are to this section. We don't have any links going specifically to the Release Channels section where the terms are defined, so they might get overlooked.

Alternatively, could move this section to before **Security** so it's right next to the two sections about LTS/Stable.
```suggestion
The following table contains the exact lifecycle for each published version of Astro Runtime. These timelines are based on the LTS and Stable [release channel maintenance policies](#release-channels).
```

lzdanski (commented):
Formatting and preview LGTM but a quick question about positioning.

jedcunningham (approved):

harels (approved):

PR #2727, opened 2023-08-23
[Tiny PR] Learn -> connections -> Redshift

manmeetkaur:
Fix the following:

- line numbers
- remove redundant step
- add note to link to Authorize Deployment page (cc: @fritz-astronomer )

### DIFF
--- a/learn/connections/redshift.md
+++ b/learn/connections/redshift.md
@@ -45,7 +45,7 @@ DB user credentials can be used to establish a connection to an Amazon Redshift 
 
 :::tip
 
-On Astro, you can use your [Deployment's workload identity](https://docs.astronomer.io/astro/authorize-deployments-to-your-cloud) to authorize connections to Redshift, thus avoiding the need to embed secrets in an Airflow connection.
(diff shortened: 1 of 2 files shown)


learn/connections/redshift.md
```diff
@@ -43,6 +43,12 @@ DB user credentials can be used to establish a connection to an Amazon Redshift
 - User
 - Password
 
+:::tip
+
+On Astro, you can use your [Deployment's workload identity](https://docs.astronomer.io/astro/authorize-deployments-to-your-cloud) to authorize connections to Redshift, thus avoiding the need to embed secrets in an Airflow connection.
```
jwitz, line 48:
```suggestion
If you're an Astro user, Astronomer recommends using workload identity to authorize to your Deployments to Redshift. This eliminates the need to specify secrets in your Airflow connections. See [Authorize Deployments to your cloud](https://docs.astronomer.io/astro/authorize-deployments-to-your-cloud).
```

jwitz: This tip should be at the end of the overview section instead of in the tabs. That way it's more visible to users and they don't need to have decided on an auth method before they see it.

jwitz (approved):
LGTM, I would just move the note to the overview and out of the tabs!

manmeetkaur:
> LGTM, I would just move the note to the overview and out of the tabs!

I did not want to do that, because then each method had a different message explaining the benefit when using it. But I think it is more cleaner if you also think so. will make the change.

PR #2752, opened 2023-08-28
fix `.md` in release notes link

fritz-astronomer:


lzdanski (approved):
Tested fixed link in preview

PR #2765, opened 2023-08-30
Add app detail to Azure AD SCIM setup

chomipi88:


lzdanski (commented):
Fixed some small style/grammar bits.

lzdanski (approved):
Changes look good in preview. I don't have access to the 3rd party dashboard to confirm accuracy, but PR creator does, so moving forward to merge.

PR #2775, opened 2023-08-31
Simplify sidebar titles

paolaperaza:
Hey Team -- Some of these sidebar titles were driving me crazy in how they were inconsistent and too many words so they wrapped around when I thought they didn't need to. A few things:

- I shortened titles everywhere I could, often dropping a verb or a mention of "Astro" that I didn't think we needed
- I tried to update the description of docs to have the full `Apache Airflow` reference, but I didn't catch it all
- I lengthened the true title where needed

I did NOT modify the sidebar file -- so whoever reviews this, double check the build & that file to make sure I didn't mess anything up. I'll let one of you guys merge this whenever you approve it as I won't have time to follow up 😊


astro/import-export-connections-variables.md
```diff
@@ -1,8 +1,8 @@
 ---
-sidebar_label: 'Import and export connections and variables'
-title: 'Import and export connections and variables'
+sidebar_label: 'Import and export'
```
jwitz, line 2:
```suggestion
sidebar_label: 'Import and export connections and variables'
```
Idk what to do about this yet, but we need some sort of noun to root the sidebar labels.

astro/manage-connections-variables.md
```diff
@@ -1,6 +1,6 @@
 ---
-sidebar_label: 'Manage connections and variables'
-title: 'Manage Airflow connections and variables'
+sidebar_label: 'Manage'
+title: 'Manage Apache Airflow connections and variables'
```
jwitz, line 3:
```suggestion
title: 'Manage Airflow connections and variables'
```
Other doc in this menu didn't have Apache, so updating this for consistency

astro/secrets-backend.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: 'Configure an external secrets backend on Astro'
-sidebar_label: 'Configure a secrets backend'
+sidebar_label: 'Secrets backend'
```
jwitz, line 3:
```suggestion
sidebar_label: 'Secrets backends'
```

astro/upgrade-runtime.md
```diff
@@ -1,6 +1,6 @@
 ---
-sidebar_label: 'Upgrade Astro Runtime'
-title: 'Upgrade Astro Runtime'
+sidebar_label: 'Upgrade Airflow'
+title: 'Upgrade Apache Airflow and Astro Runtime'
```
jwitz, line 3:
```suggestion
sidebar_label: 'Upgrade Astro Runtime'
title: 'Upgrade Astro Runtime'
```
Hm let's wait to give this one some more consideration, "Apache Airflow and Astro Runtime" makes it sound like you can upgrade one/ the other/ both when it's just one operation.

astro/audit-logs.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: 'Export Astro audit logs'
 id: audit-logs
-sidebar_label: Export audit logs
+sidebar_label: Audit logs
```
jwitz, line 4:
```suggestion
sidebar_label: Export Audit logs
```
I agree with removing instances of "Configure", "Manage" etc from the sidebars because those words don't really mean anything in this context, but I still think we should use verbs when the operation is more specific. This will be more important when we have multiple docs dealing with a single feature

astro/manage-connections-variables.md
```diff
@@ -1,6 +1,6 @@
 ---
-sidebar_label: 'Manage connections and variables'
-title: 'Manage Airflow connections and variables'
+sidebar_label: 'Manage'
```
jwitz, line 2:
```suggestion
sidebar_label: 'Manage connections and variables'
```
Same comment here as for the other

jwitz (approved):
I think we can derive a new style item from this PR: Sidebar titles should not contain verbs like "Configure" or "Manage". Additionally, we should only include a verb if the doc covers a single, stepped process, rather than a suite of options.

jwitz:
I only reverted a couple of sidebar items that seemed like they deserved more consideration, such as "Upgrade Astro Runtime"

PR #2781, opened 2023-09-01
fix wording for VPC peering steps

### LINKED ISSUE #2777: Contradicting VPC peering instructions
The instructions here for accepting the peering request seem wrong. I thought the idea was that we had the role then the customer would not have to accept it themselves

 https://github.com/astronomer/docs/blob/05759d7c3227b0a357f40a86a032efbbaefd38a9/astro/connect-aws.md?plain=1#L98

afastronomer:
Resolves #2777

### DIFF
--- a/astro/connect-aws.md
+++ b/astro/connect-aws.md
@@ -95,7 +95,7 @@ To create a VPC peering connection between an Astro VPC and an AWS VPC, you must
     
     Astronomer support will initiate a peering request and create the routing table entries in the Astro VPC.
 
-6. Wait for Astronomer support to send you the Astro VPC CIDR and VPC Peering ID. Then, the owner of the accepter VPC needs to [create back routes entries](https://docs.aws.amazon.com/vpc/latest/userguide/WorkWithRouteTables.html#AddRemoveRoutes) in accepter VPC, with Astro VPC CIDR as Destination and provided VPC Peering ID as Target.
+6. Wait for Astronomer support to send you the Astro VPC CIDR and VPC Peering ID. Then, the owner of the accepter VPC needs to [create back routes entries](https://docs.aws.amazon.com/vpc/latest/userguide/WorkWithRouteTables.html#AddRemoveRoutes) in the accepter VPC, with the Astro VPC CIDR as the **Destination** and the provided VPC Peering ID as the **Target**.
 
 7. (Optional) Delete the stack that you created. This will delete the temporary assumable role.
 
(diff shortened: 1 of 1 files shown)


astro/connect-aws.md
```diff
@@ -95,7 +95,7 @@ To create a VPC peering connection between an Astro VPC and an AWS VPC, you must
     
     Astronomer support will initiate a peering request and create the routing table entries in the Astro VPC.
 
-6. Wait for Astronomer support to send you the Astro VPC CIDR. Then, the owner of the target VPC needs to [accept the peering request](https://docs.aws.amazon.com/vpc/latest/peering/accept-vpc-peering-connection.html) and [create the routing table entries](https://docs.aws.amazon.com/vpc/latest/userguide/WorkWithRouteTables.html#AddRemoveRoutes) in the target VPC.
+6. Wait for Astronomer support to send you the Astro VPC CIDR and VPC Peering ID. Then, the owner of the accepter VPC needs to [create back routes entries](https://docs.aws.amazon.com/vpc/latest/userguide/WorkWithRouteTables.html#AddRemoveRoutes) in accepter VPC, with Astro VPC CIDR as Destination and provided VPC Peering ID as Target.
```
lzdanski, line 98:
```suggestion
6. Wait for Astronomer support to send you the Astro VPC CIDR and VPC Peering ID. Then, the owner of the accepter VPC needs to [create back routes entries](https://docs.aws.amazon.com/vpc/latest/userguide/WorkWithRouteTables.html#AddRemoveRoutes) in the accepter VPC, with the Astro VPC CIDR as the **Destination** and the provided VPC Peering ID as the **Target**.
```

lzdanski: @afastronomer or @stuart23 Making an assumption that the Destination and the Target are buttons or UI entry fields. If they're parameters or values defined in `code`, then they'll need to be code formatted instead. They do need to have a decorator in addition to capitalization. Can you confirm that these changes are accurate?

lzdanski (commented):
Question about some wording and text formatting for style.

astro/connect-aws.md
afastronomer: Those are column names. Now it looks better.

jwitz (changes requested):
Thanks for this update @afastronomer , just one question about the phrase "accepter VPC"!

jwitz (approved):
Confirmed with @afastronomer on Slack that "external" works here!

Issue #2780, opened 2023-09-01
[bug] broken link to connect clusters


chomipi88:
### Issue Type

Typo or broken link

### Links to Affected Docs

https://docs.astronomer.io/astro/authorize-deployments-to-your-cloud#prerequisites

### The Issue or Context

I believe the link should be pointing to: https://docs.astronomer.io/astro/category/connect-to-external-resources

lzdanski:
Link confirmed working in prod.

PR #2782, opened 2023-09-01
Fix broken link

lzdanski:
https://github.com/astronomer/docs/issues/2780


lzdanski:
Tested link in the preview and confirmed that it works.

Issue #2787, opened 2023-09-04
Administration sidebar


manmeetkaur:
**Current Setup:**

- Administration:

  - User access
    - Manage Organization users
    - Manage Workspace users
    - Configure Teams
    - Setup SSO
    - Setup SCIM provisioning
    - Manage domains
    - User permissions reference

  - Organizations
    - Organization API tokens

  - Workspaces
    - Configure Workspaces
    - Workspace API tokens

  - Clusters
    - Create a dedicated Astro cluster
    - Resource reference
    - Connect to external resources
      - AWS
      - Azure
      - GCP

  - Astro Hybrid
    - Hybrid overview
    - Install Astro Hybrid
      - AWS
      - Azure
      - GCP
    - Manage Hybrid Clusters
    - Hybrid cluster settings reference

  - Billing

**Problems:** 

1. No rule for menu titles? User access was earlier part of Organizations and I proposed to move it out as it applies to all the other Astro components as well. It seems now other topics can also come out similarly
2. Some menu just like organization and now deployment will have only 1 sub-topic

**Proposed setup:**

- Administration

  - User access (same no change)
    - Manage Organization users
    - Manage Workspace users
    - Configure Teams
    - Setup SSO
    - Setup SCIM provisioning
    - Manage domains
    - User permissions reference
  
  - Manage Astro components
    - Create a dedicated cluster
    - Manage Workspaces (should have create/update/delete workspace settings? what about the other overview part )
    - Transfer a Deployment
    - Resource reference
  
  - Networking & Authentication
    - Connect to external resources
      - AWS
      - Azure
      - GCP
    - Authorize Deployment
    - API keys and tokens (this will also mean that as part of consolidation of the Automation menu, deployment api keys and workspace and org api tokens come under one umbrella ?) 

  - Billing
 
  - Astro Hybrid (no change)
    - Hybrid overview
    - Install Astro Hybrid
      - AWS
      - Azure
      - GCP
    - Manage Hybrid Clusters
    - Hybrid cluster settings reference

jwitz:
I think this is a good discussion to have! My thoughts: The "Manage Astro components" menu feels like a step backwards. We're going to end up with a lot of docs covering a lot of different topics all in one menu, making it harder for a user to discover what they need. 

When structuring the sidebars, I try to think about what kind of questions a user is asking as they're looking through the menu. I like the current structure because I'm making a bet that users are thinking "I need to do something on my Workspace" instead of "I need to manage an Astro component".

[...]

PR #2789, opened 2023-09-04
move auth to adminstration

manmeetkaur:
Related to the discussion about moving Auth items to Admin menu. This is a temp change. But rather than having too many changes in our docs and confuse users, I am inclined to first have a discussion on this https://github.com/astronomer/docs/issues/2787 and then do the shuffle. it not a big change, we just have to agree on the pattern.


astro/configure-deployment-resources.md
```diff
@@ -171,22 +171,6 @@ To give your tasks more or less resources, change the worker type in the task's
 
 :::
 
-## Transfer a Deployment to another Workspace
```
jwitz, line 174: We need a redirect for this

astro/transfer-a-deployment.md
```diff
@@ -0,0 +1,25 @@
+---
+sidebar_label: 'Transfer a Deployment'
+title: 'Transfer a Deployment to another Workspace'
+id: transfer-a-deployment
+description: "Transfer a Deployment to another Workspace in the same cluster"
```
jwitz, line 5:
```suggestion
description: "Transfer a Deployment to another Workspace in the same cluster."
```

sidebarsAstro.js
```diff
@@ -93,8 +93,6 @@ module.exports = {
         "api-keys",
         "environment-variables",
         "secrets-backend",
```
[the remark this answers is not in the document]
manmeetkaur: manage DAGs is actually not something that you do in a Deployment and more of a Observability feature. So I think it better fits there. nvm. I reverted.

sidebarsAstro.js
```diff
@@ -204,6 +201,13 @@ module.exports = {
             },
           ],
         },
+        {
+          type: "category",
+          label: "Organizations",
```
jwitz, line 206:
```suggestion
          label: "Organization",
```
I'm fine either way but I feel like the vast vast majority of users will only ever have one Org, so it's okay to refer to it like that

jwitz (approved):
LGTM!

Issue #2753, opened 2023-08-28
Add how to run a different Python version with Runtime


lzdanski:
### Issue Type

Missing content for an existing feature

### Links to Affected Docs

- https://docs.astronomer.io/astro/develop-project

Link updates to new section to add to:

- https://docs.astronomer.io/astro/runtime-image-architecture#python-versioning
- https://docs.astronomer.io/astro/runtime-release-notes#new-python-version-distributions-of-astro-runtime
- https://docs.astronomer.io/astro/upgrade-runtime#upgrade-to-python-311)

### The Issue or Context

We have release notes with info about selecting a particular Runtime image depending on the users desired Python version, but there aren't specific instructions for updating an Astro project to pull a particular Runtime image for a specific Python version.

### Fix or Proposed Change

- Add a step by step process and code examples for updating which Runtime image to use for a particular version of Python. 
- Link out to runtime architecture doc for more conceptual info

### Additional Notes

- https://astronomer.slack.com/archives/C30GY05NG/p1693240076276849
- This change might need to use the [Runtime Reorg PR as a base PR](https://github.com/astronomer/docs/pull/2726/files#diff-627b62dc63204c64512df3bb118bdbaf7c91436f030a787cff02c6d1ad7884f5R83)

PR #2793, opened 2023-09-05
Add topic for Python distributions

jwitz:
Resolves https://github.com/astronomer/docs/issues/2753

I made the title generic because we have upcoming non-Python distributions of Runtime, and I'm inclined to group those setups together instead of create a separate topic for each distribution.

### DIFF
--- a/astro/develop-project.md
+++ b/astro/develop-project.md
@@ -350,6 +350,22 @@ RUN ls
 
 This is supported both on Astro and in the context of local development.
 
+### Use an alternative Astro Runtime distribution
+
+Starting with Astro Runtime 9, each version of Astro Runtime has a separate distribution for each currently supported Python version. Use an alternative Python distribution if any of your dependencies require a Python version other than the [default Runtime Python version](runtime-image-architecture.md#python-versioning).
+
+To use a Python distribution, update the first line in your Astro project `Dockerfile` to reference the required distribution:
+
+```text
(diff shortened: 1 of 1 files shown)


astro/develop-project.md
```diff
@@ -350,6 +350,22 @@ RUN ls
 
 This is supported both on Astro and in the context of local development.
 
+### Use an alternative Astro Runtime distribution
+
+Starting with Astro Runtime 9, each version of Astro Runtime has a separate distribution for each currently supported Python version. Use an alternative Python distribution if any of your dependencies require a Python version other than the [default Runtime Python version](runtime-image-architecture.md#python-versioning).
+
+To use a Python distribution, update the first line in your Astro project `Dockerfile` to reference the required distribution:
```
lzdanski, line 357:
```suggestion
To use a specific Python distribution, update the first line in your Astro project `Dockerfile` to reference the required distribution:
```

astro/develop-project.md
```diff
@@ -350,6 +350,22 @@ RUN ls
...
+
+```text
+FROM quay.io/astronomer/astro-runtime:<runtime-version>-python-<python-version>
+```
+
+For example, to use Python 3.10 with Astro Runtime version 9.0.0, you would update the first line of your Dockerfile to the following:
```
lzdanski, line 363:
```suggestion
For example, to use Python 3.10 with Astro Runtime version 9.0.0, you update the first line of your Dockerfile to the following:
```

Issue #2769, opened 2023-08-30
CI CD Warning


erdos2n:
### Issue Type

Missing content for an existing feature

### Links to Affected Docs

https://docs.astronomer.io/astro/ci-cd-templates/template-overview

### The Issue or Context

There is an edge case to account for:

When authenticating using an API Key, the `astro cli` will rewrite the `.astro/config.yaml` file. If an enterprise's GitHub (or Version Control) has the repositories sharing a node, then all of the deployments will have access to this file. In the case of GitHub it's located here: `/home/gitrunner/.astro/config.yaml`

So repoA and repoB can both then push to the deploymentA if the config file is set for A at that time.

```mermaid
flowchart LR;
subgraph github
    repoA
    repoB[repo B\nhas not set API keys]
end
subgraph astro
    deploymentA
    deploymentB
end

repoA-->config.yaml
repoB-->config.yaml
config.yaml-->deploymentA
```

### Fix or Proposed Change

Emphasize and add warning about this potential issue

### Additional Notes

ZD Ticket: 28420

erdos2n:
This issue occurs when customer uses a self-hosted github runner and not `ubuntu-latest`.

erdos2n:
@jwitz , PR #2805 , covers some of this, but I think a warning stating that `self-hosted runners` should be handled differently is important. There are still cases where adding a deployment-id may not suffice. For example, a customer copies a workflow file verbatim from another repo in their organization.

PR #2805, opened 2023-09-06
add deployment id to CI/CD scripts

sunkickr:
add deployment id to docs to CI/CD docs to avoid encouraging users into a situation where they may accidentally deploy to the wrong deployment. Additionally specifying the deployment id make it so users avoid the issue outlined here: https://github.com/astronomer/docs/issues/2769

@jwitz should we add copy to warn users that if they don't specify a deployment id that run the risk of deploying to the wrong deployment and/or what is specified in the issue mentioned above?


astro/ci-cd-templates/github-actions.md
```diff
@@ -70,6 +70,7 @@ To automate code deploys to a single Deployment using [GitHub Actions](https://g
         steps:
         - name: Deploy to Astro
           uses: astronomer/deploy-action@v0.2
+          deployment-id: <deployment-id>
```
jwitz, line 73:
```suggestion
          deployment-id: <your-deployment-id>
```

astro/ci-cd-templates/github-actions.md
```diff
@@ -169,6 +170,7 @@ If your Astro project requires additional build-time arguments to build an image
         - name: Build image
           uses: docker/build-push-action@v2
           with:
+            deployment-id: <deployment-id>
```
jwitz, line 173:
```suggestion
            deployment-id: <your-deployment-id>
```

astro/ci-cd-templates/github-actions.md
```diff
@@ -181,6 +183,7 @@ If your Astro project requires additional build-time arguments to build an image
         - name: Deploy to Astro
           uses: astronomer/deploy-action@v0.2
           with:
+            deployment-id: <deployment-id>
```
jwitz, line 186:
```suggestion
            deployment-id: <your-deployment-id>
```

astro/ci-cd-templates/github-actions.md
```diff
@@ -222,6 +225,7 @@ If your Astro project requires additional build-time arguments to build an image
...
```
jwitz, line 228:
```suggestion
            deployment-id: <your-deployment-id>
```

jwitz (approved):

jwitz (commented):
This LGTM, as long as our teams are consistently recommending these templates then I think we don't have to include an extra warning on top of improving the template. WDYT @erdos2n ? Would this have been enough to cover your edge case, or are folks still likely to build their own templates without the docs and miss this?

erdos2n:
> 

Hey @jwitz , sorry I'm just seeing this. This will cover most of the edge case, but having some sort of documentation around self-hosted runners is necessary. For now, at least a warning that this process only applies when the deployment process isn't generating a persisted `config.yaml` file.

lzdanski:
@jwitz - do you want to add the additional CI/CD warning to this PR, or would you like to assign the issue to me and I can add it so we can move to merge this PR?

jwitz:
@erdos2n Sorry should have gotten your review on this before we merged, but we added a new note about self-hosted runners to the prerequisites! 
Screenshot 2023-09-20 at 9 47 37 AM

PR #2801, opened 2023-09-06
Reorg env vars

manmeetkaur:
Resolves #2658 

### New content
1. Read new page in preview
2. review content in file
- [ ] astro/env-vars-astro.md
- [ ] astro/environment-variables.md

### Changed content
1. Read new page in preview
2. review content in file
Content that originally appeared in astro/environment-variables.md has now been moved to
- [ ]  learn/airflow-variables.md
- [ ] astro/cli/develop-project.md
- [ ] astro/manage-connections-variables.md

See project plan for specific sections that have been moved and why.

### Link changes/minor edits (16 files)
1. Test results in preview
- [ ] astro/airflow-api.md
- [ ] astro/cli/astro-deployment-variable-update.md
- [ ] astro/cloud-ide/configure-project-environment.md
- [ ] astro/deployment-metrics.md
- [ ] astro/deployment-settings.md
- [ ] astro/import-export-connections-variables.md
- [ ] astro/migration-partials/deployments.mdx
- [ ] astro/platform-variables.md
- [ ] astro/release-notes.md
- [ ] astro/runtime-image-architecture.md
- [ ] astro/secrets-backend.md
- [ ] astro/secrets-management.md
- [ ] astro/set-up-data-lineage.md
- [ ] astro/view-logs.md
- [ ] sidebarsAstro.js
- [ ] static/_redirects

## Additional notes

[Project doc for this is in Notion.](https://www.notion.so/astronomerio/Doc-improvement-Environment-variable-cbca5b8ecf344f72ae78bd7e4fc24dd1?pvs=4)

Note: Replaced the content of `environment-variables.md` with `env-vars-overview.md` and removed `env-vars-overview.md`, to reduce the number of redirects required.

PR #2054, opened 2023-03-24
Use secret environment variable values in Kubernetes-based tasks

jwitz:
Resolves https://github.com/astronomer/docs/issues/1930
Resolves https://github.com/astronomer/docs/issues/2025

Software content is still WIP, but Astro content is ready for review

Issue #2025, opened 2023-03-17
Referencing Kubernetes Secrets in Astro


wolfier:
**Request type:**

- [x] **New document:** Add the `docs-content` label to this issue.
- [ ] **Addition to an existing document**: Add the `enhancement` label to this issue.

**What is the purpose of the document/enhancement? Select all that apply:**

- [ ] Document a new feature for an upcoming release. Add the appropriate release version label to this issue.
- [x] Information for an existing feature that isn't yet documented.
- [ ] Additional considerations/ reference material for an existing document.

**Problem/feature that requires documentation:**

I would like a documentation section describing how environment variables marked as secrets are stored as Kubernetes Secrets and specifically in a secret named `env-secrets`.

Users who execute `KubernetesPodOperators` or task instances with `KubernetesExecutor` may want to mount an environment variable that is marked secrets in their tasks. In Astro, Environment Variables marked as secrets are stored as a Kubernetes secret named `env-secrets`. Informing the customer of this Kubernetes Secret enables them to mount a secret on their tasks. In Airflow, it would look like this.

```
import pendulum
from airflow.kubernetes.secret import Secret

from airflow.models import DAG
from airflow.providers.cncf.kubernetes.operators.kubernetes_pod import KubernetesPodOperator
from airflow.configuration import conf

with DAG(
        dag_id='test-kube-pod-secret',
        start_date=pendulum.datetime(2022, 1, 1, tz="UTC"),
        end_date=pendulum.datetime(2022, 1, 5, tz="UTC"),
        schedule_interval="@once",
) as dag:

    secret_env = Secret(deploy_type="env", deploy_target="MY_SECRET", secret="env-secrets", key="MY_SECRET")

    namespace = conf.get("kubernetes", "NAMESPACE")

    k = KubernetesPodOperator(
        namespace=namespace,
        image="ubuntu:16.04",
        cmds=["bash", "-cx"],
        arguments=["echo $MY_SECRET && sleep 150"],
        name="test-name",
        task_id="test-task",
        get_logs=True,
        secrets=[secret_env],
    )
```

For `KubernetesPodOperators`, users can inject the environment variable through the `env_var` but the value is exposed in the templated view. For tasks instance queued by the `KubernetesExecutor`, users can only reference existing Kubernetes objects and to reference a secret, they would need to know the Kubernetes Secret name.

**Are you the right person to review a draft of the document for technical accuracy? If not, who is?**

I can help review the documentation!

[...]

PR #2814, opened 2023-09-08
Update K8s executor secret env var section

### LINKED ISSUE #2770: Workers on K8s Executor already have Astro env variables available
### Issue Type

### Links to Affected Docs

https://docs.astronomer.io/astro/kubernetes-executor#mount-secret-environment-variables-to-worker-pods

### The Issue or Context

Astro environment variables are already available to Worker Pods in K8s Executor. [Slack thread](https://astronomer.slack.com/archives/C01MCEU9EV7/p1692895708008429).

### Fix or Proposed Change

_No response_

### Additional Notes

_No response_

### Required Reviewers

_No response_

manmeetkaur:
Resolves #2770 

Add a simple method to access env variables in addition to the complex method that requires `pod_override` in only special cases. Also change how env vars are stored on astro.

### DIFF
--- a/astro/environment-variables.md
+++ b/astro/environment-variables.md
@@ -66,13 +66,12 @@ After you set an environment variable key, only the environment variable value c
 
 ### How environment variables are stored on Astro
 
-Non-secret environment variables set in the Cloud UI are stored in a database that is managed by Astronomer and hosted in the Astro control plane. When you configure a secret environment variable in the Cloud UI, the following methodology is used:
+Non-secret environment variables set in the Cloud UI are stored in a Hashicorp Vault secrets manager hosted in Astro control plane. These environment variables are available to your Astro Deployment’s Kubernetes namespace as a Kubernetes secret.
 
-- Astro generates a manifest that defines a Kubernetes secret, named `env-secrets`, that contains your variable's key and value.
-- Astro applies this manifest to your Deployment's namespace.
-- After the manifest is applied, the key and value of your environment variable are stored in a managed [etcd cluster](https://etcd.io/) at rest within Astro.
-
-This process occurs every time you update the environment variable's key or value. To use a secret environment variable value in a KubernetesPodOperator, you need to mount the value from the Astro kubernetes secret to your Kubernetes Pod. See [Use secret environment variables with the KuberernetesPodOperator](kubernetespodoperator.md#use-secret-environment-variables-with-the-kubernetespodoperator).
+To use these environment variables in an Astro Deployment:
+- In a regular PythonOperator or Python code, you can use `os.environ` method.
+- If you can’t use Python or are using a pre-defined code that expects specific keys for environment variables, you need to mount the secret environment variables using BaseOperator’s `executor_config` by providing a `pod_override`. See [Use secret environment variables in worker Pods](kubernetes-executor.md#use-secret-environment-variables-in-worker-pods).
+- If you need to use these secret environment variables in a KubernetesPodOperator, see [Use secret environment variables with KubernetesPodOperator](kubernetespodoperator.md#use-secret-environment-variables-with-the-kubernetespodoperator).
 
 :::caution
 
--- a/astro/environment-variables.md
+++ b/astro/environment-variables.md
@@ -72,10 +72,7 @@ Non-secret environment variables set in the Cloud UI are stored in a database th
 - Astro applies this manifest to your Deployment's namespace.
 - After the manifest is applied, the key and value of your environment variable are stored in a managed [etcd cluster](https://etcd.io/) at rest within Astro.
 
-This process occurs every time you update the environment variable's key or value. To use a secret environment variable value in a task running on the Kubernetes executor or the KubernetesPodOperator, you need to mount the value from the Astro kubernetes secret to your Kubernetes Pod. See:
-
-- [Mount secret environment variables to worker pods](kubernetes-executor.md#mount-secret-environment-variables-to-worker-pods)
-- [Use secret environment variables with the KubernetesPodOperator](kubernetespodoperator.md#use-secret-environment-variables-with-the-kubernetespodoperator)
+This process occurs every time you update the environment variable's key or value. To use a secret environment variable value in a KubernetesPodOperator, you need to mount the value from the Astro kubernetes secret to your Kubernetes Pod. See [Use secret environment variables with the KuberernetesPodOperator](kubernetespodoperator.md#use-secret-environment-variables-with-the-kubernetespodoperator).
 
 :::caution
 
--- a/astro/kubernetes-executor.md
+++ b/astro/kubernetes-executor.md
@@ -94,9 +94,89 @@ When this DAG runs, it launches a Kubernetes Pod with exactly 0.5m of CPU and 10
 
 ## Use secret environment variables in worker Pods
 
-In Astro, [environment variables](environment-variables.md) marked as secrets are stored in a Kubernetes secret called `env-secrets`. These environment variables are already available to your worker Pods and can be accessed in your tasks just like any other environment variable, `os.environ[<your-secret-env-var>]` or `os.getenv(<your-secret-env-var>, None)`. 
+In Astro, [environment variables](environment-variables.md) marked as secrets are stored in a Kubernetes secret called `env-secrets`. These environment variables are already available to your worker Pods and can be accessed in your tasks just like any other environment variable. For example, you can use `os.environ[<your-secret-env-var>]` or `os.getenv(<your-secret-env-var>, None)` in your Python code. 
 
-However, in KubernetesPodOperator, you need to read the value from `env-secrets` as described in [Run the KubernetesPodOperator on Astro](kubernetespodoperator.md#use-secret-environment-variables-with-the-kubernetespodoperator).
+However, if you can’t use Python or are using a pre-defined code that expects specific keys for environment variables, you need to mount the secret environment variables by pulling the value from `env-secrets` and mount it to the Pod running your task as a new Kubernetes Secret.
+
+1. Add the following import to your DAG file:
+   
+    ```python
+    from airflow.kubernetes.secret import Secret
+    ```
+
+2. Define a Kubernetes `Secret` in your DAG instantiation using the following format:
+
+    ```python
+    secret_env = Secret(deploy_type="env", deploy_target="<VARIABLE_KEY>", secret="env-secrets", key="<VARIABLE_KEY>")
+    namespace = conf.get("kubernetes", "NAMESPACE")
+    ```
+
+3. Specify the `Secret` in the `secret_key_ref` section of your `pod_override` configuration.
+
+4. In the task where you want to use the secret value, add the following task-level argument:
+
+    ```python
+    op_kwargs={
+            "env_name": secret_env.deploy_target
+    },
+    ```
+
+5. In the executable for the task, call the secret value using `os.environ[env_name]`.
+
+In the following example, a secret named `MY_SECRET` is pulled from `env-secrets` and printed to logs.
+ 
+```python
+import pendulum
+from kubernetes.client import models as k8s
+
+from airflow.configuration import conf
+from airflow.kubernetes.secret import Secret
+from airflow.models import DAG
+from airflow.providers.cncf.kubernetes.operators.kubernetes_pod import KubernetesPodOperator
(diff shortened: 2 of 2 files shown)


jwitz (changes requested):
@manmeetkaur The KE and KPO have two different use cases, so I don't see this as redundant content. What if we just rewrite this section to have the standard `os.environ` usage? We don't have that documented anywhere else right now- to be honest we should have it documented for Celery too, or in your [env var docs](https://github.com/astronomer/docs/pull/2801), but a simple topic for KE would work for the time being.

manmeetkaur:
> @manmeetkaur The KE and KPO have two different use cases, so I don't see this as redundant content. What if we just rewrite this section to have the standard `os.environ` usage? We don't have that documented anywhere else right now- to be honest we should have it documented for Celery too, or in your [env var docs](https://github.com/astronomer/docs/pull/2801), but a simple topic for KE would work for the time being.

@jwitz These are two same use-cases: using secrets in KPO; irrespective of if it is in KE or CE, it works the exactly the same. And this example is incorrect. Why do you want to keep the exact same thing in two places  when we have one generic doc for KPO  which is irrespective of the executor used ? 

`os.environ()` is the use in PythonOperator and is the standard across the python world, nothing special. **if** it needs to be added it should be in LEARN.

jwitz:
@manmeetkaur These are not the exact same thing, because one uses the KPO and one doesn't. We're about to remove instructions that, even if they were inaccurate, only applied to the KE. Where will KE users go to find this information now? We shouldn't remove the topic until we have a plan to replace it with the correct content. In short, this PR is only half of the necessary fix. 

I wrote this content based on https://github.com/astronomer/docs/issues/2025 and https://support.astronomer.io/hc/en-us/articles/14968015052179. Tagging @wolfier on this PR to review the content and see if we have evolved our recommendations on this since.

jwitz:
@manmeetkaur These setups are not the exact same thing, because one uses the KPO and one doesn't. We're about to remove instructions that, even if they were inaccurate, only applied to the KE. Where will KE users go to find this information now? We shouldn't remove the topic until we have a plan to replace it with the correct content. In short, this PR is only half of the necessary fix.

I wrote this content based on https://github.com/astronomer/docs/issues/2025 and https://support.astronomer.io/hc/en-us/articles/14968015052179. Tagging @wolfier on this PR to review the content and see if we have evolved our recommendations on this since.

manmeetkaur:
@jwitz I am not sure you understand, let me try and summarize one last time: 

- the section being removed is **does not work**
- **no special instructions are required** for KPO on KE
- `os.environ` **does not work** with KPO for secret env variables set on Astro, what works with KPO is already in the KPO doc.

manmeetkaur:
@jwitz the content in this doc that we are changing and https://github.com/astronomer/docs/issues/2025 is totally different. :)

jwitz:
I understand all of that @manmeetkaur. What I am saying is that we previously identified a need to tell users how to use environment variable secrets with the Kubernetes Executor. That need is no longer being met. Even if the content is just "You can use `os.environ` without further configuration", we should say that here instead of deleting the topic entirely. 

As for your second point, the exact topic you're removing was added in the PR I linked - the `pod_override` code sample is included in the linked ZenDesk article (see https://github.com/astronomer/docs/pull/2054/files#diff-7524d728cea0757a4c90237e05e0910cb40015d397ff24a3905e40f8dbe2fee8R188). Please take a closer look and, if needed, notify the team that our KB articles need to be updated.

wolfier:
> the section being removed does not work

@manmeetkaur Can you be more specific about what is not working? Does the DAG fail? Does the pod requested by the Kubernetes Executor not have the secret specific in the `pod_override`? Please provide a recreation of the behaviour.

If the [documentation](https://docs.astronomer.io/astro/kubernetes-executor#mount-secret-environment-variables-to-worker-pods) and [Knowledge Base article](https://support.astronomer.io/hc/en-us/articles/14968015052179) are wrong, we will need to correct it.

---

The section that is slated for removal in this PR is different than the code that enables mounting a secret for a pod defined by the KubernetesPodOperator.

To mount a secret in for a pod defined by the KubernetesPodOperator, a user can specify the following as shown in our documentation on [Use secret environment variables with the KubernetesPodOperator](https://docs.astronomer.io/astro/kubernetespodoperator?tab=PrivateRegistry#use-secret-environment-variables-with-the-kubernetespodoperator).

```python
    secret_env = Secret(deploy_type="env", deploy_target="MY_SECRET", secret="env-secrets", key="MY_SECRET")
    namespace = conf.get("kubernetes", "NAMESPACE")

    k = KubernetesPodOperator(
        namespace=namespace,
        image="ubuntu:16.04",
        cmds=["bash", "-cx"],
        arguments=["echo $MY_SECRET && sleep 150"],
        name="test-name",
        task_id="test-task",
        get_logs=True,
        secrets=[secret_env],
    )
```

To mount a secret in for a pod requested by the KubernetesExecutor, a user can specify the following as shown in our documentation on [Mount secret environment variables to worker Pods](https://docs.astronomer.io/astro/kubernetes-executor#mount-secret-environment-variables-to-worker-pods).

```python
    secret_env = Secret(deploy_type="env", deploy_target="MY_SECRET", secret="env-secrets", key="MY_SECRET")
    namespace = conf.get("kubernetes", "NAMESPACE")

    p = PythonOperator(
        python_callable=print_env,
        op_kwargs={
            "env_name": secret_env.deploy_target
        },
        task_id='test-py-env',
        executor_config={
            "pod_override": k8s.V1Pod(
                spec=k8s.V1PodSpec(
                    containers=[
                        k8s.V1Container(
                            name="base",
                            env=[
                                k8s.V1EnvVar(
                                    name=secret_env.deploy_target,
                                    value_from=k8s.V1EnvVarSource(
                                        secret_key_ref=k8s.V1SecretKeySelector(name=secret_env.secret,
                                                                               key=secret_env.key)
                                    ),
                                )
                            ],
                        )
                    ]
                )
            ),
        }
    )
```

Please indicate clearly why these two use cases are the same thus only needed one documentation section? The goal is the same but the method to do so is different for different environments. One distinct difference is one uses `executor_config` and `pod_override` and the other uses `KubernetesPodOperator` parameters.

manmeetkaur:
@wolfier Apologies for the delay on this. On K8s executor, adding the pod override makes the task go in a hung state ; it just stays in Queued state. 

Let me run my tests on diff deployments and get the logs ready and then we can chat.

manmeetkaur:
@wolfier Here is the update from my testing with logs:

DAG:
```
from airflow import DAG
from kubernetes.client import models as k8s
from airflow.providers.cncf.kubernetes.operators.kubernetes_pod import KubernetesPodOperator
from airflow.utils.dates import days_ago
from airflow.configuration import conf
from airflow.operators.python import PythonOperator
from airflow.kubernetes.secret import Secret


def print_env(env_name):
    import os
    print(os.environ[env_name])


def print_env_k8s(secrets):
    print(secrets)


with DAG(
    'env_var_k8s',
    schedule_interval='@daily',
    start_date=days_ago(0),
    tags=['example'],
) as dag:

    secret_env = Secret(deploy_type="env", deploy_target="OWNER", secret="env-secrets", key="OWNER")
    namespace = conf.get("kubernetes", "NAMESPACE")

    p_override = PythonOperator(
        python_callable=print_env,
        op_kwargs={
            "env_name": secret_env.deploy_target
        },
        task_id='test-py-env-override',
        executor_config={
            "pod_override": k8s.V1Pod(
                spec=k8s.V1PodSpec(
                    containers=[
                        k8s.V1Container(
                            name="base",
                            env=[
                                k8s.V1EnvVar(
                                    name=secret_env.deploy_target,
                                    value_from=k8s.V1EnvVarSource(
                                        secret_key_ref=k8s.V1SecretKeySelector(name=secret_env.secret,
                                                                               key=secret_env.key)
                                    ),
                                )
                            ],
                        )
                    ]
                )
            ),
        }
    )

    regular_p = PythonOperator(
        task_id='test-py-env-regular',
        python_callable=print_env,
        op_kwargs={
            "env_name": "OWNER"
        }
    )

    k8s_p = PythonOperator(
        task_id='test-py-env-k8s',
        python_callable=print_env_k8s,
        op_kwargs={
            "secrets": secret_env
        }
    )

    k_env_secrets = KubernetesPodOperator(
        namespace=namespace,
        image="ubuntu:16.04",
        cmds=["bash", "-cx"],
        arguments=["echo $OWNER"],
        name="test-name",
        task_id="test-kpo",
        get_logs=True,
        secrets=[secret_env],
    )

    regular_k = KubernetesPodOperator(
        namespace=namespace,
        image="ubuntu:16.04",
        cmds=["bash", "-cx"],
        arguments=["echo $OWNER"],
        name="test-name",
        task_id="test-kpo-regular",
        get_logs=True,
    )

    p_override >> regular_p >> k8s_p
    k_env_secrets >> regular_k

```

**Problem Statement:** Environment variables, secret or otherwise, are already available to Worker Pods on K8s executor - no override necessary.

**Setup**: 
Create two environment variables
```
ENVIRONMENT: K8s
OWNER(secret): Manmeet
```

**Test Cases as seen in the DAG**:

1. Read a secret environment variable in PythonOperator using Pod override. (task `p_override`): This was stuck/queued in all my earlier tests in one deployment - it succeeded this test - i was using a diff runtime and a new deployment. Which is weird.
2. Read a secret environment variable in a regular pythonic way. using `os.environ`. (task `regular_p`): **_Success_**
3. Read a secret environment variable in a KPO using `k8s.secret` (task: `k_env_secrets`): **Success**
4. Read a secret environment variable in a KPO using regular bash way (task: `regular_k`): **FAILED**

**Airflow URL:** https://cosmicenergy.astronomer.run/ddjjz9lz/dags/env_var_k8s/grid?dag_run_id=manual__2023-09-13T08%3A22%3A27.943020%2B00%3A00&task_id=test-kpo-regular&tab=logs
**Deployment Variables**: https://cloud.astronomer.io/cl81pguhg3568354syfo88o4kjz/deployments/clm7m9l6l474338fdw0gdjjz9lz/variables

**Summary**: 

If you notice, test case #1 uses `pod_override` to achieve the same thing that test case #2 can do using the regular pythonic way. My point is, if we do not have to do anything special to read secret environment variables in a PythonOperator, then why use `pod_override` and complicate for users. Users can seamlessly just use the environment variables using `os.environ` irrespective of the executor. 

Basically there are no "special instructions" for using environment variables in PythonOperator, where as for KPO there are and they are documented in the KPO page. 

I hope this makes sense @wolfier . Let me know if you want to have a call on this and discuss.

manmeetkaur:
As per discussion between @wolfier and @jwitz putting this on hold, because the code itself is correct and works. But we need better understanding of the benefits of using the current way mentioned in our docs and why not use `os.environ`.

wolfier:
> Read a secret environment variable in PythonOperator using Pod override. (task p_override): This was stuck/queued in all my earlier tests in one deployment - it succeeded this test - i was using a diff runtime and a new deployment. Which is weird.

In the above test case, a user could use `os.environ` to achieve the same outcome in a PythonOperator. Because a PythonOperator is flexible, it allows the use of the `os` python package. 

Lets imagine an operator where the user cannot execute arbitrary python code. An operator that expects certain environment variables to exist in the execution environment. Even a python operator whose `python_callable` is predefined and not alterable by the user could fit this scenario.

The only way for the operator to execute successfully is to mount the expected environment variable to the Kubernetes worker pod. If the desired value already exists as a secret environment variable mounted to the deployment via the platform Variables view, the `pod_override` method seems suitable for this case. On the other hand, the `os.environ` method would not work because the operator is not under the control of the user.

manmeetkaur:
@wolfier Agreed on the use-case given. But isn't the current section gives the wrong impression to the user that the only way they can use it by using `pod_override`. What you have specified is a use-case that should be in our docs, but we should clarify that `os.environ` will work and in case you are not able to edit or extend or inherit the pre-defined python method, then this is an option ? I just feel this is a special case as you described, and then we can add details to clarify it. WDYT ?

manmeetkaur:
@jwitz I have updated this PR from the technical perspective. Even though I still feel the use-case of "mount" is also applicable to Celery and not just K8s and it should be not here on the executor page. But you can take that decision where to put it. Also updated the section in environment variables as per the [discussion](https://astronomer.slack.com/archives/C01UA03TPV1/p1695623526387419?thread_ts=1694550962.494019&cid=C01UA03TPV1) in Slack.

PR #2827, opened 2023-09-11
Add v9 to table

lzdanski:
Hotfix to update runtime version in reference table


lzdanski:
Confirmed build in preview. Merging.

PR #2830, opened 2023-09-12
[Tiny PR] Add another note about CLI upgrading to the Create your first DAG doc

TJaniF:
There was another user running into dag-only deploy not working because they had an old CLI version (they reached out over intercom). This is just a proposal, maybe it makes more sense to have a note at the end of Step 3 that covers the deployment step generally?


jwitz (approved):
Great troubleshooting tip!

PR #2836, opened 2023-09-13
[Tiny PR] Use ROOT user before `ssh` setup and package install

manmeetkaur:
This issue was identified by a prospect when he was getting the following error by following our instructions.
```
fatal: Could not read from remote repository
```

Resolved this by adding USER root before the `ssh` setup and switching back to USER astro after installation. This has resolved the error for me and for the user as well. Relevant slack threads in #intercom: https://astronomer.slack.com/archives/C30GY05NG/p1694587804294879

Other related Slack thread: https://astronomer.slack.com/archives/C01MCEU9EV7/p1694436254383719

This was tested on local as well as Astro Deployment.

### DIFF
--- a/astro/develop-project.md
+++ b/astro/develop-project.md
@@ -453,16 +453,24 @@ This example assumes that the name of each of your Python packages is identical 
 
 1. Optional. Copy and save any existing build steps in your `Dockerfile`.
 
-2. Add `openssh-client` and `git` to your `packages.txt` file.
+2. Add the following to your `packages.txt` file:
+
+    ```bash
+    openssh-client
+    git
+    ```
 
 3. In your Dockerfile, add the following instructions:
 
     ```docker
+    USER root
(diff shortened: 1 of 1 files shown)


lzdanski (approved):
Changes LGTM in preview - I don't have intercom access, so I can't read through the original user problem, but the description in our Astro slack channel of the problem and why this solution works makes sense to me. Up to your discretion if you'd like a technical reviewer in addition to my content review!

manmeetkaur:
@lzdanski Thanks for reviewing quickly. I noticed that I missed the step explanation and also an important note for more available github fingerprints, so asked for a review again. coz there will be some english mistakes!

Also, the code works, as I shared the same and the prospect was able to resolve the error. So you can trust me :)

lzdanski (approved):
Preview LGTM

Issue #2834, opened 2023-09-12
Astro CLI v1.19.2 Release Notes


sunkickr:
### New Feature Description

Release Notes on v1.19.2 Astro CLI Patch

Bug Fixes:

- Fixes an issue where upgrade test command would tell a user it was testing an upgrade to a latest for upgrades to all custom images
- Fixes an issue where the upgrade test command was not creating an HTML report for the DAG test
- The HTML report will now tell you have many DAGs passed and Failed the DAG test. You need to replace the parse file for these changes to take effect 
- Fixes Deployment URL shown after deploy
- Fixes issue where you couldn't create a hosted dedicated deployment with a deployment file and API token 

### Required Reviewers

@jwitz

### Release Date

September 13th, 2023

PR #2837, opened 2023-09-13
CLI 1.19.2 release notes

jwitz:
Resolves https://github.com/astronomer/docs/issues/2834

Looks like upgrade-test is stabilizing and we're using it a lot with customers, so also added that into our upgrade Runtime guide.


jwitz:
@lzdanski A previous writer thought we should always include the upgrade considerations before the core setup. What do you think about instead keeping those at the bottom of the doc so that users don't have to scroll loads just to see how to upgrade?

lzdanski:
@lzdanski I think that makes sense and then have relative linking so that users can skip down to where they need would work!

So like 

```
## Upgrade considerations
... <existing content>

Each version of Astro runtime introduces new features, behaviors, and bug fixes. Before upgrading to a particular version, check to see if there are any version specific considerations you need to account for:

- Runtime version 9 (link to v9 at bottom)
-  etc
```

And then at the end

```
## Runtime Version considerations

### Version 9 
<add existing content>

```

lzdanski:
@jwitz  I think that makes sense and then have relative linking so that users can skip down to where they need would work!

So like 

```
## Upgrade considerations
... <existing content>

Each version of Astro runtime introduces new features, behaviors, and bug fixes. Before upgrading to a particular version, check to see if there are any version specific considerations you need to account for:

- Runtime version 9 (link to v9 at bottom)
-  etc
```

And then at the end

```
## Runtime Version considerations

### Version 9 
<add existing content>

```

sunkickr (approved):

sunkickr (approved):

astro/upgrade-runtime.md
```diff
@@ -13,6 +13,102 @@ import {siteVariables} from '@site/src/versions';
...
+
+:::info
+
+If you're only upgrading a local Airflow environment, you don't need an Astro Deployment and you can skip steps 7-8.
+
+:::
+
+## Step 1: Review upgrade considerations
+
+Astro upgrades can include breaking changes, especially when you're upgrading to a new major version. Check the [upgrade considerations](#upgrade-considerations) for your upgrade version to anticipate any breaking changes or upgrade-specific instructions.
```
jwitz, line 30:
```suggestion
Astro upgrades can include breaking changes, especially when you're upgrading to a new major version. Check the [upgrade considerations](#upgrade-considerations) for your upgrade version to anticipate any breaking changes or upgrade-specific instructions before you proceed.
```

PR #2838, opened 2023-09-13
`astro dev` add generated-index and change description

manmeetkaur:
As per our discussion yesterday in Doc's Jam, created a `generated-index` page for `astro dev`. Also changed the description so that brief one-liner appears on the index page.


sidebarsAstro.js
```diff
@@ -408,6 +408,14 @@ module.exports = {
         {
           type: "category",
           label: "astro dev",
+          link: {
+            type: 'generated-index',
```
lzdanski, line 412: I really like the content and purpose of this page, and I didn't realize we had a generated-index option with docusaurus! We'll need to add a few other pages that are similar, but before moving forward on it, @jwitz do we have any styling options for these pages? I find the icon structure pretty difficult to skim because the definitions and descriptions are cut off, and that's what one of these indexes is partially for.

I don't know that this is ready to go out with the defaults because the standard icon makes it look like either images have failed to load and we're kind of forcing this landing page to do something it wasn't designed for (act as a reference index). This might be something we need to consult with Paola/Katie more about to get the design right before we move forward with it. 

on the other hand - i think that it would be great if we considered adding these to other parts of our reference docs after we can get some icons to describe the different doc types. Maybe that'll be a good part of doc types v2!

Screenshot 2023-09-13 at 10 29 17 AM

jwitz: Hm we could definitely modify the CSS to show more of the description before it cuts off, but the icons are trickier. We have two options:

- Feed an agreed-upon icon for CLI docs into the `sidebar_custom_props: { icon: '<image>' }` frontmatter on all docs pages. This would be easier to maintain and the best option if we want to keep these bubble boxes, but the icon will obvi be a point of debate.
- Instead of using the autogenerated TOC, we create dedicated overview pages for each menu and link those instead. Sidebar would look like:
```
type: 'category',
      label: 'astro dev',
      link: {type: 'doc', id: 'cli/astro-dev-overview'},
```

The latter option is much harder to maintain as we'd have to constantly update the page with links to new commands, so I'd opt for the former: lightly customizing the auto-TOC to fit our needs.

manmeetkaur: My take on this:  
- We currently have pages in our Docs, that already rely on that icon (see [Basics](https://docs.astronomer.io/learn/category/basics) & [Airflow Tutorials](https://docs.astronomer.io/learn/category/airflow-tutorials)). I had brought this point earlier that we need to have icons couple of months back and I think @lzdanski and I also discussed last week. 
- I think we discussed option #2 in DocsJam, and it was suggested to use generated-index, as an easy solution to this problem

tl;dr not sure if icons or a new doc is a blocker. but I think having this page adds value, and other things can be improved upon later.

jwitz: @manmeetkaur this is the same icon (as a PNG) that we use for CLI note types. Can you try to pass that into the files in this menu to see if that makes it look better? 

We can deal with the description lengths another time, or just remove them for now.

noun-1819808

sidebarsAstro.js
```diff
@@ -408,6 +408,14 @@ module.exports = {
...
+            title: 'Interact with local Airflow environment',
+            description: 'Commands to interact with your local Airflow environment using Astro CLI.'
```
jwitz, line 414:
```suggestion
            description: 'Use `astro dev` commands to manage your Astro project and interact with your local Airflow environment.'
```

jwitz (approved):
New icons look great :)

Screenshot 2023-09-20 at 10 57 39 AM

Just some minor copy and formatting suggestions, otherwise LGTM

PR #2841, opened 2023-09-13
update to remove custom tags step - there's a bug

cmarteepants:
https://github.com/astronomer/astronomer-providers-logging/issues/52


lzdanski (approved):
Changes in preview LGTM. will add note to dev issue about adding this step back to the docs after the bug is fixed.

PR #2864, opened 2023-09-20
Try a layout for Connection and Example docs

fritz-astronomer:
Had an idea for how to put together the integration docs. Currently the connection docs are semi-buried and hard to find.

image
image

### DIFF
--- a/learn/airflow-redshift.md
+++ b/learn/airflow-redshift.md
@@ -1,6 +1,6 @@
 ---
 title: "Orchestrate Redshift operations with Airflow"
-sidebar_label: "Amazon Redshift"
+sidebar_label: "Example"
 description: "Orchestrate Redshift queries from your Airflow DAGs."
 id: airflow-redshift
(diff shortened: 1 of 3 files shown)


learn/airflow-redshift.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Orchestrate Redshift operations with Airflow"
-sidebar_label: "Amazon Redshift"
+sidebar_label: "Example"
```
kentdanas, line 3: I like this organization, but I don't love "Example" here, partly because we have another "Examples" section and I think that's confusing. What about "Tutorial" or "Implementation tutorial" or something like that?

kentdanas (commented):
Thanks for starting this change @fritz-astronomer, definitely on my radar now that we're post-summit. I like this organization, but might want to think a bit on how we do the sidebar labels. 

Also, I think if we do this, for any Connection docs that don't have an integration tutorial, we should have those still under integrations, so everything gets moved at once.

kentdanas (approved):

PR #2868, opened 2023-09-22
Add warning for self-hosted runners

### LINKED ISSUE #2769: CI CD Warning

manmeetkaur:
Resolves #2769

### DIFF
--- a/astro/ci-cd-templates/github-actions.md
+++ b/astro/ci-cd-templates/github-actions.md
@@ -35,7 +35,7 @@ If you use a [self-hosted runner](https://docs.github.com/en/actions/hosting-you
 
 - Add `ASTRO_API_TOKEN` to your repository and include a check in your GitHub workflow to verify that it exists. 
 - Use Deployment API tokens, which are scoped only to one Deployment, instead of Workspace or Organization API tokens.
-- Specify `deployment-id` or `deplyment-name` in your action as shown in the following templates.
+- Specify `deployment-id` or `deployment-name` in your action. For example, `astro deploy <deployment-id> or `astro deploy -n <deployment-name>`.
 - Add the command `astro logout` at the end of your workflow to ensure that your authentication token is cleared from the `config.yaml` file.
 
 :::
--- a/astro/ci-cd-templates/github-actions.md
+++ b/astro/ci-cd-templates/github-actions.md
@@ -35,7 +35,7 @@ If you use a [self-hosted runner](https://docs.github.com/en/actions/hosting-you
 
 - Add `ASTRO_API_TOKEN` to your repository and include a check in your GitHub workflow to verify that it exists. 
 - Use Deployment API tokens, which are scoped only to one Deployment, instead of Workspace or Organization API tokens.
-- Specify `deployment-id` or `deployment-name` in your action. For example, `astro deploy <deployment-id> or `astro deploy -n <deployment-name>`.
+- Specify `deployment-id` or `deployment-name` in your action. For example, `astro deploy <deployment-id>` or `astro deploy -n <deployment-name>`.
 - Add the command `astro logout` at the end of your workflow to ensure that your authentication token is cleared from the `config.yaml` file.
 
 :::
--- a/astro/ci-cd-templates/github-actions.md
+++ b/astro/ci-cd-templates/github-actions.md
@@ -528,13 +528,6 @@ If your Astro project requires additional build-time arguments to build an image
             astro deploy <your-deployment-id> --image-name ${{ steps.image_tag.outputs.image_tag }}
     ```
 
-</TabItem>
-
-</Tabs>
-
-
-### Example
-
 For example, to create a CI/CD pipeline that deploys a project which [installs Python packages from a private GitHub repository](cli/develop-project.md#install-python-packages-from-private-sources), you would use the following configuration:
 
   ```yaml
@@ -584,3 +577,6 @@ If you need guidance configuring a CI/CD pipeline for a more complex use case in
 :::
 
 
+</TabItem>
+
+</Tabs>
--- a/astro/ci-cd-templates/github-actions.md
+++ b/astro/ci-cd-templates/github-actions.md
@@ -29,9 +29,13 @@ To learn more about CI/CD on Astro, see [Choose a CI/CD strategy](set-up-ci-cd.m
 
 Each CI/CD template implementation might have additional requirements.
 
-:::info
+:::caution
 
-If you use a self-hosted runner for GitHub Actions and you're using a Workspace or Organization API token that has access to more than one Deployment, ensure that you specify `deployment-id` in your action as shown in the following templates. This reduces the risk of accidentally deploying to the wrong Deployment.
+If you use a self-hosted runner to execute jobs from GitHub Actions, Astro CLI's `config.yaml` might be shared across your organization. In such cases, to reduce the risk of accidentally deploying to the wrong Deployment, ensure the following:
+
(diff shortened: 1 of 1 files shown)


astro/ci-cd-templates/github-actions.md
```diff
@@ -29,9 +29,13 @@ To learn more about CI/CD on Astro, see [Choose a CI/CD strategy](set-up-ci-cd.m
 
 Each CI/CD template implementation might have additional requirements.
 
-:::info
+:::caution
+
+If you use a self-hosted runner to execute jobs from GitHub Actions, Astro CLI's `config.yaml` might be shared across your organization. In such cases, to reduce the risk of accidentally deploying to the wrong Deployment, ensure the following:
```
jwitz, line 34:
```suggestion
If you use a [self-hosted runner](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners) to execute jobs from GitHub Actions, the Astro CLI's `config.yaml` file, which stores default deploy details, might be shared across multiple CI/CD pipelines. To reduce the risk of accidentally deploying to the wrong Deployment, ensure the following:
```

astro/ci-cd-templates/github-actions.md
```diff
@@ -523,6 +527,13 @@ If your Astro project requires additional build-time arguments to build an image
             astro deploy <your-deployment-id> --image-name ${{ steps.image_tag.outputs.image_tag }}
     ```
 
+</TabItem>
```
jwitz, line 530: I moved this topic back into the tabs on purpose. It's an example for using a custom image (tab 3), not a general purpose example. Can you move this back?

astro/ci-cd-templates/github-actions.md
```diff
@@ -29,9 +29,13 @@ To learn more about CI/CD on Astro, see [Choose a CI/CD strategy](set-up-ci-cd.m
...
 
-:::info
+:::caution
+
+If you use a self-hosted runner to execute jobs from GitHub Actions, Astro CLI's `config.yaml` might be shared across your organization. In such cases, to reduce the risk of accidentally deploying to the wrong Deployment, ensure the following:
 
-If you use a self-hosted runner for GitHub Actions and you're using a Workspace or Organization API token that has access to more than one Deployment, ensure that you specify `deployment-id` in your action as shown in the following templates. This reduces the risk of accidentally deploying to the wrong Deployment.
+- Add the Astronomer GitHub secrets to your repository and include a check in your GitHub workflow to verify this; otherwise, throw an error.
+- Specify `deployment-id` or `deplyment-name` in your [`astro deploy`](cli/astro-deploy.md) command.
+- Add an `astro logout` command at the end of your workflow to ensure the token is cleared from the `config.yaml` file.
```
jwitz, line 38:
```suggestion
- Add `ASTRO_API_TOKEN` to your repository and include a check in your GitHub workflow to verify that it exists. 
- Use Deployment API tokens, which are scoped only to one Deployment, instead of Workspace or Organization API tokens.
- Specify `deployment-id` or `deplyment-name` in your action as shown in the following templates.
- Add the command `astro logout` at the end of your workflow to ensure that your authentication token is cleared from the `config.yaml` file.
```
Does `config.yaml` store the token in plaintext, or does it just reference the GH secret?

jwitz: I also added a point to use Deployment API tokens, which at a minimum will throw an error of the CLI tries deploying to the wrong deployment.

jwitz (approved):
Once @erdos2n reviews this LGTM!

astro/ci-cd-templates/github-actions.md
manmeetkaur: plain-text

astro/ci-cd-templates/github-actions.md
manmeetkaur: If you notice the example is a use-case that is a mix of a single-branch deploy from a private network, meaning using the regular checkout github action `actions/checkout@v3` and also creating a custom image. So a user might straightaway want to see the example, and see the use github action. It is easier to look at an example atleast and understand than read lines. If you want I can add a one liner in the example to explain a bit more for clarity. WDYT? lmk :)

jwitz: For all of the other deploy types, the example _is_ the template. If we're going to include an example that applies to all "private network" templates, it should be a more general implementation. For the average user who doesn't have a custom image consideration, I think framing this as a general example creates more questions than answers.

So I'm sticking with my original point: It makes more sense to keep the custom image example in the custom image tab. The other tabs already have templates that users can read to understand the implementations that suit their specific use case.

manmeetkaur:
Hey @erdos2n , could you please take a look at this ?

erdos2n:
Thanks for doing this @manmeetkaur! One thing, can we add a line showing the command  specifying the deployment name. This has been used as a safe guard to deploy to a different deployment. 

```diff
+ astro deploy -n <deployment-name>
```

astro/ci-cd-templates/github-actions.md
```diff
@@ -29,9 +29,14 @@ To learn more about CI/CD on Astro, see [Choose a CI/CD strategy](set-up-ci-cd.m
...
 
-:::info
+:::caution
+
+If you use a [self-hosted runner](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners) to execute jobs from GitHub Actions, the Astro CLI's `config.yaml` file, which stores default deploy details, might be shared across your organization and hence multiple CI/CD pipelines. To reduce the risk of accidentally deploying to the wrong Deployment, ensure the following:
 
-If you use a self-hosted runner for GitHub Actions and you're using a Workspace or Organization API token that has access to more than one Deployment, ensure that you specify `deployment-id` in your action as shown in the following templates. This reduces the risk of accidentally deploying to the wrong Deployment.
+- Add `ASTRO_API_TOKEN` to your repository and include a check in your GitHub workflow to verify that it exists. 
+- Use Deployment API tokens, which are scoped only to one Deployment, instead of Workspace or Organization API tokens.
+- Specify `deployment-id` or `deplyment-name` in your action as shown in the following templates.
```
erdos2n, line 38:
```suggestion
- Specify `deployment-id` or `deplyment-name` in your action as shown in the following templates.
+ Specify `deployment-id` or `deplyment-name` in your action as shown in the following templates. `astro deploy -n <deployment-name>`
```

erdos2n: Hello @manmeetkaur , can we make these changes first?

erdos2n (changes requested):
I left a comment in the updated instructions

PR #2874, opened 2023-09-25
Remove errant underscore

grahamcarman:
The variable was incorrectly named (with an extra trailing underscore), causing authentication failure. See [this doc for the correct variable name, without underscore](https://docs.astronomer.io/astro/ci-cd-templates/template-overview#image-deploy-templates).


jwitz (approved):
Thanks so much for this fix @grahamcarman ! LGTM

PR #2892, opened 2023-09-28
Make it clear example pytest is an example

sunkickr:
Some customers were confused why they needed to run a pytest that does not fit their context. The truth is that they do not need to run this test it is just example. Changing the tests name and updating this doc should make that clear


erdos2n:
LGTM @sunkickr

lzdanski:
@TJaniF + @kentdanas - I took a look at the PR and realized I don't have the context about why the test script was changed. Can you PTAL?

kentdanas (approved):

sunkickr:
Yes just the name is changing to make it clear it is an example

PR #2893, opened 2023-09-28
Update Runtime 8 EOL

jedcunningham:


jwitz (approved):
LGTM, is there a record on Slack where we decided to make this exception?

jedcunningham:
@jwitz it's the "3 months after the next stable release" clause that caused it to be extended. Sorry, should have put that in the description.

PR #2912, opened 2023-10-04
Update airflow-sql.md

joescottswanson:
* Fixes a typo where a ` was in the wrong spot
* Refers to the SQLExecuteQueryOperator instead of the deprecated SnowflakeOperator and PostgresOperator
* Includes some helpful notes about parameterized queries and templates

### DIFF
--- a/learn/airflow-sql.md
+++ b/learn/airflow-sql.md
@@ -82,10 +82,9 @@ In Airflow 2+, provider packages are separate from the core of Airflow. If you'r
 
 In Airflow, action operators execute a function. You can use action operators (or hooks if no operator is available) to execute a SQL query against a database. Commonly used SQL-related action operators include:
 
-- [PostgresOperator](https://registry.astronomer.io/providers/postgres/modules/postgresoperator)
+- [SQLExecuteQueryOperator](https://registry.astronomer.io/providers/apache-airflow-providers-common-sql/modules/SQLExecuteQueryOperator)
 - [MssqlHook](https://registry.astronomer.io/providers/mssql/modules/mssqlhook)
 - [MysqlOperator](https://registry.astronomer.io/providers/mysql/modules/mysqloperator)
-- [SnowflakeOperator](https://registry.astronomer.io/providers/snowflake/modules/snowflakeoperator)
 - [BigQueryOperator](https://registry.astronomer.io/providers/google/modules/bigqueryexecutequeryoperator)
 
 ### Transfer operators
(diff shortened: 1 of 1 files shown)


learn/airflow-sql.md
```diff
@@ -82,10 +82,9 @@ In Airflow 2+, provider packages are separate from the core of Airflow. If you'r
 
 In Airflow, action operators execute a function. You can use action operators (or hooks if no operator is available) to execute a SQL query against a database. Commonly used SQL-related action operators include:
 
-- [PostgresOperator](https://registry.astronomer.io/providers/postgres/modules/postgresoperator)
```
joescottswanson, line 85: I was checking out the source code of the `PostgresOperator` and the `SnowflakeOperator` and I noticed that both were deprecated in favor of the `SQLExecuteQueryOperator`. This doesn't seem to be the case for the MySQLOperator. I am not sure about the `BigQueryOperator`; I'm not familiar with BigQuery.

learn/airflow-sql.md
```diff
@@ -162,22 +161,24 @@ WHERE state = {{ conf['state_variable'] }}
 If you need a parameter that is not available as a built-in variable or a macro, such as a value from another task in your DAG, you can also pass that parameter into your query using the operator:
 
 ```python
-opr_param_query = SnowflakeOperator(
+opr_param_query = SQLExecuteQueryOperator(
     task_id="param_query",
     snowflake_conn_id="snowflake",
     sql="param-query.sql",
-	params={"date":mydatevariable}
+    params={"date":mydatevariable}
```
TJaniF, line 168:
```suggestion
    parameters={"my_date": mydatevariable}
```

Switching the example to use `parameters` instead of `params` to not have the DAG level params interfere and changing the key to `my_date` to be more explicit.

learn/airflow-sql.md
```diff
@@ -162,22 +161,24 @@ WHERE state = {{ conf['state_variable'] }}
...
 )
 ```
 
-And then reference that param in your SQL file:
+And then reference that param in your SQL file (note that above you're passing a dictionary to `parameters`, but in your query below you reference `params` to access the values of `parameters`):
```
TJaniF, line 172:
```suggestion
And then reference that param in your SQL file:
```

Changes when using `parameters` instead of `params` in the operator :)

learn/airflow-sql.md
```diff
@@ -162,22 +161,24 @@ WHERE state = {{ conf['state_variable'] }}
...
 
 ```sql
 SELECT *
 FROM STATE_DATA
 WHERE date = {{ params.date }}
```
TJaniF, line 177:
```suggestion
SELECT *
FROM STATE_DATA
WHERE date = %(my_date)s
```

Using SQL parameter rendering instead of a jinja template.

learn/airflow-sql.md
```diff
@@ -162,22 +161,24 @@ WHERE state = {{ conf['state_variable'] }}
...
 ```
 
+N.b. If you run your DAG with user-defined parameters (by running your DAG via the 'Trigger DAG w/config' option), these user-defined parameters will overwrite any value you pass into the `parameters` argument for the task run.
```
TJaniF, line 180:
~~~suggestion
An alternative option is to use [Jinja](templating.md) templates in your SQL statements and pass the relevant values using the `params` parameter.

```python
    opr_param_query = SQLExecuteQueryOperator(
        task_id="param_query",
        conn_id="snowflake_default",
        sql="param-query.sql",
        params={"my_date": "2022-09-01"},
    )
```

The value can be referenced using Jinja syntax in the SQL statement:

```sql
SELECT *
FROM STATE_DATA
WHERE date = '{{ params.date }}'
```

:::tip

Airflow params can also be defined at the DAG-level and passed to a DAG at runtime. Params passed at runtime will override param defaults defined at the DAG or the task level. For more information on params see the [Create and use params in Airflow](airflow-params.md) guide.

:::
~~~

Good idea to explain the relationship to DAG params here! I moved the `param` example down here to have everything about using that parameter together and contrast it with `parameters` above.

TJaniF (changes requested):
Thank you for the PR, this section was definitely in need of more explanation! I added some suggestions and small corrections :)

kentdanas (approved):

kentdanas:
Hey @joescottswanson thanks again for the PR! We made a couple small updates, but have now merged your changes.

learn/airflow-sql.md
joescottswanson: Thank you so much for this review/help with the docs! I wasn't super clear on how all of this worked (which was where this PR came from), but this really helped sort everything out for me.

joescottswanson:
Appreciate the review and changes! That answered a couple of questions for me.

PR #2936, opened 2023-10-11
Two snippets of info for Cosmos content

TJaniF:
BLOCKED behind https://github.com/astronomer/astronomer-cosmos/pull/592
to get the wording and fix right 

This is based on this Slack thread: https://astronomer.slack.com/archives/C015V2JFKT5/p1696873122011889 

Putting a note in how to resolve the `<> is not a valid ResourceType` error and setting retries to 2 for all our Cosmos content.


TJaniF:
Since the related PR to the Cosmos docs was merged with the same wording I'll merge this now :)

PR #2974, opened 2023-10-20
Related content notes

TJaniF:
Trying to interlink Academy modules, learn content and webinars more. :) 
I am not sure whether I picked the best spot for the note here, I think it could also go at the end of the intro, as long as it is in the same spot for all content.
Originally I wanted to make a custom admonition "Related Content" in purple but JS/CSS won. If you think it is something worth looking into I'll poke Katie for help 😅 .

Very open to discussing the best location and format for this :)

Aside from that I tried to get all relevant Academy modules and for webinars the newest relevant webinar or in a few cases Live (for example for DuckDB there is a live of webinar length 😄 ). I did add a couple of blog post when I knew there was a relevant technical blog post around, but very possible I missed some there.

### DIFF
--- a/learn/airflow-branch-operator.md
+++ b/learn/airflow-branch-operator.md
@@ -16,7 +16,7 @@ When designing your data pipelines, you may encounter use cases that require mor
 
 In this guide, you'll learn how you can use `@task.branch` (BranchPythonOperator) and `@task.short_circuit` (ShortCircuitOperator), other available branching operators, and additional resources to implement conditional logic in your Airflow DAGs.
 
-:::tip Related Content
+:::tip Related content
 
 There are multiple resources for learning about this topic. See also:
 
--- a/learn/airflow-branch-operator.md
+++ b/learn/airflow-branch-operator.md
@@ -16,7 +16,7 @@ When designing your data pipelines, you may encounter use cases that require mor
 
 In this guide, you'll learn how you can use `@task.branch` (BranchPythonOperator) and `@task.short_circuit` (ShortCircuitOperator), other available branching operators, and additional resources to implement conditional logic in your Airflow DAGs.
 
-:::tip Other ways to learn:
+:::tip Other ways to learn
 
 There are multiple resources for learning about this topic. See also:
 
--- a/learn/airflow-branch-operator.md
+++ b/learn/airflow-branch-operator.md
@@ -16,9 +16,9 @@ When designing your data pipelines, you may encounter use cases that require mor
 
 In this guide, you'll learn how you can use `@task.branch` (BranchPythonOperator) and `@task.short_circuit` (ShortCircuitOperator), other available branching operators, and additional resources to implement conditional logic in your Airflow DAGs.
 
-:::tip Related content
+:::tip Other ways to learn:
 
-Other ways to learn:
+There are multiple resources for learning about this topic. See also:
 
 - Astronomer Academy: [Airflow: Branching](https://academy.astronomer.io/astro-runtime-branching) module.
 
--- a/learn/airflow-branch-operator.md
+++ b/learn/airflow-branch-operator.md
@@ -18,6 +18,8 @@ In this guide, you'll learn how you can use `@task.branch` (BranchPythonOperator
 
 :::tip Related Content
 
+There are multiple resources for learning about this topic. See also:
+
 - Astronomer Academy: [Airflow: Branching](https://academy.astronomer.io/astro-runtime-branching) module.
 
 :::
--- a/learn/airflow-branch-operator.md
+++ b/learn/airflow-branch-operator.md
@@ -24,6 +24,12 @@ To get the most out of this guide, you should have an understanding of:
 - Dependencies in Airflow. See [Managing Dependencies in Apache Airflow](managing-dependencies.md).
 - Using Airflow decorators. See [Introduction to Airflow decorators](airflow-decorators.md).
 
+:::tip Related Content
+
+- Astronomer Academy: [Airflow: Branching](https://academy.astronomer.io/astro-runtime-branching) module.
+
+:::
+
 ## `@task.branch` (BranchPythonOperator)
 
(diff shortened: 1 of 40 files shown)


learn/airflow-databricks.md
```diff
@@ -74,6 +74,14 @@ To get the most out of this tutorial, make sure you have an understanding of:
 - Airflow operators. See [Operators 101](what-is-an-operator.md).
 - Airflow connections. See [Managing your Connections in Apache Airflow](connections.md).
 
+:::tip Related Content
+
+- Webinar: [How to Orchestrate Databricks Jobs Using Airflow](https://www.astronomer.io/events/webinars/how-to-orchestrate-databricks-jobs-using-airflow/).
+- Blog post: [Comparing Data Orchestration: Databricks Workflows vs. Apache Airflow, Part 1](https://www.astronomer.io/blog/comparing-data-orchestration-databricks-workflows-vs-apache-airflow-part-1/).
```
kentdanas, line 80: My initial reaction here is we should not include blog posts in these - blog posts can (and should) link to Learn/webinars/academy, but from the more instructional content we should only link to other instructional content, if that makes sense. Especially since blog posts don't really get updated over time. Happy to hear other thoughts on this though, I could potentially be convinced either way.

learn/airflow-branch-operator.md
```diff
@@ -24,6 +24,12 @@ To get the most out of this guide, you should have an understanding of:
 - Dependencies in Airflow. See [Managing Dependencies in Apache Airflow](managing-dependencies.md).
 - Using Airflow decorators. See [Introduction to Airflow decorators](airflow-decorators.md).
 
+:::tip Related Content
```
kentdanas, line 27: Location-wise, I would prefer at the end of the intro. Within `Assumed knowledge` makes it seem to me like we're saying they should review the other content before reading the guide.

kentdanas (changes requested):
Thanks for doing this @TJaniF! I left a few comments on placement and inclusion of blog posts. Language-wise, I'm fine with how you have it, but will let docs team make suggestions if they have any :)
I didn't review which content you're suggesting on all of them super closely, but trust you picked the right ones

learn/airflow-databricks.md
TJaniF: I was very on the fence whether to even include them in this first draft, but then thought easier to delete than add haha. Yes, I agree :)

TJaniF (commented):
Thank you @kentdanas :) 
I moved all the notes, and yes looks much better.

jwitz (changes requested):
The more linking the better! The way we handle this kind of content on the Astro side is to have a section at the bottom of each doc called "[See also](https://docs.astronomer.io/astro/deploy-code#see-also)" with links to other pages. I think that would probably also be best here - it's less disruptive and positioned where users are most likely to need it: at the end of the article after they realize the thing they're looking for isn't there.

kentdanas:
@jwitz I get what you're saying re: See Also at the end, but also I do think there's a distinction (that maybe we're not currently making) between "related" content, and essentially the same content in different forms. e.g. if I'm a user wanting to learn about Datasets, and from google I land on our Learn doc, but really I would prefer to learn by watching a video, I'd rather have a note up front that says, hey, if you'd prefer to learn in this format you also have this academy resource available to you. Which is a bit different than, hey, here's some additional content you might be interested in. Let me know what you think here. We can also discuss more in our collab meeting Thurs.

jwitz:
@kentdanas That makes a lot of sense. Because the note has a more specific purpose, I agree that it makes sense to keep at the top of the doc. But we might want to make that point more clear in the note itself, with a generic sentence leading up to the bulleted list like: "There are multiple resources for learning about this topic. See also:"

Just so it's clear that these are parallel resources, rather than successive resources that you'd use only after finishing a given piece of content.

TJaniF:
@jwitz I like that idea! Added that sentence to all notes. :) 

I think in general how we link this content will probably evolve, this is just a first step bringing webinar, edu and learn content closer together ❤️

learn/airflow-branch-operator.md
```diff
@@ -16,6 +16,14 @@ When designing your data pipelines, you may encounter use cases that require mor
 
 In this guide, you'll learn how you can use `@task.branch` (BranchPythonOperator) and `@task.short_circuit` (ShortCircuitOperator), other available branching operators, and additional resources to implement conditional logic in your Airflow DAGs.
 
+:::tip Related Content
```
jwitz, line 19:
```suggestion
:::tip Related content
```
Tiny nit - we should use sentence case in note title just like for normal topics

jwitz (approved):
One tiny suggestion, otherwise LGTM!

marclamberti:
@TJaniF love it! I agree with @kentdanas as it's more "related content" like a different format than a "see also" or "what's next" type of content. I'm unsure about "There are multiple resources for learning about this topic. See also:". I'm not a pro at wording, but I would love to express that modules have quizzes and videos. To make the distinction more explicit. Something like "If you enjoy learning through videos and want to test your knowledge with quizzes, this could be a great option for you."

Another idea (not related). Why not add a link to a quiz (from a module) at the end of the "learn" pages? Not only do you read the content but also you can test your knowledge.

TJaniF:
@marclamberti thank you! Yeah I think we can discuss the best wording tomorrow, also depends on if we think all of those links will be video based content or if we also include quickstart repositories and use cases as some point. 🤔 

And I love the idea to link to a quiz at the end, active recall is the (second*) best way of learning! And a getting a question right gives some nice dopamine. 😄 
I think we'd have to make sure the quiz question can be answered with the information in the learn guide. Other than that I think we just have to find the best way to link/present it. 🤔 I'd put that in a second PR so we can iterate on the idea separately :)

*the best of course is by teaching the material to someone else 😄

dpatlut (approved):
Overall, LGTM -- EDU will have to document that we need to let y'all know if names or links change, but we will cross that bridge when we get there and figure out the best process.

We chatted with Kenten and the team while you were out about changing the header of the box so I will approve this but Kenten will likely add another request change for that.

kentdanas (approved):

PR #2977, opened 2023-10-20
Updated airflow_dbt_simple.py

zvijayakumar:
In dbt cloud, we get empty data if the job hasn't run at least once and added the edge case to ignore the empty data set. So it returns true. 

To recreate the error:
1. Create a job in dbt cloud - (Do not run it from dbt cloud)
2. Use the airflow_dbt_simple to trigger from Airflow
3. It will throw an error, Index not found.


kentdanas (approved):
@zvijayakumar thanks for catching this issue and submitting a fix, we appreciate it!

Issue #2956, opened 2023-10-17
Find a new way to present "Software" k8s compatibility


danielhoherd:
### Issue Type

Inaccurate, misleading, or out-of-date information

### Links to Affected Docs

https://github.com/astronomer/docs/pull/2772#pullrequestreview-1605594418

### The Issue or Context

We are running into trouble communicating which patch versions of our "Software" are compatible with which versions of kubernetes. It is working OK right now, but I think it could be better, and since we are potentially going to double the duration of our LTS support, the compatibility matrix will get much more complicated since the number of supported k8s versions will double, and half of those will be deprecated during the lifecycle of the "Software" release.

### Fix or Proposed Change

I think a gantt chart would be good, or maybe just a simple table:

|        | 1.23 | 1.24 | 1.25 | 1.26 | 1.27 |
| :----: | :--: | :--: | :--: | :--: | :--: |
| 0.30.0 |  ✅  |  ✅  |  ✅  |      |      |
| 0.30.1 |  ✅  |  ✅  |  ✅  |  ✅  |      |
| 0.30.2 |  ✅  |  ✅  |  ✅  |  ✅  |  ✅  |
| 0.31.0 |  ✅  |  ✅  |  ✅  |      |      |
| 0.31.2 |      |  ✅  |  ✅  |  ✅  |      |
| 0.32.0 |  ✅  |  ✅  |  ✅  |  ✅  |  ✅  |
| 0.32.1 |      |  ✅  |  ✅  |  ✅  |  ✅  |
| 0.33.1 |      |      |  ✅  |  ✅  |  ✅  |

### Additional Notes

Thinking about the presentation problem in general terms, we are essentially trying to present three dimensions of information on a two dimensional canvas. These informational dimensions are `(software_major_minor, software_patch, k8s_version)`. I think that the answer is we will just need to collapse the first two dimensions there and enumerate every possible pair. This is the solution that is displayed on the example table above. It's not as awesome as I'd like, but I think it is a good compromise. Obviously the style could be changed to be easier to visually grok.

PR #2999, opened 2023-10-24
Create new Kubernetes matrix

jwitz:
Resolves https://github.com/astronomer/docs/issues/2956

I edited @danielhoherd 's example chart based on what I know about current version compatibility, and also condensed some of the rows so that each row covered multiple releases.


software/version-compatibility-reference.md
```diff
@@ -43,10 +34,20 @@ Due to the [deprecation of Dockershim](https://kubernetes.io/blog/2020/12/02/doc
...
 
+See the following table for all supported Kubernetes versions in each maintained version of Astronomer Software.
+
+| Astronomer platform | Kubernetes 1.22 | Kubernetes 1.23 | Kubernetes 1.24 | Kubernetes 1.25 | Kubernetes 1.26 | Kubernetes 1.27 |
+| :-----------------: | :-------------: | :-------------: | :-------------: | :-------------: | :-------------: | :-------------: |
+|   0.30.0- 0.30.7    |        ✔️        |        ✔️        |        ✔️        |        ✔️        |        ✔️        |                 |
+|       0.30.8        |                 |        ✔️        |        ✔️        |        ✔️        |        ✔️        |        ✔️        |
+|   0.32.0- 0.32.2    |        ✔️        |        ✔️        |        ✔️        |        ✔️        |        ✔️        |        ✔️        |
+|    0.32.3-0.32.4    |                 |        ✔️        |        ✔️        |        ✔️        |        ✔️        |        ✔️        |
+|   0.33.0 - 0.33.1   |                 |                 |        ✔️        |        ✔️        |        ✔️        |        ✔️        |
```
danielhoherd, line 49: If we're going to use this dash syntax ("0.33.0 - 0.33.1") then we should be consistent with the spacing. Honestly though I think it's better to list each version because it is searchable. With the dash method, searching for 0.33.5 would return zero results. I understand that searchability is only aspect of the presentation here though so maybe that's not the right course of action.

danielhoherd: Also 0.33.1 and 0.32.4 support 1.28. In all future "Software" versions, the supported k8s list can be checked by inspecting https://github.com/astronomer/astronomer/blob/v${VERSION}/metadata.yaml eg: https://github.com/astronomer/astronomer/blob/v0.33.1/metadata.yaml

Alternatively, all old "Software" versions, including old versions, can be checked by inspecting .circleci/config.yml:
```
$ grep '^  platform-' .circleci/config.yml
  platform-1-23-17:
  platform-1-24-15:
  platform-1-25-11:
  platform-1-26-6:
  platform-1-27-3:
  platform-1-28-0:
```
or
```
>>> [x for x in yaml.safe_load(Path('.circleci/config.yml').read_text())['jobs'].keys() if x.startswith('platform-')]
['platform-1-23-17', 'platform-1-24-15', 'platform-1-25-11', 'platform-1-26-6', 'platform-1-27-3', 'platform-1-28-0']
```

jwitz: I'm not super worried about the search, only because I our search might still catch this table if a user searched "Kubernetes" and any given release regardless if it appears in the table.

danielhoherd (approved):
Seems like an improvement to me.

This is how it shows up now:
Screenshot 2023-10-25 at 9 34 49 AM

jwitz:
Me too! Let's merge it.

PR #3016, opened 2023-10-25
add azure standard

chomipi88:


lzdanski:
@chomipi88 - Is there a docs issue for any other docs changes that we'll need to make for standard cluster support on Azure? I think there might be other places we'll need to add mentions and other supporting information.

lzdanski (approved):
Confirmed in preview

chomipi88:
@lzdanski i couldn't find any other references where we get into cloud specifics for Standard cluster support beyond this doc. This was a soft launch of Azure Standard clusters.
We're going to make a lot more noise with the launch of the Azure Native Service.

lzdanski:
@chomipi88 - Awesome, thank you for being thorough!

PR #3027, opened 2023-10-30
Fix Azure instructions for changed navigation

ryanahamilton:
The navigation behavior changed several months ago and the current instructions are no longer valid. I've updated to valid instructions.


jwitz (approved):
Thank you @ryanahamilton !

PR #3042, opened 2023-10-31
Note that the security tab is not in Astro in airflow-ui.md

RNHTTR:
Because RBAC is handled via Astro, this tab was removed from Airflow on Astro.

### DIFF
--- a/learn/airflow-ui.md
+++ b/learn/airflow-ui.md
@@ -158,6 +158,12 @@ Click a dataset to open the history of all updates to the dataset that were reco
 
 ## Security tab
 
+:::info Astro does not support the Security tab
+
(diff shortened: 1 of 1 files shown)


learn/airflow-ui.md
```diff
@@ -158,6 +158,12 @@ Click a dataset to open the history of all updates to the dataset that were reco
 
 ## Security tab
 
+:::info Astro does not support the Security tab
+
+Because role-based acceess control is managed at the Astro platform level, the Security tab is not useful for Airflow when running on Astro. As a result, this tab does not exist in Airflow on Astro.
```
kentdanas, line 163:
```suggestion
On Astro, role-based access control is managed at the platform level. As a result, the Security tab is not needed and is not available on Airflow deployments on Astro.
```

kentdanas (approved):
Thanks @RNHTTR! Small wording suggestion, otherwise looks great and I'll go ahead and merge.

PR #3049, opened 2023-11-02
Update Software table to include legacy versions

jwitz:
See https://astronomer.slack.com/archives/C04JS0GADTL/p1698809649137609

Also used this opportunity to update our upgrade paths

### DIFF
--- a/software/version-compatibility-reference.md
+++ b/software/version-compatibility-reference.md
@@ -17,9 +17,9 @@ The following table shows version compatibility information for all currently su
 
 | Astronomer Platform | Postgres | Python                                         | Astro Runtime        | Helm |
 | ------------------- | -------- | ---------------------------------------------- | -------------------- | ---- |
-| v0.30               | 11+      | 3.6 - 3.11 (_3.9-3.11 Require Airflow 2.2.0+_) | All Runtime versions | 3.6  |
-| v0.32               | 11+      | 3.6 - 3.11 (_3.9-3.11 Require Airflow 2.2.0+_) | All Runtime versions | 3.6  |
-| v0.33               | 11+      | 3.6 - 3.11 (_3.9-3.11 Require Airflow 2.2.0+_) | All Runtime versions | 3.6  |
+| v0.30               | 11+      | 3.6 - 3.11 (_3.9-3.11 Require Airflow 2.2.0+_) | All Runtime versions | 3.13  |
+| v0.32               | 11+      | 3.6 - 3.11 (_3.9-3.11 Require Airflow 2.2.0+_) | All Runtime versions | 3.13 |
+| v0.33               | 11+      | 3.6 - 3.11 (_3.9-3.11 Require Airflow 2.2.0+_) | All Runtime versions | 3.13  |
 
 See [Kubernetes version support table and policy](#kubernetes-version-support-table-and-policy) for Astronomer platform compatibility with Kubernetes.
 
--- a/software/version-compatibility-reference.md
+++ b/software/version-compatibility-reference.md
@@ -68,7 +68,7 @@ The following table shows version compatibility information for all versions of 
 | v0.26               | 1.17, 1.18, 1.19, 1.20, 1.21              | 9.6+                                      | 3.6, 3.7, 3.8, 3.9 (_requires AC 2.2.0+_)      | All Astronomer Certified versions | 3.6  |
 | v0.27               | 1.18, 1.19, 1.20, 1.21                    | 9.6+                                      | 3.6, 3.7, 3.8, 3.9 (_requires AC 2.2.0+_)      | All Astronomer Certified versions | 3.6  |
 | v0.28               | 1.19¹, 1.20¹, 1.21, 1.22, 1.23, 1.24      | 9.6+                                      | 3.6, 3.7, 3.8, 3.9 (_requires AC 2.2.0+_)      | All Astronomer Certified versions | 3.6  |
-| v0.29               | 1.19¹, 1.20¹, 1.21, 1.22, 1.23, 1.24 9.6+ | 3.6, 3.7, 3.8, 3.9 (_requires AC 2.2.0+_) | All supported Certified and Runtime versions   | 3.6                               |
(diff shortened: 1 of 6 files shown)


software/version-compatibility-reference.md
```diff
@@ -13,11 +13,13 @@ While the tables below reference the minimum compatible versions, we typically r
...
+The following table shows version compatibility information for all currently supported versions of Astronomer Software:
+
 | Astronomer Platform | Postgres | Python                                         | Astro Runtime        | Helm |
 | ------------------- | -------- | ---------------------------------------------- | -------------------- | ---- |
-| v0.30               | 11+      | 3.6, 3.7, 3.8, 3.9 (_requires Airflow 2.2.0+_) | All Runtime versions | 3.6  |
-| v0.32               | 11+      | 3.6, 3.7, 3.8, 3.9 (_requires Airflow 2.2.0+_) | All Runtime versions | 3.6  |
-| v0.33               | 11+      | 3.6, 3.7, 3.8, 3.9 (_requires Airflow 2.2.0+_) | All Runtime versions | 3.6  |
+| v0.30               | 11+      | 3.6 - 3.11 (_3.9-3.11 Require Airflow 2.2.0+_) | All Runtime versions | 3.6  |
+| v0.32               | 11+      | 3.6 - 3.11 (_3.9-3.11 Require Airflow 2.2.0+_) | All Runtime versions | 3.6  |
+| v0.33               | 11+      | 3.6 - 3.11 (_3.9-3.11 Require Airflow 2.2.0+_) | All Runtime versions | 3.6  |
```
danielhoherd, line 22: FWIW helm 3.6 is EOL, and we no longer use it. [Here is the helm compatibility chart](https://helm.sh/docs/topics/version_skew/#supported-version-skew), which shows 3.6 being compatible with up to k8s 1.21, which has EOL in all cloud providers for over a year.

We are using helm 3.13 now. https://github.com/astronomer/ci-images/blob/main/helm-release/Dockerfile#L5

Since the helm version directly correlates to the k8s version, we could auto-generate compatible helm versions by inspecting the compatible k8s versions.

software/version-compatibility-reference.md
```diff
@@ -56,3 +58,24 @@ For more information on upgrading Kubernetes versions, follow the guidelines off
...
+| ------------------- | ----------------------------------------- | ----------------------------------------- | ---------------------------------------------- | --------------------------------- |
+| v0.26               | 1.17, 1.18, 1.19, 1.20, 1.21              | 9.6+                                      | 3.6, 3.7, 3.8, 3.9 (_requires AC 2.2.0+_)      | All Astronomer Certified versions | 3.6  |
+| v0.27               | 1.18, 1.19, 1.20, 1.21                    | 9.6+                                      | 3.6, 3.7, 3.8, 3.9 (_requires AC 2.2.0+_)      | All Astronomer Certified versions | 3.6  |
+| v0.28               | 1.19¹, 1.20¹, 1.21, 1.22, 1.23, 1.24      | 9.6+                                      | 3.6, 3.7, 3.8, 3.9 (_requires AC 2.2.0+_)      | All Astronomer Certified versions | 3.6  |
+| v0.29               | 1.19¹, 1.20¹, 1.21, 1.22, 1.23, 1.24 9.6+ | 3.6, 3.7, 3.8, 3.9 (_requires AC 2.2.0+_) | All supported Certified and Runtime versions   | 3.6                               |
```
danielhoherd, line 71:
```suggestion
| v0.29               | 1.19¹, 1.20¹, 1.21, 1.22, 1.23, 1.24 | 9.6+ | 3.6, 3.7, 3.8, 3.9 (_requires AC 2.2.0+_) | All supported Certified and Runtime versions   | 3.6                               |
```

software/version-compatibility-reference.md
```diff
@@ -56,3 +58,24 @@ For more information on upgrading Kubernetes versions, follow the guidelines off
...
+| v0.31               | 1.21, 1.22, 1.23, 1.24 , 1.25¹, 1.26¹     | 11.19+                                    | 3.6, 3.7, 3.8, 3.9 (_requires Airflow 2.2.0+_) | All Runtime versions              | 3.6  |
+
+:::info
+
+¹ Support for some Kubernetes versions is limited to specific Astronomer Software patch versions.
+
+- Support for Kubernetes 1.19 and 1.20 ends with Astronomer Software versions 0.28.7 and 0.29.5.
+- Support for Kubernetes 1.25 and 1.26 starts in Astronomer Software 0.31.2.
```
danielhoherd, line 79: Why do we list these two specific bits of information outside of the table? This seems either unneeded, or incomplete.

jwitz: This was the format from before. I figured that if we're going to keep adding Kubernetes versions, we need to condense how we present them. Can't have a run on table, which would happen if we use the Gantt style chart above.

danielhoherd: Should we add information about when kubernetes 1.21-1.23 ended, and when 1.27 and 1.28 began?

software/version-compatibility-reference.md
jwitz: Do you think Helm is even a compatibility consideration for folks? It's always been a part of this table, but this is the first time someone's noticed how out of date it is. 

Definitely open to have this sort of thing generated in a private repo, and the docs team could then update our public docs whenever it changes!

software/version-compatibility-reference.md
jwitz: From what I can tell, neither of those support timelines started/ ended in these legacy versions, so we don't need to list their start/ end date

software/version-compatibility-reference.md
danielhoherd: @jwitz honestly we can probably remove it. the helm version is tied to the k8s version, and that should be sufficient. there was a behavior in 3.6 that was corrected, which IIRC is why we started explicitly saying we needed 3.6. there have been no such significant fixes or improvements since then, so i think we can drop the helm requirement column. if we want to replace it with something we can just link to the helm compatibility page.

PR #3052, opened 2023-11-02
Add Astro Runtime Provider Packages versions to our Docs

odaneau-astro:
The goal of this PR is to provide the list of installed provider packages for each Astro Runtime version we support.

I automated the retrieval with a Python script:
```python
import json
import logging
import subprocess

import pandas as pd
import requests


def get_runtime_versions():
    # URL to fetch JSON data
    url = "https://updates.astronomer.io/astronomer-runtime"
    response = requests.get(url)
    if response.status_code == 200:
        data = response.json()
    else:
        print("Failed to retrieve data from the URL.")
    versions = list(data.get("runtimeVersions").keys())
    versions.sort(reverse=True)
    logging.info(versions)
    return versions


def cleanup(text: str):
    text = text.replace("\b", "")
    text = text.replace("\f", "")
    text = text.replace("\n", "")
    text = text.replace("\r", "")
    text = text.replace("\t", "")
    return text


def get_version_providers(runtime_version: str):
    # Use subprocess to run the docker command and capture its output
    cmd = f"docker run --rm -e AIRFLOW__LOGGING__LOGGING_LEVEL=CRITICAL quay.io/astronomer/astro-runtime:{runtime_version} airflow providers list --output json"
    # Parse the JSON output from the docker command
    output = subprocess.check_output(cmd, shell=True).decode("utf-8")
    output = cleanup(output)
    packages = json.loads(output)

    df = pd.DataFrame.from_records(packages, exclude=["description"])
    print(df.to_markdown(index=False))


if __name__ == "__main__":
    versions = get_runtime_versions()
    print(versions)
    for version in versions:
        print("")
        print("")
        print(f"## Runtime version {version}")
        try:
            get_version_providers(version)
        except Exception:
            print(f"! Could not get packages for {version}")
            continue

```

### DIFF
--- a/astro/runtime-image-architecture.md
+++ b/astro/runtime-image-architecture.md
@@ -93,15 +93,7 @@ The latest version of the Astro Runtime image has the following open source prov
 - Redis [`apache-airflow-providers-redis`](https://pypi.org/project/apache-airflow-providers-redis/)*
 - SQLite [`apache-airflow-providers-sqlite`](https://pypi.org/project/apache-airflow-providers-sqlite/)
 
-### Provider package versioning
-
-If an Astro Runtime release includes changes to an installed version of a provider package that is maintained by Astronomer (`astronomer-providers` or `openlineage-airflow`), the version change is documented in the [Astro Runtime release notes](runtime-release-notes.md).
-
-To determine the version of any provider package installed in your current Astro Runtime image, run:
-
-```
-docker run --rm <runtime-image> pip freeze | grep <provider>
-```
+See [Astro Runtime provider package reference](runtime-providers-reference.md) to see which versions of these provider packages are installed for each version of Astro Runtime. 
 
 ## Python versioning
 
--- a/astro/runtime-providers-reference.md
+++ b/astro/runtime-providers-reference.md
@@ -1,22 +1,28 @@
 ---
-sidebar_label: "Provider packages"
-title: "Astro Runtime Provider packages"
-id: runtime-providers-versions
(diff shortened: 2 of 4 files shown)


odaneau-astro:
@lzdanski @jwitz: I would appreciate it if you could help me figure out where to put this in our docs since we don't really have an Astro Runtime section for it.

BasPH (commented):
Nice! This would be a helpful reference. I'm missing the page in the menu. How about under Reference -> Astro Runtime? You can add that in docs/sidebarsAstro.js.

jwitz (commented):
This is looking great @odaneau-astro ! We've needed this resource for a long time. One logistical question before we proceed:

I'd prefer if the docs team wasn't responsible for of maintaining and running the Python script. Is there someone on the Runtime team who could commit to running this script as part of a workflow and sharing the results either in #releases or in a docs PR for all Runtime releases?

odaneau-astro:
That's good feedback. I shared the script above as a one-off thing not knowing what tools/processes already existed to update dynamic Docs content like this one. We can certainly loop in the Runtime team to formalize this. I'll ask in Slack and link the thread here.

jwitz:
But we should hold off merging until Jed's finalized his solution for publishing provider versions on slack.

odaneau-astro:
@jedcunningham bumping this thread

jwitz:
Closing out in favor of https://github.com/astronomer/docs/pull/3754 :D

PR #3053, opened 2023-11-02
Updates to top-level code explanation in best practices

### LINKED ISSUE #2984: A Doc for when Laminar Creates a SlowDagParsing Incident
### Issue Type

### Links to Affected Docs

I'm working on some user-facing text here: https://github.com/astronomer/laminar/pull/1281, and there'a fork in my road:

1. The incident should link to the most relevant doc that I can find, which in this case is: https://docs.astronomer.io/learn/dag-best-practices#avoid-top-level-code-in-your-dag-file
2. The incident should link to a doc created specifically its type

I'm uncertain.  If you prefer 2, please speak up in [#project-deployment-health](https://astronomer.slack.com/archives/C05UTPD574K) or create an issue [here](https://github.com/astronomer/laminar/issues), and we'll change it to link to the new doc.

For now I'll assume that 1 is the preference.

### The Issue or Context

I'm considering sending users to this page when `dag_processing.total_parse_time` ([documented here](https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/logging-monitoring/metrics.html)) exceeds 60s.

#### Primary Concern

It's not clear what "top-level code" means.  For instance, consider this example:
```python3
@dag(...)
def the_dag():
    
    @task
    def do_thing():
        ...

    num_of_things = call_external_system()    # this is "top level code" in this case
    # ... because it runs at dag-parse-time.  The doc doesn't make this clear.
    
    
    chain(do_thing() for _ in range(num_of_things))

my_dag = the_dag()
```

I think it would be more useful to distinguish between:

- code that runs at dag-parse-time
- code that runs at task-run-time

There's some more conversation about this [here](https://astronomer.slack.com/archives/C01UJJEN0P3/p1697821591586729?thread_ts=1697748718.646129&cid=C01UJJEN0P3).

#### Secondary Concern

In the incident text, I say:

> To see how long each DAG is taking to parse, use the following commands:
>
>    astro dev start      # start a DAG development environment
>    astro dev bash       # get a shell
>   airflow dags report  # ask airflow about the DAGs
>   astro dev stop       # clean up

I think it's relevant for somebody who is experiencing dag parse slowness to be told how to discover which dag is being slow, but maybe I should be linking to a doc instead?

#### A Nit Pick

The doc claims that:

> In the Bad practice example the connection to the other database is made outside of an operator instantiation as top-level code. When the scheduler parses this DAG, it will use the hook and result variables to query the grocery_list table. **This query is run on every Scheduler heartbeat**, which can cause performance issues.

I am not under the impression that dag parsing is so intimately linked to the scheduler processing loop as that.  Some default config values:
- `AIRFLOW__SCHEDULER__MIN_FILE_PROCESS_INTERVAL` = 30
- `AIRFLOW__SCHEDULER__SCHEDULER_HEARTBEAT_SEC` = 5

The fact that these are different numbers makes me think that they run at different cadences.  They're not independent though, if your scheduler doesn't have enough muscle and it's working too hard at processing dags, the scheduling loop may indeed run at reduced performance.

### Fix or Proposed Change

Either tweak the article with the expectation that users are coming there after being notified that their dags are taking too long to parse, or suggest a better place to send them in this case.

### Additional Notes

If you decide that what is needed is a dedicated article per incident type, I'm happy to contribute content to those articles as we come up with new incident types--but you have a better idea of the big picture re: docs, so i'd want to do so with your guidace.

### Required Reviewers

_No response_

kentdanas:
Resolves #2984 

Updated top-level code description to be clearer about dag-parse-time vs task-run-time and added another example. Also removed erroneous scheduler heartbeat mention (thought we had done that a while ago, must have missed one).

@MatrixManAtYrService you mind taking a look here and letting me know if this resolves your concerns on this particular doc?

### DIFF
--- a/learn/dag-best-practices.md
+++ b/learn/dag-best-practices.md
@@ -93,13 +93,32 @@ When a last modified date is unavailable, a sequence or incrementing ID can be u
 
 ### Avoid top-level code in your DAG file
 
-In the context of Airflow, top-level code refers to any code that isn't part of your DAG or operator instantiations, particularly code making requests to external systems.
+In the context of Airflow, top-level code refers to any code that is run at the time the DAG is parsed, as opposed to the time the task is run.
 
-Airflow executes all code in the `dags_folder` on every `min_file_process_interval`, which defaults to 30 seconds. You can read more about this parameter in the [Airflow docs](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#min-file-process-interval)). Because of this, top-level code that makes requests to external systems, like an API or a database, or makes function calls outside of your tasks can cause performance issues since these requests and connections are being made every 30 seconds rather than only when the DAG is scheduled to run. 
+Code that is part of an operator or a decorated task is run by Airflow only when the tasks runs, not  when the DAG is parsed. For example, in the following code, `call_external_systems()` is considered top-level code because it runs when the DAG is parsed. `x + y` is not top-level code, because it is part of the task definition and only runs when the task runs.
 
-The following DAG example dynamically generates tasks using the PostgresOperator based on records pulled from a different database. 
+```python
+@dag(...)
+def the_dag():
+    
+    @task
+    def do_thing():
+        x + y
 
-In the **Bad practice** example the connection to the other database is made outside of an operator instantiation as top-level code. When the scheduler parses this DAG, it will use the `hook` and `result` variables to query the `grocery_list` table. This query is run on every Scheduler heartbeat, which can cause performance issues. 
+    num_of_things = call_external_system()    # this is "top level code"
+    
+    
+    chain(do_thing() for _ in range(num_of_things))
+
+my_dag = the_dag()
+
+```
+
+Generally, any code that isn't part of your DAG or operator instantiations and that makes requests to external systems is of concern. Airflow executes all code in the `dags_folder` on every [`min_file_process_interval`](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#min-file-process-interval), which defaults to 30 seconds. Therefore, any code that is run when the DAG is parsed and makes requests to external systems, like an API or a database, or makes function calls outside of your tasks can cause performance issues since these requests and connections are being made every 30 seconds rather than only when the DAG is scheduled to run. 
+
+To see another example, the following DAG example dynamically generates tasks using the PostgresOperator based on records pulled from a different database. 
+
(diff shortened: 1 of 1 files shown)


learn/dag-best-practices.md
```diff
@@ -93,13 +93,32 @@ When a last modified date is unavailable, a sequence or incrementing ID can be u
...
+In the context of Airflow, top-level code refers to any code that is run at the time the DAG is parsed, as opposed to the time the task is run.
 
-Airflow executes all code in the `dags_folder` on every `min_file_process_interval`, which defaults to 30 seconds. You can read more about this parameter in the [Airflow docs](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#min-file-process-interval)). Because of this, top-level code that makes requests to external systems, like an API or a database, or makes function calls outside of your tasks can cause performance issues since these requests and connections are being made every 30 seconds rather than only when the DAG is scheduled to run. 
+Code that is part of an operator or a decorated task is run by Airflow only when the tasks runs, not  when the DAG is parsed. For example, in the following code, `call_external_systems()` is considered top-level code because it runs when the DAG is parsed. `x + y` is not top-level code, because it is part of the task definition and only runs when the task runs.
```
TJaniF, line 98:
```suggestion
Code that is part of an operator or a decorated task is run by Airflow only when the task runs, not when the DAG is parsed. For example, in the following code, `call_external_systems()` is considered top-level code because it runs when the DAG is parsed. `x + y` is not top-level code, because it is part of the task definition and only runs when the task runs.
```

learn/dag-best-practices.md
```diff
@@ -93,13 +93,32 @@ When a last modified date is unavailable, a sequence or incrementing ID can be u
...
+    @task
+    def do_thing():
+        x + y
+
+    num_of_things = call_external_system()    # this is "top level code"
+    
+    
+    chain(do_thing() for _ in range(num_of_things))
+
+my_dag = the_dag()
```
TJaniF, line 113:
```suggestion
the_dag()
```

Just a style suggestion haha feel free to disregard :)

TJaniF (approved):
LGTM, had two minor suggestions :)

kentdanas:
@MatrixManAtYrService friendly bump here

kentdanas:
Going to go ahead and merge this, but anybody feel free to comment again if further changes need to be made!

MatrixManAtYrService:
@kentdanas Sorry I lost your notification in the pile, I've since rejiggered my settings, so I'll be more attentive in the future.  This looks fine to me

PR #3058, opened 2023-11-03
Networking docs

### LINKED ISSUE #3057: Reorganize and rename "Connect to external resources" to "Networking
### Issue Type

Inaccurate, misleading, or out-of-date information

### Links to Affected Docs

- https://docs.astronomer.io/astro/category/connect-to-external-resources
- https://docs.astronomer.io/astro/connect-aws
- https://docs.astronomer.io/astro/connect-azure
- https://docs.astronomer.io/astro/connect-gcp
-

### The Issue or Context

Referring to these actions as "connections" is misleading and makes it difficult for users to distinguish between Networking actions and creating Airflow Connections. 

Renaming these pages and the section they are consolidated under to "Networking" will help clarify and improve findability.

### Fix or Proposed Change

- Create new "Networking" section in docs

### Additional Notes

See proposed IA described by [project doc](https://www.notion.so/astronomerio/Doc-improvement-Administration-Sidebar-80c0dfc059e4429a8438130926faef89?pvs=4)

### Required Reviewers

_No response_

lzdanski:
Resolves #3057

### DIFF
--- a/astro/connect-aws.md
+++ b/astro/connect-aws.md
@@ -10,7 +10,7 @@ import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 import {siteVariables} from '@site/src/versions';
 
-Use this document to learn how you can connect an Astro cluster and its Deployments to your external AWS resources.
+Use this document to learn how you can grant Astro cluster and its Deployments access to your external AWS resources.
 
 ## Connection options
 
--- a/astro/connect-gcp.md
+++ b/astro/connect-gcp.md
@@ -10,7 +10,7 @@ import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 import {siteVariables} from '@site/src/versions';
 
-Use this document to learn how you can connect an Astro cluster and its Deployments to your external GCP resources.
+Use this document to learn how you can grant an Astro cluster and its Deployments access to your external GCP resources.
 
 ## Connection options
 
--- a/sidebarsAstro.js
+++ b/sidebarsAstro.js
@@ -216,10 +216,10 @@ module.exports = {
             "resource-reference-hosted",
             {
               type: "category",
-              label: "Connect to external resources",
+              label: "Networking",
               link: {
                 type: 'generated-index',
-                title: 'Connect clusters',
+                title: 'Networking',
                 description: 'Connect Astro to your existing cloud resources.'
               },
               items: ["connect-aws", "connect-azure", "connect-gcp"],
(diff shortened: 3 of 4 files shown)


astro/connect-aws.md
jwitz, line 1: I think the title of these docs should be "Create a network connection between Astro and [cloud]".

sidebarsAstro.js
```diff
@@ -216,10 +216,10 @@ module.exports = {
             "resource-reference-hosted",
             {
               type: "category",
-              label: "Connect to external resources",
+              label: "Networking",
               link: {
                 type: 'generated-index',
```
jwitz, line 221: We had this generated index at a time when Astro only supported one cloud connection at a time. It was also because we needed something to generally link for "networking" instead of linking three cloud specific docs each time we mention the feature. 

With Hosted, you can now have clusters across multiple clouds. Additionally, we're planning to create an "Overview" doc for this topic that applies to all clouds. In that case, does it make sense to make this a normal menu without an index page?

sidebarsAstro.js
```diff
@@ -216,10 +216,10 @@ module.exports = {
             "resource-reference-hosted",
             {
               type: "category",
```
jwitz, line 218: Is it within the scope of this PR to also add "[Authorize Deployments to cloud resources](https://docs.astronomer.io/astro/authorize-deployments-to-your-cloud)" to this sidebar? I remember Rocco saying that this was needed because most users would want to configure this as a "next step", but it's in a different submenu.

jwitz (changes requested):
Base changes look good, I think we can go a little further with this PR tho - added some more suggestions for quick wins

sidebarsAstro.js
lzdanski: I think that's going to be out of scope. It's also somewhat associated with **Secrets backends** so it might have more consequences than a simple move.

astro/connect-gcp.md
```diff
@@ -1,16 +1,16 @@
 ---
 sidebar_label: 'GCP'
-title: 'Connect Astro to GCP data sources'
+title: 'Create a network connection between Astro and GCP'
 id: connect-gcp
-description: Connect Astro to GCP.
+description: Create network connection to Google Cloud Platform.
```
jwitz, line 5:
```suggestion
description: Create a network connection to Google Cloud Platform.
```

sidebarsAstro.js
```diff
@@ -216,13 +216,12 @@ module.exports = {
...
-                type: 'generated-index',
-                title: 'Connect clusters',
-                description: 'Connect Astro to your existing cloud resources.'
-              },
-              items: ["connect-aws", "connect-azure", "connect-gcp"],
+              label: "Networking",
+              items: [
+                "connect-aws", 
+              "connect-azure", 
+              "connect-gcp"
```
jwitz, line 223:
```suggestion
                "connect-azure", 
                "connect-gcp"
```

PR #3066, opened 2023-11-06
clarify passwordless criteria

ophie200:


fritz-astronomer (approved):

astro/kubernetespodoperator.md
```diff
@@ -217,7 +217,7 @@ KubernetesPodOperator(
 
 :::info
 
-This setup is available only on Astro Hosted dedicated clusters and Astro Hybrid. It is not available on Astro Hosted standard clusters.
+Passwordless setup is available only on Astro Hosted dedicated clusters and Astro Hybrid. For Astro Hosted standard clusters, please follow the steps in [Private Registry](https://docs.astronomer.io/astro/kubernetespodoperator?tab=PrivateRegistry#step-1-create-a-kubernetes-secret) to create a Kubernetes secret containing your registry credentials.
```
jwitz, line 220:
```suggestion
Policy-based setup is available only on Astro Hosted dedicated clusters and Astro Hybrid. To run images from a private registry on Astro Hosted standard clusters, follow the steps in [Private Registry](kubernetespodoperator.md?tab=PrivateRegistry#step-1-create-a-kubernetes-secret).
```
Is this wording okay? Suggesting because we haven't mentioned "passwordless" elsewhere in the topic, so I'm not sure whether users will immediately understand what this is in reference to. Feel free to not implement that change and let me know if it's not quite right.

fritz-astronomer: to also be pedantic - it's also not a password (though that was the wording we were using in the slack thread talking about this). It's a secret id+key. I don't have great wording to describe it all succinctly.

jwitz (approved):

Issue #3047, opened 2023-11-01
Use case doc showing how to move large files on Astro


kentdanas:
### Content type

Examples

### Request type

New guide for a specific use case (for example, how to integrate Airflow with a specific 3rd party tool)

### Describe your Learn request.

More info in this slack thread: https://astronomer.slack.com/archives/C015V2JFKT5/p1698798727436949

Based on customer feedback, we don't have any resources for creating DAGs that move large datasets around, and how to utilize Astro features for this. Should most likely show using the SDK and managing the worker(s) accordingly, but will investigate.

### Reviewer

Fritz, Ryan Hatter

kentdanas:
No longer relevant with Object Storage feature

PR #3070, opened 2023-11-08
Astro use case - Moving large files and scaling with worker queues

TJaniF:
Resolves: #3047 

This is a POC for Astro related use cases and will probably change in format and location :)

Code examples/documentation on how to move very large files between object storages is something that was requested by a customer and this is my attempt at a solution that is as general as possible, it should be usable no matter what format and size the files are and adaptable to other object storages than S3 by changing the 4 functions in the DAG file.

Also: please help me find a better title 😅


astro/astro-use-case-large-file-transfer.md
```diff
@@ -0,0 +1,572 @@
+---
+title: "Move large files between object storages with dynamic task mapping"
+navTitle: "Parallelized large file transfer"
+id: astro-use-case-large-file-transfer
+sidebar_label: "Parallelized large file transfer"
+---
+
+Moving large files is a common ETL/ELT operation that necessitates a parallelized approach and appropriate scaling of your Astro deployment. This use case shows a custom templated task group that moves any number of large files from one S3 bucket to another and can be adjusted to work with other object storage providers. Additionally relevant scaling parameters are discussed.
```
kentdanas, line 8: I think we should be clear in the intro that this use case is for how to implement this on Astro specifically, as opposed to any Airflow deployment.

kentdanas: Or, now that I read a little further, be clear that some of it applies anywhere, but some is presented in the context of running it in production on Astro. Or something like that

astro/astro-use-case-large-file-transfer.md
```diff
@@ -0,0 +1,572 @@
...
+
+![Screenshot of the Airflow UI Graph view of a DAG with 8 tasks, 7 of which are contained in the `load_big_files` task group.](/img/examples/astro-use-case-large-file-transfer_task_group_dag.png)
+
+## Before you start
+
+Before trying this example, make sure you have:
```
kentdanas, line 14: Add an Astro deployment here?

astro/astro-use-case-large-file-transfer.md
```diff
@@ -0,0 +1,572 @@
...
+This use case shows how to move large files from one S3 bucket to another. To adjust the code for other object storage solutions you will need to change the hook and methods used in the following functions:
+
+- `get_file_size_func`: retrieves the size of the file to be moved in bytes.
+- `initialize_multipart_upload_func`: initializes a multipart upload.
+- `move_chunk_func`: moves a file chunk from the source to the destination.
+- `complete_multipart_upload_func`: completes a multipart upload.
+
+:::
+
+## Run the project locally
```
kentdanas, line 113: I think if we're going to frame this as an Astro use case, we shouldn't have a run locally option. Or, if we decide to put this in Learn use cases, maybe frame the astro part as how to run in production? Let's chat through that

astro/astro-use-case-large-file-transfer.md
```diff
@@ -0,0 +1,572 @@
...
+        },
+    ]
+
+    return file_config_dicts
+```
+
+Next, a few key parameters are specified. 
+
+- `CHUNK_SIZE` determines the size of the chunks the file is split into. The default value of 5MB is a good starting point, but you can adjust it to your needs.
+- `LARGE_QUEUE` is the name of the worker queue that will handle the resource intensive moving of the file chunks. This queue needs to be configured in the Astro UI as described in the [Deploy the project](#deploy-the-project) section.
```
kentdanas, line 210: I think we should be clear in the text why this DAG needs worker queues - so making it a little more astro-feature forward.

kentdanas (commented):
Did a quick first pass - let's chat more, but overall I think if we're going to make it an "astro" use case we can be more astro-forward with how we frame things. But overall I think this is a great example!

astro/astro-use-case-large-file-transfer.md
TJaniF: Yeah that makes sense, maybe even a "Astro features highlighted/used/?" section? It could contain worker queues, an Astro connection, environment variables, horizontal scaling 🤔

astro/astro-use-case-large-file-transfer.md
TJaniF: That makes sense, I removed the run locally section for now. I'd maybe mention in the intro that it is possible to run the project locally to test it but not recommended for large files/large numbers of files (it would work, just take long 😅 ). But yeah lets chat about this later today :)

kentdanas:
No longer relevant with 2.8 features and new Astro best practices structure. We may revisit in another form (blog post, etc.) later

PR #3104, opened 2023-11-20
Document DAG trigger payload + example code

BasPH:
The DAG trigger feature is still in private preview, so this PR adds a warning to the docs.

This PR also changes some details about the new DAG trigger channel to align with how it's displayed in the Astro UI. Additionally, it documents the payload structure and shows an example DAG implementation.

### DIFF
--- a/astro/alerts.md
+++ b/astro/alerts.md
@@ -103,7 +103,7 @@ This feature is in [Private Preview](https://docs.astronomer.io/astro/feature-pr
 
 The **DAG Trigger** communication channel works differently from other communication channel types. Instead of sending a pre-formatted alert message, Astro makes a generic request through the Airflow REST API to trigger a DAG on Astro. You can configure the triggered DAG to complete any action, such as sending a message to your own incident management system or writing data about an incident to a table.
 
-1. Create the DAG that you want to trigger on an alert. Here's an example of the payload and DAG implementation:
+1. Create a DAG that you want to run when the alert is triggered. For example, you can use the following DAG to run arbitrary Python code when the alert is triggered:
 
   ```python
   import datetime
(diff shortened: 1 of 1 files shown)


astro/alerts.md
```diff
@@ -95,10 +95,48 @@ No external configuration is required for the email integration. Astronomer reco
...
 
-1. In the Workspace where you want to configure the communication channel, create and deploy the DAG that you want to trigger. You can deploy the DAG to any Deployment in the Workspace even if the alert is not applied to that Deployment.
-2. Create a [Deployment API token](deployment-api-tokens.md) for the Deployment where you deployed the DAG. Copy the token to use in the next step.
+This feature is in [Private Preview](https://docs.astronomer.io/astro/feature-previews). Please reach out to your customer success manager to enable this feature.
+
+:::
+
+The **DAG Trigger** communication channel works differently from other communication channel types. Instead of sending a pre-formatted alert message, Astro makes a generic request through the Airflow REST API to trigger a DAG on Astro. You can configure the triggered DAG to complete any action, such as sending a message to your own incident management system or writing data about an incident to a table.
+
+1. Create the DAG that you want to trigger on an alert. Here's an example of the payload and DAG implementation:
```
jwitz, line 106:
```suggestion
1. Create a DAG that you want to run when the alert is triggered. For example, you can use the following DAG to run arbitrary Python code when the alert is triggered:
```

astro/alerts.md
```diff
@@ -95,10 +95,48 @@ No external configuration is required for the email integration. Astronomer reco
...
+  
+          # Example:
+          failed_dag = params["dagName"]
+          print(f"Register an incident in my system for DAG {failed_dag}.")
+  
+      PythonOperator(task_id="register_incident", python_callable=_register_incident)
+
+  ```
+
+2. Deploy the DAG to any Deployment in the same Workspace as the alert will be triggered it. The DAG to alert on, and the DAG to be triggered by the alert, can be deployed in different Deployments, but must be deployed within the same Workspace.
```
jwitz, line 138:
```suggestion
2. Deploy the DAG to any Deployment in the Workspace where you want to create the alert. The DAG that triggers the alert and the DAG that the alert runs can be in different Deployments, but they must be deployed in the same Workspace.
```

astro/alerts.md
```diff
@@ -95,10 +95,48 @@ No external configuration is required for the email integration. Astronomer reco
...
+3. Create a [Deployment API token](deployment-api-tokens.md) for the Deployment where you deployed the DAG that will be triggered. Copy the token to use in the next step.
```
jwitz, line 139:
```suggestion
2. Create a [Deployment API token](deployment-api-tokens.md) for the Deployment where you deployed the DAG that the alert will run. Copy the token to use in the next step.
```

jwitz (approved):
LGTM! Had no idea this was in private preview 😞 I'd even be up for removing the topic/ release note for now if we don't want this to have a large reach?

PR #3128, opened 2023-11-28
Add note about Entra limitation

jwitz:


astro/install-azure.md
```diff
@@ -51,6 +51,12 @@ An _Organization_ is the highest management level on Astro. An Organization cont
     ![The Azure SSO URL and the 'copy to clipboard' button](/img/docs/azure-sso.png)
     
     Share this URL with anyone at your company who needs to access your newly created Organization. Any users that access Astro with this URL will automatically be added to your Organization as an Organization Member. You can add them to your Workspace from Astro’s Cloud UI so they can start deploying code and running DAGs. See [Manage users in your Astro Workspace](https://docs.astronomer.io/astro/manage-workspace-users).
+
+    :::info 
+
+    Microsoft Entra ID is automatically configured only for the email domain you used to create your Astro Organization. If you need to invite users with emails from other domains to Astro, contact [Astronomer support](https://cloud.astronomer.io/open-support-request) and provide your Organization ID and the email domain you want to manage through Microsoft Entra ID.
```
conorbev, line 57: They actually can invite other users with different emails who would login via a password or google/github auth. Like e.g. say they were working with a contracting firm and wanted to invite them in they could do so. They only need to contact us in the case of other email domains that should auth through their same Azure setup. Hope it makes sense?

Therefore could we tweak this to say something like: If your Azure tenant has other email domains which you would like to be able to login via Azure, contact ...

conorbev (changes requested):
Left one comment

astro/install-azure.md
jwitz, line 57:
```suggestion
    Microsoft Entra ID is automatically configured only for the email domain you used to create your Astro Organization. By default, you can invite users from other domains to log in to Astro using passwords, GitHub auth, or Google auth.
    
    If you need to invite users with emails from other domains to Astro using Entra ID, contact [Astronomer support](https://cloud.astronomer.io/open-support-request) and provide your Organization ID and the email domain you want to manage through Entra ID. 
```

lzdanski (approved):

conorbev (approved):
lgtm

Issue #3136, opened 2023-11-30
NFS does not work when ClusterRoles are disabled


jwitz:
### Issue Type

Inaccurate, misleading, or out-of-date information

### Links to Affected Docs

https://docs.astronomer.io/software/deploy-nfs
There might be others as well that are affected when ClusterRoles are disabled

### The Issue or Context

We should document when features are not supported in this circumstance

### Additional Notes

https://kubernetes.io/docs/reference/access-authn-authz/rbac/

PR #3149, opened 2023-12-04
Add note on NFS deploy limitations

jwitz:
Resolves https://github.com/astronomer/docs/issues/3136

### DIFF
--- a/software/namespace-pools.md
+++ b/software/namespace-pools.md
@@ -290,7 +290,7 @@ astronomer:
         value: true
 ```
 
-## Troubleshooting
+## Troubleshooting namespace pools
 
 ### My Deployment is in an unknown state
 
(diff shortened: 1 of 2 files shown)


software/namespace-pools.md
```diff
@@ -312,3 +312,7 @@ This error shows that the service accounts couldn't be created in the pre-create
 ### My namespace is not returning to the pool
 
 If you're not using hard deletion, it can take several days for pre-created namespaces to become available after the associated Deployment is deleted. To enable hard deletes, see [Delete a Deployment](configure-deployment.md#delete-a-deployment).
+
+### My Deployments using NFS deploys stopped working
+
+[NFS deploys](deploy-nfs.md) don't work if you both use namespace pools and set `global.clusterRoles` to `false` in your `config.yaml` file.
```
lzdanski, line 318: Do we also need to include what the fix is? Or is this something we can assume the user goes "oh okay" and knows how to fix?

lzdanski (approved):
One clarifying question but seems mostly straightforward!

software/namespace-pools.md
jwitz: It's not really a fix, but more of an incompatibility. Like you have to choose one or the other, there's no fix where you get both the config and NFS deploys

PR #3153, opened 2023-12-04
Update alerts.md -- make email configuration steps more obvious

RNHTTR:
The current links to set up email notifications for Astro are sort of hidden under a different info block. This change makes it a lot more obvious that users need to configure email notifications to leverage Astro Alerts for email.

### DIFF
--- a/astro/alerts.md
+++ b/astro/alerts.md
@@ -19,8 +19,6 @@ Astro alerts requires OpenLineage. By default, every Astro Deployment has OpenLi
 
 :::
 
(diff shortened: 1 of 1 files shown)


jwitz (approved):
Agreed with boosting visibility here @RNHTTR , I just shuffled around the existing notes so that we could still keep the link to Airflow alerts in the overview. If it's all the way in Step 1, people might leave the page before seeing the link. The OL note was taking up a lot of space and probably works better within prerequisites

PR #3177, opened 2023-12-11
Update links and add redirects

lzdanski:
Slack reported broken links


jwitz (approved):

jwitz (commented):
Oh dear, I tried some of our other redirects with anchor links, it seems like Netlify nor our broken link checker identify a link as broken if the base URL exists, even if the anchor link doesn't. I'll see if I can find a Netlify/ Docusaurus configuration that will correctly identify these links as broken, but in the meantime our only hope is to manually update the URL in the docs and hope the bad link hasn't been shared too much in the wild

Issue #3165, opened 2023-12-07
Additional info on security of connection management feature


kentdanas:
### Issue Type

Missing content for an existing feature

### Links to Affected Docs

https://docs.astronomer.io/astro/create-and-link-connections
https://docs.astronomer.io/astro/user-permissions

### The Issue or Context

The field has provided some feedback on additional information that would be helpful for customers around the security of the connections management feature. Currently, there is not information on which roles can access the feature, and how connections are stored, encrypted, and transmitted. These will be common security concerns for many customers, so we should address them directly in the docs.

### Fix or Proposed Change

- [x] - Add connection management feature to the [workspace roles](https://docs.astronomer.io/astro/user-permissions#workspace-roles) permissions table.
- [x] - Add a note to the [Create Airflow connections in the Cloud UI](https://docs.astronomer.io/astro/create-and-link-connections) doc to note that Workspace authors can import/export/view connections (including secret values). This is covered in the import/export connections/variables doc, but will be notable to many customers using the feature, so should be explicitly mentioned in the doc for creating connections as well.
- [ ] - Add specifics on how connections are stored, encrypted, and transmitted, similar to what we have for [env vars](https://docs.astronomer.io/astro/environment-variables#how-environment-variables-are-stored-in-the-cloud-ui) 

### Required Reviewers

Chris Abella

lzdanski:
Additional updates:

- [ ] - Remove `Airflow` from sidebar menu
- [ ] - Rename `Requirements` to `Prerequisites`
- [ ] - Typo on `Celery executors`
- [ ] - Pull out `Public Preview`
- [ ] - Need to add `Org Settings` to docs

PR #3181, opened 2023-12-12
Connection Management - quick fixes and security update

lzdanski:
Address part of #3165 

Includes:
- Adding info about workspace author abilities for conn management
- Added Env Manager/Conn management to User Permissions reference
- Address general typos/inconsistency in doc

Does not yet include:

- Adding details about the "under the hood" connection management processes similar to what we have in the [Env Vars doc](https://docs.astronomer.io/astro/environment-variables#how-environment-variables-are-stored-in-the-cloud-ui)
- Environment secrets fetching information added to a new *Organization Settings* section.

### DIFF
--- a/astro/user-permissions.md
+++ b/astro/user-permissions.md
@@ -68,7 +68,7 @@ The following table lists the specific permissions that each Workspace role has:
 | Push code to Deployments or Astro Cloud IDE projects                                            |                      | ✔️                   | ✔️                     | ✔️                  |
 | Create, update, and delete Astro Cloud IDE projects                                             |                      | ✔️                   | ✔️                     | ✔️                  |
 | Create, update, and delete Astro alerts                                                         |                      | ✔️                   | ✔️                     | ✔️                  |
-| View Airflow connections, variables, plugins, providers, pools, and XComs                       |                      | ✔️                   | ✔️                     | ✔️                  |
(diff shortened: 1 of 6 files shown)


astro/user-permissions.md
```diff
@@ -70,6 +70,7 @@ The following table lists the specific permissions that each Workspace role has:
 | Create, update, and delete Astro alerts                                                         |                      | ✔️                   | ✔️                     | ✔️                  |
 | View Airflow connections, variables, plugins, providers, pools, and XComs                       |                      | ✔️                   | ✔️                     | ✔️                  |
```
kentdanas, line 71: Is there any distinction in user perms between viewing connections in the Airflow UI vs in Astro? I would hope it is the same, but I wonder if it's worth calling that out explicitly since they can be potentially added in both places?

kentdanas (approved):
LGTM - left one high level question, but generally I think this addresses the parts of the request you indicated.

astro/user-permissions.md
lzdanski: I'm not quite sure the best way to do that in this table. I think we make it pretty clear in the actual docs but having trouble thinking of how to communicate that without this being a huge chunk of text.

lzdanski:
@jwitz if this looks good to you - feel free to make changes + merge it after your review!

jwitz (approved):

PR #3194, opened 2023-12-18
Update Code samples to preferred `DAG` import path

TJaniF:
Noticed that there are plans to deprecate importing from `airflow` directly when going through release notes. For 2.8 it looks like only [Dataset](https://github.com/apache/airflow/pull/34586) and [AirflowException](https://github.com/apache/airflow/pull/34512) direct import is deprecated [but this PR](https://github.com/apache/airflow/pull/34600) says there are plans to include `DAG`, so I switched over to the preferred import statement because why not :)

### DIFF
--- a/astro/airflow-api.md
+++ b/astro/airflow-api.md
@@ -250,7 +250,7 @@ This topic has guidelines on how to trigger a DAG run, but you can modify the ex
 
     ```python
     from datetime import datetime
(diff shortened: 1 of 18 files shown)


jwitz (approved):

kentdanas (approved):
👍

PR #3217, opened 2023-12-26
Fix openlineage link

lzdanski:
Link directs to wrong openlineage resource.


lzdanski:
Confirmed correct behavior in preview. merging.

PR #3233, opened 2024-01-02
Explicitly denying on-prem connectivity

stuart23:
Specifically calling out that we do not want to setup and manage connections between Astro and the customers network.

@cdabella thoughts?

### DIFF
--- a/astro/connect-aws.md
+++ b/astro/connect-aws.md
@@ -38,7 +38,7 @@ Dedicated clusters can connect to AWS in the same ways as standard clusters. Add
 - Transit Gateways
 - AWS PrivateLink
 
-If you require a private connection between Astro and AWS, Astronomer recommends configuring a dedicated cluster. See [Create a dedicated cluster](create-dedicated-cluster.md). Transitive connectivity to on-premise networks is also possible through customer managed VPCs; architectures with a demarcation point between Astro and the customers on-premise network are not supported.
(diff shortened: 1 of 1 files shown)


chomipi88:
+1 from me. We should be connecting into a customer's network that has an established connection to on-premise.

astro/connect-aws.md
```diff
@@ -38,7 +38,7 @@ Dedicated clusters can connect to AWS in the same ways as standard clusters. Add
 - Transit Gateways
 - AWS PrivateLink
 
-If you require a private connection between Astro and AWS, Astronomer recommends configuring a dedicated cluster. See [Create a dedicated cluster](create-dedicated-cluster.md).
+If you require a private connection between Astro and AWS, Astronomer recommends configuring a dedicated cluster. See [Create a dedicated cluster](create-dedicated-cluster.md). Transitive connectivity to on-premise networks is also possible through customer managed VPCs; architectures with a demarcation point between Astro and the customers on-premise network are not supported.
```
jwitz, line 41:
```suggestion
If you require a private connection between Astro and AWS, Astronomer recommends configuring a dedicated cluster. See [Create a dedicated cluster](create-dedicated-cluster.md). Transitive connectivity to on-premise networks is also possible through your managed VPCs. However, architectures with a demarcation point between Astro and your on-premise network are not supported.
```

jwitz (approved):
I lost track of this draft, but did a quick look and it LGTM! Assuming there's no further tech review required, I'll go ahead and merge this.

PR #3246, opened 2024-01-05
Refactor sitevariable usage to be more flexible

jwitz:
This PR changes the way we use environment variables around the site. Variables defined in the preprocessor can be used within any type of markdown content, be it raw text or code fences.


jwitz:
Checked a variety of refactored variables to confirm functionality works both in and out of code fences, all mentions of siteVariables have been removed.

PR #3268, opened 2024-01-12
Update _redirects to catch broken anchor links

jwitz:
Test to see whether forcing the redirect catches broken anchor redirects


jwitz:
Doesn't work :( back to the drawing board

PR #3276, opened 2024-01-15
Snowpark tutorial code updates for new whl file and Airflow 2.8

TJaniF:
There were a couple of changes in Airflow 2.8, the Snowpark provider, the Snowpark ML package and I think also the Astro SDK that caused the DAG in this tutorial to not work anymore. 

This PR updates the whl file that is linked and adjusts the code to work with the new versions.


kentdanas (approved):
LGTM!

Issue #3271, opened 2024-01-12
More details on expectations/ limitations of hibernation


jwitz:
### Issue Type

Missing content for an existing feature

### Links to Affected Docs

https://docs.astronomer.io/astro/deployment-settings#hibernate-a-development-deployment

### The Issue or Context

This topic needs more information about:
- What happens to scheduled tasks when a Deployment hibernates, and what this means for Deployment alerts
- What happens to tasks when backfilling is enabled

PR #3278, opened 2024-01-15
Add some more details about task catchup for hibernation

jwitz:
Resolves https://github.com/astronomer/docs/issues/3271

Just some minor copy updates to help users understand the impact of hibernation. I think anything more prescriptive than this should go in its own best practices/ use case doc, probably after private preview is up and we have a better understanding of how customers are using the feature.

### DIFF
--- a/astro/deployment-settings.md
+++ b/astro/deployment-settings.md
@@ -314,7 +314,7 @@ When your hibernation schedule starts:
 - You can't access the Airflow UI for the Deployment.
 - You can't deploy project images or DAGs to the Deployment.
 
-When your hibernation schedule ends, the Deployment will rerun any failed tasks from DAGs with `catchup=true`. To avoid incurring additional resource costs, Astronomer recommends disabling catchup on DAGs in hibernating Deployments.
+When your hibernation schedule ends, the Deployment will start any DAG runs for data intervals that were missed during hibernation for DAGs with `catchup=true`. To avoid incurring additional resource costs, Astronomer recommends disabling catchup on DAGs in hibernating Deployments.
 
 #### Manually hibernate a Deployment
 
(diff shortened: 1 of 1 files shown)


astro/deployment-settings.md
```diff
@@ -309,11 +309,13 @@ When your hibernation schedule starts:
     ![A Deployment with a Hibernating status on the Deployments page of the Cloud UI](/img/docs/hibernating-status.png)
 
 - Any task that was previously running will be killed and marked as failed.
-- Tasks and DAGs do not run. Scheduled tasks will fail.
+- Tasks and DAGs do not run. Task instances that were already running or scheduled at the time of hibernation will fail and trigger any related notifications.
 - No Deployment resources are available. This includes the scheduler, webserver, and all workers.
 - You can't access the Airflow UI for the Deployment.
 - You can't deploy project images or DAGs to the Deployment.
 
+When your hibernation schedule ends, the Deployment will rerun any failed tasks from DAGs with `catchup=true`. To avoid incurring additional resource costs, Astronomer recommends disabling catchup on DAGs in hibernating Deployments.
```
kentdanas, line 317:
```suggestion
When your hibernation schedule ends, the Deployment will start any DAG runs for data intervals that were missed during hibernation for DAGs with `catchup=true`. To avoid incurring additional resource costs, Astronomer recommends disabling catchup on DAGs in hibernating Deployments.
```
Catchup won't rerun anything that had previously failed. It will just run any DAG runs that would have been scheduled had Airflow been on at the time. The suggested language is closest to what Airflow uses in its docs - open to other suggestions, but generally want to avoid "scheduled"/"skipped" or any words that also have a DAG or task status meaning in Airflow.

kentdanas (approved):
One minor adjustment on catchup behavior wording. Otherwise the update looks good!

lzdanski (approved):

PR #3284, opened 2024-01-17
Add warning about image-only deploys

jwitz:
Based on https://astronomer.slack.com/archives/CL44PA4DB/p1704211917536349


lzdanski (approved):
This kind of ends up being a big block of admonitions. I can see moving the warning to line 34, between the second paragraph and "Run the following command"... to make sense and break things up a bit, but I'll leave it to your discretion! 

Screenshot 2024-01-17 at 10 51 58 AM

PR #3291, opened 2024-01-18
Update create-and-link-connections.md

RNHTTR:


astro/create-and-link-connections.md
```diff
@@ -7,6 +7,11 @@ description: "Create Airflow connections and link them to multiple Deployments i
 
 You can create and manage Airflow connections for Deployments with the Astro Environment Manager in the Cloud UI. The Environment Manager uses an Astro-managed secrets backend to store connection configurations as Kubernetes Secrets.
 
+:::info Note
+
+Astro Environment Manager is only available to Astro Hosted users; it is not available for Astro Hybrid users.
+
+
```
jwitz, line 14:
```suggestion
:::info

The Astro Environment Manager is available only on Astro Hosted

:::
```

jwitz: This is smaller, but it might be more visually distinctive and therefore visible for readers

jwitz (approved):
LGTM @RNHTTR , hope you don't mind but I'd like to try using our Hosted badge HTML element here instead of a full note!

astro/create-and-link-connections.md
RNHTTR: I'm not opposed to the hosted badge, but should it say something more obvious like "Hosted Only"? I'm not sure a Hybrid user would find the badge to obviously imply it's hosted-only.

jwitz: Good point, we can work on creating a separate "Hosted only" badge in the next quarter to apply on docs.

PR #3295, opened 2024-01-18
Update external-python-operator.md to include astro-provider-venv

RNHTTR:


learn/external-python-operator.md
```diff
@@ -209,3 +209,4 @@ Airflow has several other options for running tasks in isolated environments:
...
 - [The PythonVirtualenvOperator](https://registry.astronomer.io/providers/apache-airflow/modules/pythonvirtualenvoperator). This operator works similarly to the ExternalPythonOperator, but it can create new virtual environments for each task. By default, the virtual environments are destroyed after the task succeeds unless a `venv_cache_path` is specified. This operator is ideal if you need to create your virtual environment on an ad-hoc basis or do not need your environment to persist. Downsides include higher task latency when the environment is created from scratch each time the task is run.
 - [The ExternalBranchPythonOperator](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/operators/python/index.html#airflow.operators.python.ExternalBranchPythonOperator) (available in Airflow 2.7+). This operator is an extension of the ExternalPythonOperator and allows you to run a [branching](airflow-branch-operator.md) task in a virtual environment.
+- [astro-provider-venv](https://github.com/astronomer/astro-provider-venv) makes it easy to execute Airflow tasks in virtual environments.
```
kentdanas, line 212: Can we add a sentence here describing when somebody might want to use this option over the others available, or what use cases it applies to, etc.? Similar to how we have ideal situations and downsides briefly described for the other bullets.

kentdanas (changes requested):
@RNHTTR thanks for adding this! One request for a bit of additional info before we merge this.

learn/external-python-operator.md
RNHTTR: Sure. I'm just going to parrot the byline from the readme as I was just casually aware of it. one sec.

kentdanas:
Ah, @RNHTTR sorry, moving too fast when I reviewed this morning. So the main content here already uses the Astro venv provider - see https://docs.astronomer.io/learn/external-python-operator#step-2-configure-your-astro-project, 
```
Update the Dockerfile of your Astro project to install pyenv and its requirements using Astronomer's Docker BuildKit frontend for Airflow:
...
Creating a pyenv environment in your Airflow project requires installing multiple operating system level packages and a series of Docker commands. The open source [Astro BuildKit](https://github.com/astronomer/astro-provider-venv) simplifies this process and allows you to create a Python virtual environment with only two lines of code.
```

So we shouldn't need to add it as another method in the end section. Let me know if you think we need to clarify anything in the steps here.

RNHTTR:
Aha! I think maybe we should make that clearer in the intro. I got the impression that this was just leveraging the raw [ExternalPythonOperator](https://airflow.apache.org/docs/apache-airflow/stable/howto/operator/python.html#externalpythonoperator). If you agree, I'll take a pass at that and update the example from Astro Runtime 7 to 10.

kentdanas:
> Aha! I think maybe we should make that clearer in the intro. I got the impression that this was just leveraging the raw [ExternalPythonOperator](https://airflow.apache.org/docs/apache-airflow/stable/howto/operator/python.html#externalpythonoperator). If you agree, I'll take a pass at that and update the example from Astro Runtime 7 to 10.

That sounds great, thank you!

PR #3296, opened 2024-01-18
KE is now supported for Environment Management connections

iancmoritz:
This line is no longer accurate in docs.


jwitz (approved):
@iancmoritz Was this change shipped in this week's release? Can I see the PR? If so, I'll also backport it to release notes.

PR #3320, opened 2024-01-24
Update product badges

jwitz:


jwitz:
@katiebrady-astro I had assumed that making these images larger would also make them larger in docs, but it they render way smaller on the site and I'm not sure why 😓

katiebrady-astro (approved):
Hey @jwitz ! Looks like the component itself just needed a min-width set. Got that fixed up and all LGTM! 🙌 
image

PR #3324, opened 2024-01-24
Update use-case-airflow-llm-rag-finance.md

mrnitrate:
Remove duplicate call to initialize `weaviate.Client()` in example code.


TJaniF (approved):
Nice job spotting that, thank you! :)

PR #3334, opened 2024-01-26
Add links in docs to knowledge base

jwitz:


src/theme/DocSidebar/Desktop/Content/index.js
```diff
@@ -43,16 +43,26 @@ export default function DocSidebarDesktopContent({ path, sidebar, className }) {
         <ul className={clsx(ThemeClassNames.docs.docSidebarMenu, 'menu__list')}>
           <DocSidebarItems items={sidebar} activePath={path} level={1} />
         </ul>
+        {path.indexOf('/software') > -1 && (
```
jwitz, line 46: @katiebrady-astro I tried to make this link appear conditionally on the Astronomer Software sidebar, but it doesn't look like this condition worked. Is there a way I can get it so that only the "Support Knowledge Base" link appears in the Software sidebar?

katiebrady-astro: Hey @jwitz, I found the issue! Looks like I added a CSS rule at one point to _always_ hide the sidebar bottom nav on `/software` 😅 got that fixed up and rewrote it so the "Support Knowledge Base" link always shows and the others only when _not_ on `/software` 🙌

PR #3366, opened 2024-02-05
Clarify DAG deploy behavior with regard to scripts

jwitz:
https://astronomer.slack.com/archives/C01UA03TPV1/p1706930034469059


jwitz:
Closing out because this is not the right technical solution 🤠

PR #3367, opened 2024-02-05
Add clarification about Python distribution

jwitz:


astro/runtime-image-architecture.md
```diff
@@ -151,6 +151,8 @@ Starting with Astro Runtime 9, Astronomer maintains different distributions Astr
 quay.io/astronomer/astro-runtime:<runtime-version>-python-<python-version>
 ```
 
+Each Python distribution also includes an alternative `quay.io/astronomer/astro-runtime:<version>-base` distribution if you need to further customize the image.
```
jwitz, line 154:
```suggestion
Each Python distribution also has an alternative `quay.io/astronomer/astro-runtime:<version>-base` distribution that you can use to further customize the image.
```

lzdanski (approved):
Does this also need to be added to the most recent software runtime image arch doc? https://docs.astronomer.io/software/runtime-image-architecture#python-version-distributions

PR #3389, opened 2024-02-12
Update Astro use cases to best practices

kentdanas:
More representative IA for what is actually going in this section.


kentdanas:
@jwitz @lzdanski can one of you double check I did the redirects correctly? Surprisingly this is actually the first time I've touched that file

jwitz (approved):
Agreed on the name change!

PR #3396, opened 2024-02-13
Hotfix: Quick clarifying note on DAG-only deploys enabled by default

paolaperaza:


astro/deployment-details.md
```diff
@@ -55,7 +55,7 @@ You can also update your Workspace so that any new Deployments in the Workspace
 
 :::info
 
-When CI/CD enforcement is enabled for a Deployment, you cannot enable or disable [DAG-only deploys](deploy-dags.md) for the Deployment.
+When CI/CD enforcement is enabled for a Deployment, you cannot enable or disable [DAG-only deploys](deploy-dags.md) for the Deployment. For new Deployments on Astro Hosted, DAG-only deploys are enabled by default and you only need to complete these steps if you want to disable the feature.
```
jwitz, line 58: We say "you only need to complete these steps if you want to disable the feature" and then "To enable or disable DAG-only deploys", we should pick what framing to use here because right now it's a contradiction

jwitz:
Closing this out because I'm not sure the sentence helps the clarity here, and also this detail should be more apparent now that DAG deployment is a toggleable feature in the Cloud UI

PR #3403, opened 2024-02-14
Add hotfixes for deployment file formatting

lzdanski:


lzdanski:
confirmed in preview and reviewed via slack

Issue #3410, opened 2024-02-15
Create site variables


jwitz:
- Additional CLI version variables to keep multi-version examples up to date
- Software version variables

PR #3425, opened 2024-02-19
Add reusable variables for CLI versions

jwitz:
Resolves https://github.com/astronomer/docs/issues/3410

Despite being a powerful feature. I couldn't find a lot of places to implement these variables. They're useful for when you have to update small values as a regular part of your workflow (especially numeric values). I almost added variables for Astronomer Software versions, but that wouldn't be sustainable because every new Software version would require updating every instance of the variable for the newly cut directory, which then ruins the purpose of the variable.


archived_software_docs/version-0.28/install-cli.md
```diff
@@ -109,7 +109,7 @@ astro version
 If the installation was successful, you should see the following output:
 
 <pre><code parentName="pre">{`% astro version
-Astro CLI Version: {{CLI_VER}}`}</code></pre>
+Astro CLI Version: {{CLI_VER_LATEST}}`}</code></pre>
```
lzdanski, line 112: Will the variables work in the unmaintained ones? Or should they be hardcoded? Not sure the extent of how powerful this is, but if it only shows up in the final build, we might just want to hardcode for the static files.

lzdanski (approved):
Small question but otherwise LGTM in preview! Stoked to have this functionality.

archived_software_docs/version-0.28/install-cli.md
jwitz: Nope this was accidentally updated from a broad sweeping ctrl f, reverting!

Issue #3409, opened 2024-02-15
Create admonition content blocks


jwitz:
- Public preview blocks
- Private preview blocks
- Astro CLI shoutout blocks
- Software only warning blocks

PR #3426, opened 2024-02-19
New admonitions

jwitz:
Resolves https://github.com/astronomer/docs/issues/3409


src/theme/Admonition/Type/PrivatePreview.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import clsx from 'clsx';
+import Translate from '@docusaurus/Translate';
+import AdmonitionLayout from '@theme/Admonition/Layout';
+import IconWarning from '@theme/Admonition/Icon/Warning';
+const infimaClassName = 'alert alert--warning';
+const defaultProps = {
+  icon: <IconWarning />,
+  children: <div class="admonitionContent">This feature is in <a href="https://docs.astronomer.io/astro/feature-previews" target="_blank" rel="noopener noreferrer">Private Preview</a>. Please reach out to your customer success manager to enable this feature.</div>,
```
jwitz, line 9: @katiebrady-astro Before I start making the rest of these for our reusable admonitions, I wanted to make sure that this looked good design wise with you

astro/reporting-dashboard.md
```diff
@@ -7,10 +7,7 @@ description: "View information about your Organization, Deployments, DAGs, and c
 
 import HostedBadge from '@site/src/components/HostedBadge';
 
-:::warning
-
-This feature is in [Private Preview](https://docs.astronomer.io/astro/feature-previews). Please reach out to your customer success manager to enable this feature.
-
+:::privatepreview
```
jwitz, line 10: @katiebrady-astro This is how a writer would apply the admonition to the page. Do you think there's a way to design this so that a writer only has to add `:::privatepreview` without the closing `:::` in a new line? I suppose there's always `:::privatepreview <br> :::` but it would be great to be able to do it without HTML

katiebrady-astro: Hey @jwitz! Unfortunately I think `markdown-it` requires the closing `:::` for all custom blocks. I agree though that having it on a new line is better than mixing in HTML.

src/theme/Admonition/Type/PrivatePreview.js
katiebrady-astro: Hey @jwitz looks good to me! 🙌

katiebrady-astro (approved):

katiebrady-astro:
@jwitz Just checked again and you actually don't need to include the closing `:::` for the admonition. Sorry for the confusion! 🙇‍♀️  `markdown-it` docs example shows it but seems to work fine without. Not seeing build errors or issues testing locally.

image
image

jwitz:
@katiebrady-astro I thought this as well at first! But when you do only the first opening line, it removes all content from the docs until the next header appears. In your example, you can see from the screenshot that "Astro provides dashboards..." doesn't appear in the rendered documentation.

katiebrady-astro:
> @katiebrady-astro I thought this as well at first! But when you do only the first opening line, it removes all content from the docs until the next header appears. In your example, you can see from the screenshot that "Astro provides dashboards..." doesn't appear in the rendered documentation.

oh right I totally missed that, sorry for double confusion! 😅 then yeah seems like we'll need to stick with just adding the closing `:::`.

src/theme/Admonition/Type/CliAstroAndSoftware.js
jwitz, line 1: @lzdanski The core content to review in this PR are the files in this `Type` folder, can you please review:

- The content of the new admonitions
- Whether there's any other admonition types you think we should tackle in this PR?

src/theme/Admonition/Type/CliAstroAndSoftware.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import clsx from 'clsx';
+import Translate from '@docusaurus/Translate';
+import AdmonitionLayout from '@theme/Admonition/Layout';
+import IconInfo from '@theme/Admonition/Icon/Info';
+const infimaClassName = 'alert alert--info';
+const defaultProps = {
+  icon: <IconInfo />,
+  children: <div class="admonitionContent">This command is available for both Astro and Astronomer Software.</div>,
```
jwitz, line 9:
```suggestion
  children: <div class="admonitionContent">This command is available for both Astro and Astronomer Software. Any difference in usage will be documented in separate tabs for each product. </div>,
```

src/theme/Admonition/Type/PrivatePreview.js
jwitz, line 9:
```suggestion
  children: <div class="admonitionContent">This feature is in <a href="https://docs.astronomer.io/astro/feature-previews" target="_blank" rel="noopener noreferrer">Private Preview</a>. Please reach out to your account team to enable this feature.</div>,
```

PR #3435, opened 2024-02-21
Astronomer Software: improve instructions for dag-deployment and airgapped install

rob-1126:
Correct a wrong configuration path that would prevent dag deployments from being enabled.
Break out optional dag-deployment settings from the initial example of enabling the feature.
Add defaultRuntimeRepository to the standard example configuration for airgapped installations.
Add configuration settings to example so that the dag-downloader sidecar gets included in the list of images customers mirror for airgapped installations.

### DIFF
--- a/software/deploy-dags.md
+++ b/software/deploy-dags.md
@@ -28,7 +28,7 @@ When you [update a Deployment](#configure-dag-only-deploys-on-a-deployment) to s
 
 ## Enable the feature on an Astronomer cluster
 
-By default, DAG-only deploys are disabled for all Deployments on Astronomer Software. To enable the feature, set `global.dagOnlyDeployment.enabled` to `true` in your `config.yaml` file, e.g.:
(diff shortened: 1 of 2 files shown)


software/deploy-dags.md
```diff
@@ -28,22 +28,32 @@ When you [update a Deployment](#configure-dag-only-deploys-on-a-deployment) to s
 
 ## Enable the feature on an Astronomer cluster
 
-By default, DAG-only deploys are disabled for all Deployments on Astronomer Software. To enable the feature, add the following configuration to your `config.yaml` file:
+By default, DAG-only deploys are disabled for all Deployments on Astronomer Software. To enable the feature, set `global.dagOnlyDeployment.enabled` to `true` in your `config.yaml` file, e.g.:
```
jwitz, line 31:
```suggestion
By default, DAG-only deploys are disabled for all Deployments on Astronomer Software. To enable the feature, set `global.dagOnlyDeployment.enabled` to `true` in your `config.yaml` file:
```

PR #3436, opened 2024-02-21
Hotfix: Fix header + add redirects

lzdanski:
Per slack [discussion](https://astronomer.slack.com/archives/C015V2JFKT5/p1708460086966649)

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -352,6 +352,7 @@
 /astro/deployment-settings#transfer-a-deployment-to-another-workspace /astro/transfer-a-deployment 301
 /runtime-image-architecture#default-environment-variables /astro/platform-variables.md#default-runtime-environment-variables 301
 /astro/resource-reference-hosted#worker-type /astro/deployment-resources#hosted-worker-types 301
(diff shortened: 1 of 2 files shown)


static/_redirects
```diff
@@ -352,6 +352,7 @@
 /astro/deployment-settings#transfer-a-deployment-to-another-workspace /astro/transfer-a-deployment 301
 /runtime-image-architecture#default-environment-variables /astro/platform-variables.md#default-runtime-environment-variables 301
 /astro/resource-reference-hosted#worker-type /astro/deployment-resources#hosted-worker-types 301
+/astro/resource-reference-hosted#deployment-resources /astro/deployment-resources 301
```
jwitz, line 355: This redirect will not work because it's for a header :(

jwitz (approved):
Content change LGTM but we should consider making a reusable module for the table and adding it back here later!

static/_redirects
jwitz: Keeping it here in case we one day have the ability to configure header redirects

PR #3440, opened 2024-02-22
Add max node clarification

### LINKED ISSUE #3400: Clarify max node setting in cluster config docs
### Issue Type

Inaccurate, misleading, or out-of-date information

### Links to Affected Docs

https://docs.astronomer.io/astro/manage-hybrid-clusters#configure-maximum-node-count

### The Issue or Context

It would be good to update the docs to clarify that max node count = 15 is reasonable but max node count = 20 would cover a worse case scenario.

### Fix or Proposed Change

Add clarification to doc

### Additional Notes

https://astronomer.slack.com/archives/C015V2JFKT5/p1707764174953009

### Required Reviewers

_No response_

lzdanski:
Resolves #3400


astro/manage-hybrid-clusters.md
```diff
@@ -213,7 +213,7 @@ For example, consider the following configurations within a cluster:
 - 1 of the 3 worker queues has a maximum worker count of 10.
 - 2 of the 3 worker queues have a maximum worker count of 5.
 
-In this scenario, the maximum node count for the `m5.2xlarge` node pool in your cluster must be equal to or greater than 15 to make sure that each worker queue can scale to its limit.
+In this scenario, the maximum node count for the `m5.2xlarge` node pool in your cluster must be equal to or greater than 15 to make sure that each worker queue can scale to its limit. However, setting your node pool to 20 for this example would be appropriate if you expect the cluster frequently reach maximum load.
```
jwitz, line 216:
```suggestion
In this scenario, the maximum node count for the `m5.2xlarge` node pool in your cluster must be equal to or greater than 20 to make sure that each worker queue can scale to its limit. 
```
I might be reading this wrong, but isn't this just a fix to some incorrect math? Perviously, we say:

>Maximum node count for a node pool must always be equal to or greater than the sum of all maximum possible workers across all worker queues that are configured with that worker node type.

In the example we provide, the "sum of all maximum possible workers across all worker queues" would be 20, so I don't think we would even allow 15 to be configured in the first place.

Also if it does end up being 20, maybe we should change the example numbers so that the calculated max node figure is different from the default of 20

jwitz (changes requested):
One technical question, I'm not quite sure why we're still listing the 15 figure

astro/manage-hybrid-clusters.md
lzdanski: @ophie200 - Can you give us some clarification here? Is it possible to have a smaller number than the total maximum worker counts (sum of maximum worker counts per each worker queue)?

lzdanski: @ophie200 - Clarification question:

Does Maximum worker count need to equal the node pool size? I think that's what's tripping us up here in clarifying this.  If max worker count = max node pool size, then this is wrong and it should be a total of 20 instead of 15.

jwitz: So basically, we want to document two numbers:
- "20" which is the _actual_ limit
- "Less than 20" which is what you can set if you don't expect Deployments to hit max worker count that often

I updated my suggestion to reflect this, after that it's good to merge

PR #3450, opened 2024-02-26
Update LLMOps use case to newest Airflow and Weaviate version

TJaniF:
As discussed on Friday :)


kentdanas (approved):
Looks good! Assume you also updated the repo to go along with it :)

Issue #2628, opened 2023-07-31
Add links to OSS documentation sets from product Astronomer


jwitz:
### Issue Type

Missing content for an existing feature

### Links to Affected Docs

All docs - might be a sidebar issue

### The Issue or Context

We don't have many visible links to Astronomer's OSS projects, such as [Cosmos](https://astronomer.github.io/astronomer-cosmos/#) and the [Astro SDK](https://astro-sdk-python.readthedocs.io/en/stable/). 

### Fix or Proposed Change

It's up for debate where we should put these. A few options are:

- Next to "Office Hours" in the sidebar
- In the bottom sidebar
- As another menu in the left sidebar

Tagging @katiebrady-astro on this issue for her guidance on whether the first two are doable

PR #3522, opened 2024-03-14
Add external docs to top navbar

jwitz:
Resolves https://github.com/astronomer/docs/issues/2628


jwitz:
@katiebrady-astro I'm trying to add a nice dropdown arrow to the right of the "more" section similar to this, do we have any standard assets for that sort of thing? 
Screenshot 2024-03-14 at 2 09 30 PM

PR #3523, opened 2024-03-15
Change the highlighted links on the Learn overview page

TJaniF:
Based on our docs review session. I think the restructuring of this page needs a larger discussion but making this PR as a more immediate fix to highlight more relevant links :)

Very open to suggestions of other guides to link and whether you think having the trial so prominently makes sense.


kentdanas (approved):
I like it. Definitely an improvement over current, and I think best not to spend too much time debating what goes on here until we figure out if a bigger redesign is necessary.

PR #3558, opened 2024-03-26
change aws auth mount locations

cpsenka:
Using the /usr/local/airflow mount location leads to a need for some permission changes. Mounting to /home/astro allows the astro user to use these files


jwitz (approved):
Looks good to me, thanks for this improvement!

Issue #3547, opened 2024-03-22
Entra ID SSO and SCIM modifications


chomipi88:
### Issue Type

Inaccurate, misleading, or out-of-date information

### Links to Affected Docs

https://docs.astronomer.io/astro/set-up-scim-provisioning
https://docs.astronomer.io/astro/configure-idp?tab=ME-ID#configure-your-sso-identity-provider

### The Issue or Context

SCIM and SSO configurations are out of sync. The current SCIM docs recommend provisioning users using the `userPrincipalName` attribute as email. We recently ran into two customer issues where they configure `userPrincipalName` as something entirely different then the mail attribute.

### Fix or Proposed Change

1. Under step 13 for user attribute mappings, change `userPrincipalName` to `mail`
2. Update the warning banner to say `mail`
3. Add to the existing warning banner or create another one to indicate that the SAME Entra ID attribute for the desired email to be used for user profiles has to be configured for both SCIM and SSO 

### Required Reviewers

chomipi88, aliotta

PR #3564, opened 2024-03-26
fixScimMappingsRecomendation

aliotta:
By default our sso integration uses email so we should suggest our users to use this for scim otherwise they will run into sync issues.

### DIFF
--- a/astro/set-up-scim-provisioning.md
+++ b/astro/set-up-scim-provisioning.md
@@ -162,7 +162,7 @@ Complete the manual setup if you configured your existing Astro app without usin
 
   This setup assumes that `mail` contains your users' email. If you use a field other than `mail` to define your user email, replace `mail` with the attribute you use.
 
-  Whether you choose to use `mail` or a different attribute, you must use the same Microsoft Entra ID Attribute mfor both your SSO and SCIM configurations.
+  Whether you choose to use `mail` or a different attribute like `userPrincipleName`, you must use the same Microsoft Entra ID Attribute for both your SSO and SCIM configurations.
 
   :::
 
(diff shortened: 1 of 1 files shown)


lzdanski:
Resolves #3547

astro/set-up-scim-provisioning.md
```diff
@@ -152,15 +152,17 @@ Complete the manual setup if you configured your existing Astro app without usin
...
+    | mail                                                        | userName        |
     | Switch([IsSoftDeleted], , "False", "True", "True", "False") | active          |
     | displayName                                                 | displayName     |
 
   :::warning
 
-  This setup assumes that `userPrincipalName` contains your users' email. If you use a field other than `userPrincipalName` to define your user email, replace `userPrincipalName` with the attribute you use.
+  This setup assumes that `mail` contains your users' email. If you use a field other than `mail` to define your user email, replace `mail` with the attribute you use.
+
+  Whether you choose to use `mail` or a different attribute, you must use the same Microsoft Entra ID Attribute mfor both your SSO and SCIM configurations.
```
aliotta, line 165: typo mfor

aliotta, line 165: It is very common for users to use userPrincipleName as their alt property maybe we want to be explicit here? I know of no other property a customer would use tbh

aliotta: i.e. instead of saying "a different attribute" say if you want to use upn instead of mail for your source of truth for username

lzdanski, line 165:
```suggestion
  Whether you choose to use `mail` or a different attribute like `userPrincipleName`, you must use the same Microsoft Entra ID Attribute for both your SSO and SCIM configurations.
```

lzdanski (approved):

PR #3587, opened 2024-04-01
Update authorize-workspaces-to-a-cluster.md

RNHTTR:
"dedicated" could cause confusion between Hosted Dedicated/Hosted Standard clusters


jwitz (approved):
Nice catch

PR #3603, opened 2024-04-04
Link to max_map_length config

BasPH:
I suggest linking to the exact Airflow config so that the reader doesn't have to search/scroll.


TJaniF (approved):
Thank you! :)

PR #3613, opened 2024-04-05
Add alternative to use Jinja template which is processed at runtime not parse time

manmeetkaur:
While reviewing the DAGs for a customer noticed that Variable.get is used in the operator as a param. But still it was parsed and was throwing an error in the Airflow UI. If we use Jinja template, it will be only be evaluated at runtime as it is just a string for the parser and will also avoid hitting the DB every couple of minutes/seconds during parsing.


kentdanas (approved):
Awesome, thanks @manmeetkaur!

PR #3617, opened 2024-04-08
Automate API changelog generation

jwitz:
This PR creates a new changelog for the Astro API. The general workflow is as follows:

1. A weekly OpenAPI spec generator runs every Wednesday, which is typically after the production Astro release is cut. The action pulls the new OpenAPI specs into our docs and generates PRs for the changes.
    - If the Astro release is delayed, a writer can also manually trigger the spec generation for whenever the API changes release. 
2. In the PR, Fern's bot will submit a summary of changes between the existing OpenAPI specs and the PR'd OpenAPI specs. This is the basis for the week's changelog.
3. A writer will review the generated changelog and add its content to `api/fern/docs/pages/changelog.mdx`, following the pattern of previous weeks. This change could be committed to the same PR that the GitHub Action generated, or it could be PR'd separately.
4.  After review of the changelog is complete, the writer merges both the updated spec changes and the changelog content.

### DIFF
--- a/api/fern/docs.yml
+++ b/api/fern/docs.yml
@@ -11,6 +11,8 @@ navigation:
         path: ./docs/pages/get-started.mdx
       - page: Versioning and support
         path: ./docs/pages/versioning-and-support.mdx
+      - page: Changelog
+        path: ./docs/pages/changelog.mdx
   - api: Platform API Reference
     api-name: platform
   - api: IAM API Reference
--- a/api/fern/docs.yml
+++ b/api/fern/docs.yml
@@ -11,8 +11,6 @@ navigation:
         path: ./docs/pages/get-started.mdx
       - page: Versioning and support
         path: ./docs/pages/versioning-and-support.mdx
-      - page: Changelog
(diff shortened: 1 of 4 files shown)


api/fern/docs.yml
```diff
@@ -11,6 +11,8 @@ navigation:
         path: ./docs/pages/get-started.mdx
       - page: Versioning and support
         path: ./docs/pages/versioning-and-support.mdx
+      - page: Changelog
+        path: ./docs/pages/changelog.mdx
```
dsinghvi, line 15: @jwitz we'll actually want to create `changelog` folders inside of each https://github.com/astronomer/docs/tree/main/api/fern/apis/iam `api`

dsinghvi: The Fern CLI will automatically register these changelogs and they will show up in your docs

api/fern/apis/iam/changelog/placeholder.mdx
jwitz, line 1: @dsinghvi Needed to add placeholders so that these folders showed up in the diff, is this structure correct?

PR #3631, opened 2024-04-11
Correct code snippet for `airflow-variable copy` command

josh-fell:
A trivial update, but the example code for copying Airflow Variables from one Deployment to another by specifying IDs has an incorrect option name. The `--target` option name should be `--target-id`.


jwitz (approved):
Thanks for jumping in with the fix @josh-fell !

PR #3633, opened 2024-04-11
Test out bearer configs

jwitz:
Do not merge - test only

### DIFF
--- a/api/fern/apis/platform/openapi/openapi.yaml
+++ b/api/fern/apis/platform/openapi/openapi.yaml
@@ -2607,8 +2607,11 @@ components:
       - totalCount
       - workspaces
       type: object
+    JWT:
+      type: http
+      scheme: bearer
 security: 
-  - Bearer: []
+  - JWT: []
 info:
   contact: {}
   description: Astro Platform API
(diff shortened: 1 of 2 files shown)


api/fern/apis/platform/openapi/openapi.yaml
```diff
@@ -2607,10 +2607,6 @@ components:
       - totalCount
       - workspaces
       type: object
-  securitySchemes:
-    Bearer:
```
ashb, line 2611: What if your keep the rest the same but change this and the next to `JWT` instead of `Bearer`?

api/fern/apis/platform/openapi/openapi.yaml
```diff
@@ -2607,12 +2607,11 @@ components:
...
+    JWT:
```
jwitz, line 2610:
```suggestion
securitySchemes:
    JWT:
```

api/fern/apis/platform/openapi/openapi.yaml
```diff
@@ -2607,12 +2607,12 @@ components:
       - totalCount
       - workspaces
       type: object
-  securitySchemes:
-    Bearer:
+securitySchemes:
```
jwitz, line 2610:
```suggestion
  securitySchemes:
```

jwitz:
I actually will go ahead and merge these, as this is the format engineering agreed could possibly be generated at source.

PR #3636, opened 2024-04-12
Simplify HA explanation

jwitz:


astro/deployment-resources.md
```diff
@@ -124,7 +124,7 @@ To configure the scheduler on an [Astro Hybrid](hybrid-overview.md) Deployment:
 
 ## Enable high availability
 
-By default, the Pods running your Deployment's Airflow components are distributed across multiple nodes. When you enable high availability, your Deployment runs two instances of [PgBouncer](https://www.pgbouncer.org/) and two instances of the Airflow Scheduler across different nodes. However, Astro makes a best effort to use different availability zones for your PGBouncer and Scheduler, which means it is possible but unlikely that they are both located in the same availability zone. This ensures that your DAGs can continue to run if there's an issue with one of your Airflow components in a specific node or availability zone.
+By default, the Pods running your Deployment's Airflow components are distributed across multiple nodes. When you enable high availability, Astro runs your nodes in different regions. This ensures that your DAGs can continue to run if there's an issue with one of your Airflow components in a specific node or availability zone.
```
ashb, line 127: I'd say also add something like "changes the config of the components to increase resiliencey" or something

jwitz, line 127:
```suggestion
By default, the Pods running your Deployment's Airflow components are distributed across multiple nodes. When you enable high availability, Astro runs your nodes in different regions and re-configures specific Airflow components, such as the scheduler, to increase resiliency. This ensures that your DAGs can continue to run if there's an issue with one of your Airflow components in a specific node or availability zone.
```

astro/deployment-resources.md
```diff
@@ -124,7 +124,7 @@ To configure the scheduler on an [Astro Hybrid](hybrid-overview.md) Deployment:
 
 ## Enable high availability
 
-By default, the Pods running your Deployment's Airflow components are distributed across multiple nodes. When you enable high availability, your Deployment runs two instances of [PgBouncer](https://www.pgbouncer.org/) and two instances of the Airflow Scheduler across different nodes. However, Astro makes a best effort to use different availability zones for your PGBouncer and Scheduler, which means it is possible but unlikely that they are both located in the same availability zone. This ensures that your DAGs can continue to run if there's an issue with one of your Airflow components in a specific node or availability zone.
+By default, the Pods running your Deployment's Airflow components are distributed across multiple nodes. When you enable high availability, Astro runs your nodes in different regions and re-configures specific Airflow components, such as the scheduler, to increase resiliency. This ensures that your DAGs can continue to run if there's an issue with one of your Airflow components in a specific node or availability zone.
```
melugoyal, line 127: "Astro runs your nodes in different regions" this is not correct, they are in the same region but in different availability zones

can we just simplify this overall and say "when you enable high availability, your airflow components are run in different availability zones to increase resiliency"? i don't see a strong reason to mention pods and nodes

chomipi88: I think we should call out that their are multiple schedulers at the very least. That's what customers typically care about when concerned with HA.

There is also pressure from competitors like GCC as they get very detailed about what HA is: https://cloud.google.com/composer/docs/composer-2/set-up-highly-resilient-environments#about-components

Issue #3639, opened 2024-04-12
Add cleanup DAG as a Learn tutorial


jwitz:
### Issue Type

Missing content for an existing feature

### Links to Affected Docs

https://github.com/astronomer/docs/pull/3573

### The Issue or Context

Outline:

## Setup

Be sure to note 
- Task times out if longer than 5 minutes
- Any Astro specific configurations
- What happens when the metadata db gets too full

PR #3641, opened 2024-04-15
Lean cleanup DAG tutorial

jwitz:
Resolves https://github.com/astronomer/docs/issues/3639
Resolves https://github.com/astronomer/laminar/issues/1773


code-samples/dags/cleanup-dag-tutorial/db_cleanup.py
```diff
@@ -0,0 +1,60 @@
...
+             --tables {{ params.tables|join(', ') }} \
+        {% endif -%}
+             --verbose \
+             --yes \
+        """,
+        do_xcom_push=False,
+    )
+
+
+astronomer_db_cleanup_dag()
```
jwitz, line 60:
```suggestion
astronomer_db_cleanup_dag()

```

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+import CodeBlock from '@theme/CodeBlock';
+import db_cleanup from '!!raw-loader!../code-samples/cleanup-dag-tutorial/db_cleanup.py';
+
+In addition to storing configuration about your Airflow environment, the Airflow [metadata database](https://docs.astronomer.io/learn/airflow-database) stores data about past and present task runs. The more you use Airflow, the more data the metadata DB stores about these task runs. Over a long enough time, this can result in a bloated metadata DB, which can affect performance across your Airflow environment.
+
+When a table in the metadata DB is larger than 50GB, you might start to experience degraded scheduler performance. This can result in:
+
+- Slow task scheduling
+- Slow DAG parsing
+- [List more specifics]
```
jwitz, line 17: Any other side effects we should mention here?

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+
+:::danger
+
+Read this section carefully before implementing this DAG in any production Airflow environments.
+
+:::
+
+Deleting data from the metadata database can be an extremely destructive action. If you delete data that future task runs depend on, it's difficult to recover the database to its previous state without interrupting your data pipelines. Before implementing the DAG in this tutorial, consider the following:
+
+- Only trigger a database cleanup if you need to. If your Airflow scheduler is running normally, there's no need to clean the DB. On Astro, you will receive alerts when specific tables contain enough data to cause performance issues.
```
jwitz, line 41: More usage warnings here would be great, these are just the ones I can think of off the top of my head

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+title: 'Clean up the Airflow metadata database using DAGs'
+sidebar_label: 'Clean up the metadata database'
+id: cleanup-dag-tutorial
+description: 'Learn how to remove unnecessary data from the Airflow metadata database using the `airflow db clean` command from a DAG.'
+---
+
+import CodeBlock from '@theme/CodeBlock';
+import db_cleanup from '!!raw-loader!../code-samples/dags/cleanup-dag-tutorial/db_cleanup.py';
+
+In addition to storing configuration about your Airflow environment, the Airflow [metadata database](https://docs.astronomer.io/learn/airflow-database) stores data about past and present task runs. The more you use Airflow, the more data the metadata DB stores about these task runs. Over a long enough time, this can result in a bloated metadata DB, which can affect performance across your Airflow environment.
```
kentdanas, line 11:
```suggestion
In addition to storing configuration about your Airflow environment, the Airflow [metadata database](https://docs.astronomer.io/learn/airflow-database) stores data about past and present task runs. Airflow never automatically removes metadata, so the longer you use it, the more task run data is stored in your metadata DB. Over a long enough time, this can result in a bloated metadata DB, which can affect performance across your Airflow environment.
```

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+- `task_reschedule`
+- `task_fail`
+
+To keep your Airflow environment running at optimal performance, you can clean the metadata DB using the `airflow db clean` command. This tutorial walks you through how to implement a cleanup DAG in Airflow so that you can clean your database using the command directly from the Airflow UI.
+
+## Warnings
+
+:::danger
+
+Read this section carefully before implementing this DAG in any production Airflow environments.
```
kentdanas, line 35: I almost wonder if we should move this up into the intro and update to "Read the Warnings section....", in case anybody happens to scroll over it (not that they couldn't do that anyway, but maybe this would help)

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+
+:::
+
+Deleting data from the metadata database can be an extremely destructive action. If you delete data that future task runs depend on, it's difficult to recover the database to its previous state without interrupting your data pipelines. Before implementing the DAG in this tutorial, consider the following:
+
+- Only trigger a database cleanup if you need to. If your Airflow scheduler is running normally, there's no need to clean the DB. On Astro, you will receive alerts when specific tables contain enough data to cause performance issues.
+- Delete older data before newer data. The older the deleted data, the less likely it is to affect your currently running DAGs.
+- The DAG will fail if it runs for longer than five minutes, which can happen when you have an exceptionally large amount of data to delete. If this happens to you, use the DAG's params to reduce the amount of data being deleted at once, either by reducing the time window for deletion or reducing the number of tables to clean.
```
kentdanas, line 43: Is this true generally, or only on Astro? I think Felix said the latter, but we should clarify.

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+
+1. In your dags folder, create a file called `db_cleanup.py`.
+
+2. Copy the following code into the file.
+
+    <CodeBlock language="python">{db_cleanup}</CodeBlock>
+
+    Rather than running on a schedule, this DAG is triggered manually by default and includes params so that you're in full control over how you clean the metadata DB.
+
+    It includes one task, `clean_db`, that runs the `airflow db clean` with the params you specify at Runtime. The params let you specify:
```
kentdanas, line 60:
```suggestion
    It includes one task, `clean_db`, that runs `airflow db clean` with the params you specify at Runtime. The params let you specify:
```

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+
+    - What age of data to delete. Any data that was created before the specified time will be deleted.
```
kentdanas, line 62: I think in all of these bullets worth stating what the default is, in case people don't read the DAG code

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+2. Copy the following code into the file.
+
+    <CodeBlock language="python">{db_cleanup}</CodeBlock>
+
+    Rather than running on a schedule, this DAG is triggered manually by default and includes params so that you're in full control over how you clean the metadata DB.
+
+    It includes one task, `clean_db`, that runs the `airflow db clean` with the params you specify at Runtime. The params let you specify:
+
+    - What age of data to delete. Any data that was created before the specified time will be deleted. 
+    - Whether to run the cleanup as a dry run, meaning that no data is deleted.
```
kentdanas, line 63:
```suggestion
    - Whether to run the cleanup as a dry run, meaning that no data is deleted. The DAG will return the SQL that would be executed based on other parameters you have specified.
```

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+    - Which tables to delete data from.
+
+## Step 2: Run your DAG
```
kentdanas, line 66: It seems like this would be more useful if the step was to deploy the DAG to astro and run as needed? Or if we want to explicitly tell people to run locally so they can "practice" with it, maybe we should state that.

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+
+- `job` 
+- `dag_run` 
+- `task_instance`
+- `log` 
+- `dags` 
+- `task_reschedule`
+- `task_fail`
+
+To keep your Airflow environment running at optimal performance, you can clean the metadata DB using the `airflow db clean` command. This tutorial walks you through how to implement a cleanup DAG in Airflow so that you can clean your database using the command directly from the Airflow UI.
```
kentdanas, line 29:
```suggestion
To keep your Airflow environment running at optimal performance, you can clean the metadata DB using the Airflow CLI `airflow db clean` command. This command was created as a way to safely clean up your metadata DB; you should never modify the Airflow metadata DB directly, as this can cause dependency issues and corrupt your Airflow instance. This tutorial walks you through how to implement a cleanup DAG in Airflow so that you can clean your database using the command directly from the Airflow UI.
```
Feel free to adjust where we put this, but I think the extra context is helpful.

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+- Only trigger a database cleanup if you need to. If your Airflow scheduler is running normally, there's no need to clean the DB. On Astro, you will receive alerts when specific tables contain enough data to cause performance issues.
+- Delete older data before newer data. The older the deleted data, the less likely it is to affect your currently running DAGs.
+- The DAG will fail if it runs for longer than five minutes, which can happen when you have an exceptionally large amount of data to delete. If this happens to you, use the DAG's params to reduce the amount of data being deleted at once, either by reducing the time window for deletion or reducing the number of tables to clean.
+
+## Prerequisites
+
+- An Astro project
+- The Astro CLI
+
+## Step 1: Create your DAG
```
kentdanas, line 50: Somewhere (in this section? the intro?) we should add a note that this DAG has been optimized for Astro users, and while it's still relevant for OSS users, they may want to review and change some of the defaults.

kentdanas (commented):
@jwitz this is a great start, left some initial comments.

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,83 @@
...
+id: cleanup-dag-tutorial
+description: 'Learn how to remove unnecessary data from the Airflow metadata database using the `airflow db clean` command from a DAG.'
+---
+
+import CodeBlock from '@theme/CodeBlock';
+import db_cleanup from '!!raw-loader!../code-samples/dags/cleanup-dag-tutorial/db_cleanup.py';
+
+In addition to storing configuration about your Airflow environment, the Airflow [metadata database](https://docs.astronomer.io/learn/airflow-database) stores data about past and present task runs. The more you use Airflow, the more data the metadata DB stores about these task runs. Over a long enough time, this can result in a bloated metadata DB, which can affect performance across your Airflow environment.
+
+When a table in the metadata DB is larger than 50GB, you might start to experience degraded scheduler performance. This can result in:
```
jwitz, line 13: Is there a more general figure that we could use here for all Airflow environments? This is when the alert triggers on Astro, but I imagine it can start to get slower a lot quicker for other enviros. What figure can OSS users check to see if a table is getting too big?

learn/cleanup-dag-tutorial.md
jwitz: Is it though? By default it deletes the data in all tables older than 90 days, which doesn't seem too Astro specific to me, but I might be missing some context here

learn/cleanup-dag-tutorial.md
jwitz: I moved this slightly higher, but not all the way to the beginning so as to avoid https://en.wikipedia.org/wiki/Banner_blindness

jwitz:
Thank you @kentdanas! Agreed with all of your comments, just left one additional question

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,87 @@
...
+
+The following tables in the database are at risk of becoming too large over time:
+
+- `job` 
+- `dag_run` 
+- `task_instance`
+- `log` 
+- `dags` 
+- `task_reschedule`
+- `task_fail`
```
jedcunningham, line 27: xcom, rendered_task_instance_fields, more so than task_reschedule and task_fail. I don't think dags is a risk either.

The order feels a bit odd to me also.

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,87 @@
...
+
+To keep your Airflow environment running at optimal performance, you can clean the metadata DB using the Airflow CLI `airflow db clean` command. This command was created as a way to safely clean up your metadata DB without modifying it directly, as this can cause dependency issues and corrupt your Airflow instance. This tutorial walks you through how to implement a cleanup DAG in Airflow so that you can clean your database using the command directly from the Airflow UI.
```
jedcunningham, line 29:
```suggestion
To keep your Airflow environment running at optimal performance, you can clean the metadata DB using the Airflow CLI `airflow db clean` command. This command was created as a way to safely clean up your metadata DB without querying it directly. This tutorial walks you through how to implement a cleanup DAG in Airflow so that you can clean your database using the command directly from the Airflow UI.
```

"querying" vs "modifying", and I'd drop the reason since FKs in theory should avoid corruption.

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,87 @@
...
+Even when using `airflow db clean`, deleting data from the metadata database can be destructive . Read the [Warnings](#warnings) section carefully before implementing this tutorial DAG in any production Airflow environments.
+
+:::
+
+## Warnings
+
+
+Deleting data from the metadata database can be an extremely destructive action. If you delete data that future task runs depend on, it's difficult to recover the database to its previous state without interrupting your data pipelines. Before implementing the DAG in this tutorial, consider the following:
+
+- Only trigger a database cleanup if you need to. If your Airflow scheduler is running normally, there's no need to clean the DB. On Astro, you will receive alerts when specific tables contain enough data to cause performance issues.
```
jedcunningham, line 42: I'm not a fan of saying "no need unless there is pain". It is good hygiene and ideally folks do it proactively.

learn/cleanup-dag-tutorial.md
rawwar: I can think of two web server issues:
1. Gunicorn timeout when a query takes a longer time to complete. 
2. Slower UI loading

learn/cleanup-dag-tutorial.md
collinmcnulty: I think this is the answer to your question above about 50 GB. That's difficult for us to know in other environments, so just state this is for Astro

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,85 @@
...
+
+:::
+
+## Warnings
+
+
+Deleting data from the metadata database can be an extremely destructive action. If you delete data that future task runs depend on, it's difficult to recover the database to its previous state without interrupting your data pipelines. Before implementing the DAG in this tutorial, consider the following:
+
+- Delete older data before newer data. The older the deleted data, the less likely it is to affect your currently running DAGs.
+- On Astro, the DAG will fail if it runs for longer than five minutes, which can happen if you have an exceptionally large amount of data to delete. If this happens, use the DAG's params to reduce the amount of data being deleted at once, either by reducing the time window for deletion or reducing the number of tables to clean.
```
collinmcnulty, line 42: Reducing the number of tables won't help. The timeout is per table.

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,85 @@
...
+Even when using `airflow db clean`, deleting data from the metadata database can be destructive . Read the [Warnings](#warnings) section carefully before implementing this tutorial DAG in any production Airflow environments.
+
+:::
+
+## Warnings
+
+
+Deleting data from the metadata database can be an extremely destructive action. If you delete data that future task runs depend on, it's difficult to recover the database to its previous state without interrupting your data pipelines. Before implementing the DAG in this tutorial, consider the following:
+
+- Delete older data before newer data. The older the deleted data, the less likely it is to affect your currently running DAGs.
```
collinmcnulty, line 41: It's not possible to delete new data before old data with this DAG

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,85 @@
...
+- On Astro, the DAG will fail if it runs for longer than five minutes, which can happen if you have an exceptionally large amount of data to delete. If this happens, use the DAG's params to reduce the amount of data being deleted at once, either by reducing the time window for deletion or reducing the number of tables to clean.
+
+## Prerequisites
+
+- An Astro project
```
collinmcnulty, line 46: An Airflow project, nothing too Astro specific here

collinmcnulty (changes requested):
I like this overall, and as long as the other

code-samples/dags/cleanup-dag-tutorial/db_cleanup.py
```diff
@@ -0,0 +1,60 @@
...
+        ),
+        "dry_run": Param(
+            default=False,
+            type="boolean",
+            description="Print the SQL queries that would be run, but do not execute them. Default is False.",
+        ),
+    },
+)
+def astronomer_db_cleanup_dag():
+    BashOperator(
```
josh-fell, line 41: Might be cool to see the new `@task.bash` TaskFlow decorator too (if you can toggle examples of how to implement the code like what's done elsewhere).

Something like (untested of course):
```py
@task.bash
def cleanup_db(params) -> str:
    bash_command = (
        "airflow db clean --clean-before-timestamp {{ params.clean_before_timestamp }} "
        "--skip-archive --verbose --yes"
    )

    if params.dry_run:
        bash_command += "--dry-run "

    if params.tables:
        bash_command += "--tables {{ params.tables|join(', ') }}"

    return bash_command
```

kentdanas: As much as I am pro-@task.bash decorator, I don't think we need to present multiple ways of authoring the same DAG here - if there's a notable benefit to using the decorator for this use case specifically, let's just switch to that. We do that in our other guides where the purpose is "how to write a DAG that does XYZ" to show different potential styles - but the purpose here isn't how to write the DAG, it's "how to clean up your database", so I don't see a lot of benefit for showing different style options.

RNHTTR: IMO the notable benefit is pushing people toward Taskflow. My 2 cents is that if it's between classic and task decorated, we should go with task decorated

feluelle: Yeah, and also likely a lot of people would have to rewrite it to match their airflow version if we use the decorator. Don't get me wrong, I like the decorator syntax much more, but I think it would be more user-friendly to give them a DAG that they could directly run.

learn/cleanup-dag-tutorial.md
feluelle: That is correct. The timeout is per query, and since the cleanup task runs a (cleanup) query per table it won't help.

learn/cleanup-dag-tutorial.md
feluelle: I think what we want to say is to start by deleting very old records. For example, start with older than 180 days, before deleting data older than 90 days.

collinmcnulty: I would state that directly then. This sentence strongly implies that it's possible to end up with a hole in your data.

RNHTTR (dismissed):
Instead of using a BashOperator, could we use [run_cleanup](https://github.com/apache/airflow/blob/main/airflow/utils/db_cleanup.py#L393)? We could use a mapped operator that runs smaller cleanup tasks in parallel instead of trying to clean up everything beyond 90 days, which could cause a statement timeout

feluelle:
@RNHTTR we could do the same with BashOperator. It doesn't have to be a python task to use task mapping.

Users could still run into statement timeout issues when using chunking mechanism, because they don't know what chunk size to use.

If we want to do it more intelligently, we would have to do table lookups first, which we also should only do with care. We have https://github.com/astronomer/astro/issues/13887 to do that automatically. So I think we should keep this tutorial DAG rather simple.

kaxil:
> @RNHTTR we could do the same with BashOperator. It doesn't have to be a python task to use task mapping.
> 
> Users could still run into statement timeout issues when using chunking mechanism, because they don't know what chunk size to use.
> 
> If we want to do it more intelligently, we would have to do table lookups first, which we also should only do with care. We have [astronomer/astro#13887](https://github.com/astronomer/astro/issues/13887) to do that automatically. So I think we should keep this tutorial DAG rather simple.

Agreed. Also the functions under utils might change in Airflow. Can we merge this PR, please?

kaxil:
I would like us to publish something on the website asap. We can always improve on it once it is there. Don't let the perfect be the enemy of the good :)

learn/cleanup-dag-tutorial.md
jwitz: Our standard for tutorials is to say "Astro project", since we instruct users to run the DAG with the Astro CLI and you can't use an OSS Airflow project for that.

collinmcnulty (approved):

kentdanas (approved):

kaxil (approved):

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,90 @@
...
+    - `dry_run`: `true`
+    - `tables`: `all_tables`
+    - `clean_before_timestamp`: `datetime.now(tz=UTC) - timedelta(days=90)`
+
+3. Click **Trigger**. 
+4. After the task completes, click **Graph**.
+5. Click a task run.
+6. Click **Instance Details**.
+7. Click **Log**.
+8. Check that the `airflow db cleanup` command completed successfully. Note that you created a new Astro project, the run will not show much data to be deleted.
```
lzdanski, line 88:
```suggestion
8. Check that the `airflow db cleanup` command completed successfully. Note that if you created a new Astro project, the run will not show much data to be deleted. 
```

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,90 @@
...
+
+    It includes one task, `clean_db`, that runs `airflow db clean` with the params you specify at runtime. The params let you specify:
+
+    - What age of data to delete. Any data that was created before the specified time will be deleted. The default is to delete all data older than 90 days.
+    - Whether to run the cleanup as a dry run, meaning that no data is deleted. The DAG will instead return the SQL that would be executed based on other parameters you have specified. The default is to run the deletion without a dry run.
+    - Which tables to delete data from. The default is all tables.
+
+## Step 2: Run the DAG
+
+In this step, you'll run the DAG in a local Airflow environment to practice the workflow for cleaning metadata DB data. In reality, you would first deploy this DAG to a production environment, such as an Astro Deployment, then run the DAG only when your metadata DB was getting full.
```
lzdanski, line 73:
```suggestion
In this step, run the DAG in a local Airflow environment to practice the workflow for cleaning metadata DB data. When using this process for a production environment, you would first deploy this DAG to a production environment, such as an Astro Deployment, then run the DAG only when your metadata DB is getting full.
```

lzdanski (changes requested):
Some clarity recommendations and typo fixes

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,90 @@
...
+
+Even when using `airflow db clean`, deleting data from the metadata database can be destructive . Read the [Warnings](#warnings) section carefully before implementing this tutorial DAG in any production Airflow environments.
+
+:::
+
+## Warnings
+
+Deleting data from the metadata database can be an extremely destructive action. If you delete data that future task runs depend on, it's difficult to recover the database to its previous state without interrupting your data pipelines. Before implementing the DAG in this tutorial, consider the following:
+
+- When specifying the `clean_before_timestamp` value, use as late of a data as possible. The older the deleted data, the less likely it is to affect your currently running DAGs.
```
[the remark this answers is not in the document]
jwitz: Typo on "data" here, but I think this should still be the "latest date". If you select a recent date, it will delete all data before the recent date, when we specifically want users only to delete later data if possible.

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,90 @@
...
+- On Astro, the DAG will fail if it runs for longer than five minutes, which can happen if you have an exceptionally large amount of data to delete. If this happens, use the DAG's params to reduce the amount of data being deleted from a single table at once by moving the `clean_before_timestamp` earlier.
```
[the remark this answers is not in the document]
jwitz: Same here, a more recent time will cause more data to be deleted.

learn/cleanup-dag-tutorial.md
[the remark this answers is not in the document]
lzdanski: Wouldn't that be the "oldest date" then? When I read "latest date" to me that translates to "most recent," not the "most distant in the past"?

learn/cleanup-dag-tutorial.md
```diff
@@ -0,0 +1,90 @@
...
+
+Even when using `airflow db clean`, deleting data from the metadata database can be destroy important data. Read the [Warnings](#warnings) section carefully before implementing this tutorial DAG in any production Airflow environments.
+
+:::
+
+## Warnings
+
+Deleting data from the metadata database can be an extremely destructive action. If you delete data that future task runs depend on, it's difficult to restore the database to its previous state without interrupting your data pipelines. Before implementing the DAG in this tutorial, consider the following:
+
+- When specifying the `clean_before_timestamp` value, use as late of a date as possible. The older the deleted data, the less likely it is to affect your currently running DAGs.
```
lzdanski, line 41:
```suggestion
- When specifying the `clean_before_timestamp` value, use as old a date as possible. The older the deleted data, the less likely it is to affect your currently running DAGs.
```

lzdanski (commented):
I think that "latest" isn't specific enough for this context - we should use "older" and "newer" or "More and less recent" because it appears "latest" has two very different interpretations among us!

lzdanski (approved):
One final wording tweak but LGTM

Issue #3634, opened 2024-04-11
Document how to add Org / Workspace tokens as Deployment tokens


jwitz:
### Issue Type

Missing content for an existing feature

### Links to Affected Docs

https://docs.astronomer.io/astro/deployment-api-tokens

### The Issue or Context

We need a topic like - https://docs.astronomer.io/astro/workspace-api-tokens#add-an-organization-api-token-to-a-workspace - but for the relationship between Organization and Deployment API tokens, as well as Workspace and Deployment API tokens.

PR #3649, opened 2024-04-16
Add topic for assigning Org/ Workspace tokens to a Deployment

jwitz:
Resolves https://github.com/astronomer/docs/issues/3634

This topic covers some existing functionality and helps to lay the language for other features that utilize this management workflow

### DIFF
--- a/astro/deployment-api-tokens.md
+++ b/astro/deployment-api-tokens.md
@@ -37,6 +37,25 @@ Unlike Workspace API tokens and Organization API tokens, Deployment API tokens a
 
 6. Copy the token and store it in a safe place. You will not be able to retrieve this value from Astro again.
 
+## Assign an Organization or Workspace API token to a Deployment
+
+To centralize API token management, you can add an Organization or Workspace API token to a Deployment instead of creating a dedicated Deployment API token. Deployment-scoped API tokens are useful if you want to manage  API tokens from the Organization or Workspace level while only granting tokens specific access to Deployments. 
+
+Deployment-scoped API tokens are functionally identical to dedicated Deployment API tokens, except that you can only rotate, update, or delete them within their original scope. 
+
+1. In the Astro UI, open your Workspace, then open the Deployment you want to create an API token for.
+
+2. Click **Access**.
+
+3. Click **API Tokens**, then click **+ Deployment API Token**. In the dropdown menu that appears, click either **Assign Workspace API token** or **Assign Organization API token**. 
+
+4. Configure the new Deployment API token:
+
+    - **Workspace/ Organization API Token**: Select the API token you want to assign to the Deployment.
+    - **Deployment Role**: Select the role that the API token has in the Deployment.
+
+5. Click **Update API Token**.
+
 ## Update or delete a Deployment API token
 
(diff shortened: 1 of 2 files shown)


astro/workspace-api-tokens.md
```diff
@@ -38,9 +38,9 @@ There are two ways to use API tokens to interact with a Workspace:
    
 6. Copy the token and store it in a safe place. You will not be able to retrieve this value from Astro again. 
 
-### Add an Organization API token to a Workspace
+### Assign an Organization API token to a Workspace
 
-You might want to add an Organization API token if your Organization API token only has Organization Member permissions, but you want to grant it Workspace Owner permissions for several Workspaces.
+To centralize API token management, you can add an Organization token to a Deployment instead of creating a dedicated Workspace API token. Workspace-scoped API tokens are useful if you want to manage API tokens from the Organization level while only granting tokens select access to Workspaces.
```
lzdanski, line 43: I'm not sure why but this concept is a little unclear to me. Can we give a scenario of why someone would want to do this? Like, what does this specifically mean:
- You add an Organization token
- The token only has permissions for specific workspaces

- Can users with only `Workspace owner` permissions create and manage these tokens if they're being granted at the `organization` level? Earlier in this doc, we say that the minimum permission required is Workspace Owner, but they might not also be an Organization owner/ admin.

lzdanski (commented):
Some questions about scope!

astro/workspace-api-tokens.md
jwitz: There's a few benefits to this setup:

- Organization Owners can manage Workspace tokens alongside all other tokens from a single screen. Workspace Owners, on the other hand, are siloed. If an Org Owner wanted to manage these tokens at the Workspace level, they would have to navigate in and out of Workspaces, making it challenging to handle many tokens at scale.
- You can use a single Organization API token for multiple Deployments. This both allows the token to only be managed by the Organization Owner and makes it quick to rotate or delete a token for multiple Workspaces. 

Workspace users can _not_ manage these tokens, which is the primary benefit. Tried to make that more clear in my edit

astro/deployment-api-tokens.md
```diff
@@ -37,6 +37,25 @@ Unlike Workspace API tokens and Organization API tokens, Deployment API tokens a
 
 6. Copy the token and store it in a safe place. You will not be able to retrieve this value from Astro again.
 
+## Assign an Organization or Workspace API token to a Deployment
+
+To centralize API token management, you can add an Organization or Workspace API token to a Deployment instead of creating a dedicated Deployment API token. Deployment-scoped API tokens are useful if you want to manage  API tokens from the Organization or Workspace level while only granting tokens specific access to Deployments.
```
lzdanski, line 42:
```suggestion
To centralize API token management, you can add an Organization or Workspace API token to a Deployment instead of creating a dedicated Deployment API token. Deployment-scoped API tokens are useful if you want to manage API tokens from the Organization or Workspace level while only granting tokens specific access to Deployments. 
```

lzdanski (approved):
Small typo but otherwise LGTM! Thanks for the clarification!

PR #3663, opened 2024-04-22
Add `--no-hooks` to platform configuration workflow

jwitz:
It's safe and faster to use `--no-hooks` during platform configuration changes. Making this change, plus a note to not use the flag when upgrading Astornomer Software

### DIFF
--- a/software/apply-platform-config.md
+++ b/software/apply-platform-config.md
@@ -71,12 +71,18 @@ A number of configurations, including those for user permissions and identity pr
 2. Save your updated `values.yaml` file and run the following command to apply it as a Helm upgrade:
 
     ```bash
-    helm upgrade <your-platform-release-name> astronomer/astronomer -f <your-updated-config-yaml-file> -n <your-platform-namespace> --set astronomer.houston.upgradeDeployments.enabled=false
+    helm upgrade <your-platform-release-name> astronomer/astronomer -f <your-updated-config-yaml-file> -n <your-platform-namespace> --set astronomer.houston.upgradeDeployments.enabled=false --no-hooks
     ```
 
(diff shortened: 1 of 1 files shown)


software/apply-platform-config.md
```diff
@@ -71,12 +71,18 @@ A number of configurations, including those for user permissions and identity pr
...
-    helm upgrade <your-platform-release-name> astronomer/astronomer -f <your-updated-config-yaml-file> -n <your-platform-namespace> --set astronomer.houston.upgradeDeployments.enabled=false
+    helm upgrade <your-platform-release-name> astronomer/astronomer -f <your-updated-config-yaml-file> -n <your-platform-namespace> --set astronomer.houston.upgradeDeployments.enabled=false --no-hooks
     ```
 
     Setting `astronomer.houston.upgradeDeployments.enabled=false` ensures that no Airflow components in Deployments are restarted during your upgrade.
 
-3. Run the following command to confirm that your configuration was applied:
+    :::caution
+   
+    Astronomer recommmends only setting `--no hooks` if your configuration change does not include a [platform upgrade](upgrade-astronomer.md). To apply platform configuration changes at the same time as a platform upgrade, remove the `--no hooks` flag from this command so that Astronomer Software can properly restart its components after an upgrade.
```
jwitz, line 81:
```suggestion
    Astronomer recommmends setting `--no hooks` only if your configuration change does not include a [platform upgrade](upgrade-astronomer.md). To apply platform configuration changes at the same time as a platform upgrade, remove the `--no hooks` flag from this command so that Astronomer Software can properly restart its components after an upgrade.
```

rob-1126:
There are certainly specific changes where it is OK, if we want to prescribe it specifically in those use cases, that would be OK.

PR #3681, opened 2024-04-25
correct astro init commands in software install

cpsenka:
The final stages of a Software install where a user is supposed to test a deploy is outdated. Updated with correct CLI commands


lzdanski (approved):
@jwitz going to let you merge this in case I'm missing some Software docs context, but the build LGTM

jwitz (approved):

PR #3700, opened 2024-05-02
Integrations sidebar update

kentdanas:
Based on feedback from Ben Garrison, it wasn't clear this section also includes connections docs. Per Paola's suggestion, updating to `Integrations & connections` to make that clearer https://astronomer.slack.com/archives/C03CT3DNZMH/p1714658483362899


TJaniF (approved):
LGTM :)

PR #3702, opened 2024-05-03
Databricks tutorial update

fritz-astronomer:
remove pin, inline task-group

### DIFF
--- a/learn/airflow-databricks.md
+++ b/learn/airflow-databricks.md
@@ -17,13 +17,11 @@ The open source [Astro Databricks provider](https://github.com/astronomer/astro-
 You can create a Databricks Workflow from existing Databricks notebooks as a task group in your Airflow DAG with just a few lines of code:
 
 ```python
-task_group = DatabricksWorkflowTaskGroup(
+with  DatabricksWorkflowTaskGroup(
     group_id="databricks_workflow",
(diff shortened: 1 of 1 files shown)


learn/airflow-databricks.md
```diff
@@ -94,7 +92,7 @@ To get the most out of this tutorial, make sure you have an understanding of:
 2. Add the [Astro Databricks provider package](https://github.com/astronomer/astro-provider-databricks) to your requirements.txt file.
 
     ```text
-    astro-provider-databricks==0.1.6
+    astro-provider-databricks
```
kentdanas, line 95:
```suggestion
    astro-provider-databricks==0.1.6
```
@fritz-astronomer we pin these on purpose so we know the tutorial works - if there's a breaking change in a future provider version we can't guarantee it won't change the steps, so by pinning we can at least know it should work for that snapshot in versions. Ideally they wouldn't get _too_ out of date, but again, bandwidth. If you want to give this a try with a newer version and confirm it works then we can update the pin. Otherwise I'd leave it as is.

fritz-astronomer: I kinda figured that was the reason. I was worried that customers (including the one I just linked this to) will blindly use that pin, though, and I didn't know how out-of-date this was :( 
I getchya, though.

PR #3731, opened 2024-05-13
Add Estimated DAG count to scheduler resources table

astro-anand:
Added details on estimated number of DAGs each size of scheduler can support.


lzdanski (approved):
Consistent with our existing published docs and https://docs.astronomer.io/astro/migrate-mwaa#step-3-create-an-astro-deployment

PR #3738, opened 2024-05-14
5.14 release notes

lzdanski:


astro/release-notes.md
```diff
@@ -25,6 +25,17 @@ Astronomer is committed to continuous delivery of both features and bug fixes to
 
 <!-- ALL LINKS TO INTERNAL DOCS MUST BE COMPLETE URLS INCLUDING HTTPS. Otherwise the links will break in RSS. -->
 
+## May 15, 2024
+
+### Additional improvements
+
+- You can now use [Astro connections](https://docs.astronomer.io/astro/create-and-link-connections) with the Airflow Scheduler. Previously, the scheduler could not access Astro-managed connections due to how the scheduler accesses the local file-systems secret's backend. <!--https://github.com/astronomer/astro/pull/20401-->
```
jwitz, line 32:
```suggestion
- Airflow connections that you configure through the Astro UI environment manager are now mounted to Deployment schedulers, meaning that scheduler processes can now make use of these Airflow connections.
```
Few things:
- Let's avoid the phrase "Astro connections", making sure to always use "Airflow connections" so that these are distinguishable from network connections to Astro.
- Our style guide dictates that we should never capitalize _any_ Airflow components, including scheduler, webserver, etc. The only capitalized components are Astronomer-specific RBAC components, like Deployment, Workspace, and RBAC role titles
- "You can now use Astro connections with the Airflow scheduler" the way this is worded I can't actually tell what the user actually wants to do with the connection, I tried rewording to be a little more generic and make it lower level. If there's a specific use case for this, let's call it out, but otherwise it seems like a tech requirement rather than a use case in and of itself.

astro/release-notes.md
```diff
@@ -25,6 +25,17 @@ Astronomer is committed to continuous delivery of both features and bug fixes to
...
+
+### Bug fixes
+
+- Fixed a bug with SCIM integrations where some updates could not be made to Organization-less users. <!--https://github.com/astronomer/astro/pull/20843-->
+- Fixed an issue that could allow users to use custom roles to create API Tokens, users, or teams with higher Deployment access privilege. Users can now only use custom roles to make resources with their access privileges or lower. <!--https://github.com/astronomer/astro/pull/20826-->
```
jwitz, line 37:
```suggestion
- Fixed an issue where you couldn't update users who were added through SCIM but didn't belong to an Organization.
- Fixed an issue where a user with a custom role could create API Tokens, users, or teams with greater permissions than their own.
```

jwitz (approved):
Some minor suggestions for simplifying the notes, otherwise LGTM!

jwitz:
I would pull in some tech review for the first one actually just to be on the safe side

aliotta (approved):

PR #3761, opened 2024-05-17
Update dag-best-practices.md

RNHTTR:
Enabling retries is so important and shouldn't be buried at the bottom of the best practices. Also, nice to highlight an added (if not the primary) benefit of idempotence.


kentdanas (approved):
@RNHTTR thanks for the PR - I agree with you this is important and should be moved up. I think we need to review this doc more broadly and update (and maybe rearrange) some of the guidance, but we won't get to that this week so I'm going to go ahead and merge for now and we can do more updates later :)

PR #3780, opened 2024-05-24
Add instructions for custom postgres image in local dev

stuart23:
Added a section to describe how to specify a different Postgres image for `astro dev start`


stuart23:
Closing this as using a different postgres version is better done with the astro project config

Issue #3829, opened 2024-06-06
Clean up formula in KEDA scaling page


RNHTTR:
### Issue Type

Issue with website functionality

### Links to Affected Docs

astro/celery-executor.md

https://www.astronomer.io/docs/astro/celery-executor#celery-worker-autoscaling-logic

### The Issue or Context

Using a code block is kinda ugly

### Fix or Proposed Change

Use something like latex to make it look like...

<img width="413" alt="image" src="https://github.com/astronomer/docs/assets/25823361/e527ce43-06b7-42d7-8527-771c8001a437">

PR #3830, opened 2024-06-06
Update celery-executor.md with cleaner formulas

RNHTTR:
Closes: #3829 

Not sure if this is the right way to go about it, but figured it was worth taking a stab at.

### DIFF
--- a/astro/celery-executor.md
+++ b/astro/celery-executor.md
@@ -29,11 +29,16 @@ The number of Celery workers running per worker queue on your Deployment at a gi
 
 The calculation is made based on the following expression:
 
-`[Number of workers]= ([Queued tasks]+[Running tasks])/(Concurrency)`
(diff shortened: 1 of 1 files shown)


lzdanski:
So there's a few ways of adding equations to docs, but typically instead of using some kind of image, we need to add support for LaTex/KaTex, which Docusaurus *can* support. But it'll take some experimenting to get right, and until then, I think an image + sufficient alt-text might be more of a lateral swap from what we currently have. There's a couple things going on with the build that need fixing before the preview works, so I'll have to come back to this PR with those suggestions, but wanted to give you a quick update! 

Here's the info about math support in docusaurus- 
https://docusaurus.io/docs/markdown-features/math-equations

lzdanski:
@RNHTTR - going to keep the related issue open for now but we need to approach improving these equations a little differently than adding images. Closing PR for now.

PR #3861, opened 2024-06-13
Add an audit logs screencap from the Airflow UI to the logging Learn guide

merobi-hub:
The Learn guide on logging could use a screencap of the audit logs.
Screenshot 2024-06-13 at 18 22 10


learn/logging.md
```diff
@@ -77,6 +77,8 @@ By default, Airflow outputs logs to the `base_log_folder` configured in `airflow
 
 Airflow also makes audit logs available in the Airflow UI under the "Browse" tab. For details about the log levels and events tracked, see [Audit logs in Airflow](https://airflow.apache.org/docs/apache-airflow/stable/security/audit_logs.html).
 
+![Audit Logs List View in the Airflow UI](/img/guides/audit-logs-cap.png)
```
kentdanas, line 80: Maybe add a box around the `browse` tab so it's clear? Based on the slack comment, I think it would also be worth adding a sentence like "For example, you can see which user triggered a DAG run", then have the screenshot, and then the existing "For more details about the log levels and events tracked...."

kentdanas (changes requested):

kentdanas (approved):

PR #3863, opened 2024-06-17
Update alerts.md to highlight DAG trigger channel is private preview

RNHTTR:


lzdanski (approved):

RNHTTR:
@lzdanski only nit is that the private preview block in the tab is a warning and not info -- should that one also be info or should they both be warning?

PR #3885, opened 2024-06-26
[0.35.0 release or later] Add yanked info to Software docs

lzdanski:
**NOT BLOCKING FOR 0.35.0 RELEASE**

We overlooked adding information about yanked runtimes versions to the Software docs when they were introduced to Astro. This PR adds them to it.

Differences:

- Since *Use a deprecated runtime version* is an Astro Hosted feature and the process of updating available Astro runtime versions for an Airgapped environment are different, the Software docs include a caution about Airgapped environments and no mention of Deprecated runtime options. 

Questions for reviewers:

@jed - Are these yanked runtimes still up to date and do the descriptions make sense for Software users to also avoid?

@tony - we briefly discussed adding this content via slack. If the changes look good to you, i think your approval will be sufficient. If we need technical review, you will need to add them as a reviewer or reach out to them via slack. 

@paolaperaza  - If reviews are complete before 0.35.0 launch, merge this PR into the release PR. If reviews complete *after* the 0.35.0 launch, you can merge it directly to `main`. 

FOR COMPARISON:
- ASTRO: Runtime maintenance policy section (https://www.astronomer.io/docs/astro/runtime-version-lifecycle-policy#restricted-runtime-versions)
- ASTRO: Upgrade considerations (https://www.astronomer.io/docs/astro/upgrade-runtime#upgrade-considerations)


software/manage-airflow-versions.md
```diff
@@ -188,8 +188,27 @@ If an Astro Runtime version isn't included in this section, then there are no sp
 
 :::
 
+##### Restricted versions of Runtime 11
+
+You cannot create new Deployments or upgrade to the following versions of Runtime 11, which are `yanked`. These [restricted runtime versions](runtime-version-lifecycle-policy.mdx#restricted-runtime-versions) prevent you from upgrading to or creating a Deployment with a version that contains a known limitation or bug.
+
+- 11.0.0
+- 11.1.0
+
```
danielhoherd, line 197: If we are going to maintain a list of yanked versions here then it is something we need to automate. Otherwise, we should just give a small reference of yanked images and a link to a source where a current list can be viewed.

lzdanski: This has been discussed with the Airflow team before we committed to having it in the docs. I don't have access to the issue/slack thread but we have it in the analogous Astro doc.

I would consider this not a blocking factor but what we can do is make sure both files that need to be maintained are listed in comments on each page so that when one is updated, the other is too.

danielhoherd: oh, yeah, i didn't meant this to be a blocker for merging this. it's just that maintaining this by hand is sisyphean.

jedcunningham:
The docs look good, but does software look at this flag?

lzdanski:
Closing, will address in future PR

PR #3890, opened 2024-06-28
Remove listener reference in cross deployment best practice + update task name

TJaniF:
Two tiny changes to the cross-deployment guide:

- noticed a sentence that still referenced a listener
- changed the task name to not be the same as the listener hook name, this is just cosmetic to avoid confusion :)


merobi-hub (approved):
Thanks for catching this @TJaniF !

kentdanas (approved):

PR #3941, opened 2024-07-17
Remove Kaniko example

stuart23:
We cannot call out alternative tooling because that makes us liable to supporting this tooling. This needs to be removed from docs.


lzdanski (approved):

lzdanski:
For some reason I'm not an approver of the best-practices docs, I'm not sure what our rules are for these resources, but merging since this is a doc I authored.

Issue #3884, opened 2024-06-26
Missing documentation for Astro CLI command "astro organization audit-logs"


wolfier:
### Issue Type

Missing content for an existing feature

### Links to Affected Docs

https://www.astronomer.io/docs/astro/category/astro-organization

### The Issue or Context

There are references in the [Astro docs](https://www.astronomer.io/docs/astro/audit-logs#export-audit-logs) but no documentation on the command `astro organization audit-logs`.

EdwardAngert:
@wolfier - is that <https://www.astronomer.io/docs/astro/audit-logs>?

looks like it's currently under **Platform observability**, but when I run `astro organization audit-logs -h`, I get:

```shell
Manage your Organization audit logs.

Current Context: Astro

Usage:
  astro organization audit-logs [command]

Aliases:
  audit-logs, al

Available Commands:
  export      Export your Organization audit logs in GZIP. Requires Organization Owner permissions.

Flags:
  -h, --help   help for audit-logs

Global Flags:
      --verbosity string   Log level (debug, info, warn, error, fatal, panic (default "warning")

Use "astro organization audit-logs [command] --help" for more information about a command.
```

It might be easy enough to add a link to it in the sidebar (leaving the file where it is), but I can't find a way to make it look consistent in Docusaurus:

<img width="1136" alt="SCR-20240718-twfh" src="https://github.com/user-attachments/assets/37cb4ec0-9f45-448a-954c-5d984159b5ca">

---

Now that I've written all that out to test it, I'll just put in a PR and let you all decide if that works.

PR #3952, opened 2024-07-19
3884: Link to astro org audit logs from astro-org category sidebar

EdwardAngert:
Fixes: https://github.com/astronomer/docs/issues/3884

Since Docusaurus is [generating the page](https://docusaurus.io/docs/sidebar/items#generated-index-page) based on frontmatter, I'm not sure there's a way to adjust how it appears on this page, but if that's where @wolfier and other user's are looking for it, could be worth it.

SCR-20240718-twfh-2

### DIFF
--- a/sidebarsAstro.js
+++ b/sidebarsAstro.js
@@ -604,6 +604,7 @@ module.exports = {
             description: 'Use `astro organization` commands to manage users and their Organization-level permissions.'
           },
           items: [
+            "audit-logs",
             "cli/astro-organization-list",
             "cli/astro-organization-switch",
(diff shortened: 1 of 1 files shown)


lzdanski:
Hey @EdwardAngert - so we'd actually make a new reference page for this instead of linking to another part of the docs in this way. I'm going to close this PR and instead add the change to the next CLI release branch. Thank you for bringing this to my attention! I'll be sure to add a link to this resource to the new page.

PR here: https://github.com/astronomer/docs/pull/3960

EdwardAngert:
Thanks, @lzdanski! That makes sense

PR #3954, opened 2024-07-19
HTML to extend AppBuilder view in Airflow Plugin

indraneel07:


TJaniF (approved):
This is great, thank you @indraneel07 ! :)

PR #4003, opened 2024-08-02
Add more IP Access danger info

### LINKED ISSUE #3970: Inform users they will need to include IPs from their IDP for IP Access Lists
### Issue Type

### Links to Affected Docs

https://www.astronomer.io/docs/astro/ip-access-list

### The Issue or Context

When customers setup IP Access lists for their Astronomer accounts they need a note / to be made aware that 
they need to include IPs from their IDP/SSO providers.

Otherwise, those will be blocked as well.

### Fix or Proposed Change

Add a 5th step calling this out or make the "danger" notice a little longer to include.

"You must include the IP address of your identity provider (IdP) or it will be locked out of the system."

### Additional Notes

https://astronomer.slack.com/archives/C01UA03TPV1/p1721773001947879?thread_ts=1720643464.522469&cid=C01UA03TPV1

### Required Reviewers

_No response_

lzdanski:
Resolves #3970

### DIFF
--- a/astro/set-up-ip-access.md
+++ b/astro/set-up-ip-access.md
@@ -25,7 +25,10 @@ After you enable the IP access list, users and user-privileged resources can onl
 4. Add the IP Adress range or ranges in CIDR format.
     :::danger
 
-    Make sure the first IP range you add is inclusive of the IP you are currently using, or you will be locked out of your organization.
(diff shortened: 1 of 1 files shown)


chobberoni (approved):
Thank you!

PR #4011, opened 2024-08-05
Update astro-support.md

RNHTTR:
people often include this kind of data as screenshots, which isn't particularly useful.

### DIFF
--- a/astro/astro-support.md
+++ b/astro/astro-support.md
@@ -33,9 +33,9 @@ The more information you can provide about the issue you're experiencing, the qu
 - What solutions have you already tried?
 - Is this a problem in more than one Deployment?
 
-### Include logs or code snippets as text and not screenshots
+### Include logs or code snippets as text and not screenshots {#include-logs-or-code-snippets}
 
-If you've already copied task logs or Airflow component logs, send them as a part of your request. The more context you can provide, the better.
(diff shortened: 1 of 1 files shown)


astro/astro-support.md
```diff
@@ -33,7 +33,7 @@ The more information you can provide about the issue you're experiencing, the qu
 - What solutions have you already tried?
 - Is this a problem in more than one Deployment?
 
-### Include logs or code snippets
+### Include logs or code snippets as text and not screenshots
```
lzdanski, line 36:
```suggestion
### Include logs or code snippets as text and not screenshots {#include-logs-or-code-snippets}
```
Totally understand why this needs to be the header and in every single link, I think if I add the `{#stuff}` anchor it means we can use both options for links.

astro/astro-support.md
```diff
@@ -33,7 +33,7 @@ The more information you can provide about the issue you're experiencing, the qu
...
 
 If you've already copied task logs or Airflow component logs, send them as a part of your request. The more context you can provide, the better.
```
lzdanski, line 38:
```suggestion
If you've already copied task logs or Airflow component logs, send them a part of your request as text. The more context you can provide in your request, the better.
```

lzdanski (approved):

PR #4023, opened 2024-08-09
Update cross-dag dependencies guide with TriggerDagRunOperator change

merobi-hub:
Adds a note about the ability, new in 2.10, to configure the operator not to attempt to trigger runs that have already been successfully triggered.


learn/cross-dag-dependencies.md
```diff
@@ -134,6 +134,12 @@ In the following image, you can see that the `trigger_dependent_dag` task in the
 
 If your dependent DAG requires a config input or a specific execution date, you can specify them in the operator using the `conf` and `execution_date` params respectively.
 
+:::tip
+
+Starting with version 2.10, you can set `skip_when_already_exists` to `True` to keep the operator from attempting to trigger runs that have already occurred and failing as a result. This could happen in the case of rerun DAGs and tasks.
```
kentdanas, line 139: Does "failing as a result" means that if the condition is met the run will fail? It seems like it would be skipped?

merobi-hub: Meant to say that if the condition is met (the variable is set to true), the operator won't trigger already-occurred runs, which HAD been causing runs to fail. Am I misunderstanding, or should I just clean up the wording/punctuation?

merobi-hub: @TJaniF what do you think?

kentdanas: ah I see, so you're saying that if the parameter didn't exist, the operator would re-trigger a run and it would fail because it has already run? I do think that's accurate, and in that case I would just clean up the wording. Maybe something like "....already occurred, which would cause the run to fail" or something like that.

TJaniF: I got a ping from this PR but can't find it 😅 but yes, using the TDRO twice with the same dag_id and execution date used to give you

```
airflow.exceptions.DagRunAlreadyExists: A DAG Run already exists for DAG tdro_downstream at 2024-08-12 00:00:00+00:00 with run id manual__2024-08-12T00:00:00+00:00
```
and the task fails.

Now it skips the TDRO operator instead of failing with

```
[2024-08-13, 13:22:32 UTC] {taskinstance.py:300} INFO - Skipping due to skip_when_already_exists is set to True and DagRunAlreadyExists
```

lzdanski, line 139:
```suggestion
Starting with Airflow version 2.10, you can set `skip_when_already_exists` to `True` to keep the operator from attempting to trigger runs that have already occurred, and failing as a result. This can happen when trying to rerun DAGs and tasks.
```

lzdanski (approved):
Small optional wording change - see if it makes sense.

PR #4046, opened 2024-08-15
Add logging and error handling to scripts in deploy-with-api.md

odaneau-astro:
adding set -ex to all shell scripts to print commands to the logs and interrupt in case of errors.


odaneau-astro:
I would also consider using environment variables like so

```
ORGANIZATION_ID=$ORGANIZATION_ID
DEPLOYMENT_ID=$DEPLOYMENT_ID
ASTRO_API_TOKEN=$ASTRO_API_TOKEN
ASTRO_PROJECT_PATH=$ASTRO_PROJECT_PATH
```

PR #4051, opened 2024-08-15
Add caveat for pulling from a custom registry for ECR

stuart23:
A customer tried these instructions with ECR but it does not work because the credentials expire.


stuart23:
Thanks @lzdanski - can you please go ahead and merge this?

PR #4055, opened 2024-08-16
Add config file note

### LINKED ISSUE #4037: Update config file instructions with info/warning
### Issue Type

### Links to Affected Docs

https://www.astronomer.io/docs/astro/deployment-file-reference#deployment-file-example
https://www.astronomer.io/docs/astro/manage-deployments-as-code?tab=dedicated#create-a-deployment-using-a-template-file

### The Issue or Context

Unclear what the minimum config values are

### Fix or Proposed Change

Need to add the following text to *before* the code example and also the deploy config file example:

These are the minimum values required to create a Deployment using a template file. Any configurations not specified are set to default values. To add more configurations, see [Deployment file reference](https://www.astronomer.io/docs/astro/deployment-file-reference).

### Additional Notes

https://astronomer.slack.com/archives/C01UA03TPV1/p1723491896360719

### Required Reviewers

_No response_

lzdanski:
Resolves #4037


ben-astro:
Hello! The change to the https://www.astronomer.io/docs/astro/manage-deployments-as-code#create-a-deployment-using-a-template-file doc looks good to me. 

Regarding the second one https://www.astronomer.io/docs/astro/deployment-file-reference, I have a couple comments. 

- The info bubble text noted "these are the defaults" which I don't think is true, I think that reference file is just giving example values for all fields/keys. For example, some of them don't even make sense together, like the cloud_provider = gcp but the node-type is an AWS node type (m5.xlarge). We probably need someone from PM team to validate what the defaults are, and change that doc to specify the defaults & maybe also configurable options (that would be my recommendation). 
- I still am not sure that `m5.xlarge` is the right `default_worker_type` value, given that this is under the Hosted section for this doc and Hosted deployments operate with our abstracted node types (A5, A10, etc) that we list in our [worker queues doc](https://www.astronomer.io/docs/astro/configure-worker-queues#hosted-worker-types) now, at least insofar as customers can see. Also would want confirmation from a PM on expectations for this in the doc.

PR #4063, opened 2024-08-19
Learn IA updates

kentdanas:


kentdanas:
@TJaniF any input on my categorizations here? I struggled a bit with extra links and object storage. Would like to keep it to 3ish categories for now.

TJaniF:
> @TJaniF any input on my categorizations here? I struggled a bit with extra links and object storage. Would like to keep it to 3ish categories for now.

I think the current sorting is the most sensible one. Extra links are tricky because they are only going to be relevant to people writing operators, so not _directly_ DAG writing but of these 3 categories still the most fitting 🤔 
The object storage 'split' of Xcom backend with managing Airflow and regular object storage use with DAG writing makes a lot of sense :)

sidebarsLearn.js
```diff
@@ -133,26 +133,59 @@ module.exports = {
...
+        {
+          type: 'category',
+          label: 'Write DAGs',
+          link: {
+            type: 'generated-index',
+            title: 'Write DAGs',
+            description: 'Tutorials on various aspects of writing Airflow DAGs',
+          },
+          items: [
+            'operator-extra-link-tutorial',
```
TJaniF, line 152: Maybe we could rename that tutorial to 'Operator Extra Links', both to make clear it is an operator property that is talked about and also so it is not the first tutorial in the alphabetic list? :) 
And I wonder if it is possible to have an entry in the list that is called something like 'First Steps' and opens up the get started tutorials? 🤔 I worry about beginners not seeing 'Get started', clicking on tutorials, and then only seeing these highly specific ones

TJaniF (approved):
Had one suggestion in a comment but generally LGTM :)

sidebarsLearn.js
kentdanas: Yeah good call on the extra links one, I updated the sidebar name. 

I'm not sure about the linking a subsection like that for the tutorials, but I see your point. Let's not worry about that for this PR, but can keep chatting about it when we get to updating the quickstarts. It may actually make sense to move the "Get started" tutorials under "Tutorials", and rename "Get started" to "Quickstart"

PR #4065, opened 2024-08-19
Astro sdk cleanup

kentdanas:
The Astro Python SDK is no longer maintained and will not be in the future. This doesn't remove all instances of it from Learn, but at least takes care of the easy ones where it's not part of a lengthy code block. I'm not too concerned about those for now, given page views, but wanted to take this opportunity for higher exposure low hanging fruit.


TJaniF (approved):
LGTM!

